PR #1, opened 2017-06-28
Write a README

timholy:
This seems to be working pretty well now. I'll give watchers a brief chance to play with it and report problems before I register this puppy.


timholy:
(I am definitely enjoying this package!)

PR #16, opened 2017-07-04
Add exclusion/silencing mechanisms and exclude GSL

### LINKED ISSUE #15: UV Error when attempting to watch (too) many files
`Revise.jl` is loaded in my `.juliarc.jl`,  then:

```
              _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.0 (2017-06-19 13:05 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> using GSL

ERROR (unhandled task failure): start_watching (File Monitor): too many open files (EMFILE)
Stacktrace:
 [1] uv_error at ./libuv.jl:68 [inlined]
 [2] start_watching(::FileMonitor) at ./poll.jl:350
 [3] wait(::FileMonitor) at ./poll.jl:428
 [4] watch_file(::String, ::Int64) at ./poll.jl:507
 [5] revise_file_queued(::String) at /Users/spowell/.julia/v0.6/Revise/src/Revise.jl:412
 [6] (::Revise.##1#2)() at ./event.jl:73
```

The error repeats 50+ times.

If the number of watched files cannot be increased, or if this would be undesirable, perhaps an easy way to manage the problem would be to have a whitelist of modules for which tracking is not required?

timholy:
Fixes #15 

There's a secondary fix possible, to solve the reported

> The error repeats 50+ times.

However, given that it's still early days for this package, I'm thinking it might be better to temporarily annoy people so that they are more likely to report the problem and therefore build up a more comprehensive list of packages-to-be-excluded. Thoughts? CC @samuelpowell.


samuelpowell:
This makes a lot of sense, and the addition of the silencing mechanism should prevent any undue annoyance.

timholy:
Thanks for looking it over!

samuelpowell:
My pleasure, this package is already saving me a _lot_ of time. Thank you for your efforts!

PR #22, opened 2017-07-14
Watch directories rather than files

timholy:
This is an attempt at a more robust solution to #15 than was #16. It has some disadvantages---for example, editors that autosave a tempfile to a watched directory will trigger a callback to Julia---but given that OSX struggles with watching a lot of files, this may be the best solution.

CC @cstjean, @sampowell. I've tested this a bit and it seems OK to me, but it might be nice to hear reports from OSX users. Note that this is back to watching GSL, so you should get an error pretty quickly if this is not working.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -419,7 +419,7 @@ function watch_files_via_dir(dirname)
     wf = watched_files[dirname]
     for file in wf.trackedfiles
         path = joinpath(dirname, file)
-        if mtime(path) > wf.timestamp
+        if mtime(path) + 1 >= floor(wf.timestamp) # OSX rounds mtime up, see #22
             push!(latestfiles, path)
         end
     end
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -15,7 +15,7 @@ end
 const watched_files = Dict{String,WatchList}()
 
 ## For excluding packages from tracking by Revise
-const dont_watch_pkgs = Set([:GSL])
+const dont_watch_pkgs = Set{Symbol}()
 const silence_pkgs = Set{Symbol}()
 const depsdir = joinpath(dirname(@__DIR__), "deps")
 const silencefile = Ref(joinpath(depsdir, "silence.txt"))  # Ref so that tests don't clobber
@@ -542,9 +542,7 @@ end
 """
     Revise.silence(pkg)
 
-Silence warnings about not tracking changes to package `pkg`. Some
-packages (e.g., GSL) are excluded because they load so many files that
-it burdens the file-watcher.
+Silence warnings about not tracking changes to package `pkg`.
 """
 function silence(pkg::Symbol)
     push!(silence_pkgs, pkg)
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -5,6 +5,14 @@ module Revise
 export revise
 
 const revision_queue = Set{String}()  # file names that have changed since last revision
+# Some platforms (OSX) have trouble watching too many files. So we
+# watch parent directories, and keep track of which files in them
+# should be tracked.
+mutable struct WatchList
+    timestamp::Float64         # unix time of last revision
+    trackedfiles::Set{String}
+end
+const watched_files = Dict{String,WatchList}()
 
 ## For excluding packages from tracking by Revise
 const dont_watch_pkgs = Set([:GSL])
@@ -405,6 +413,20 @@ function parse_module!(md::ModDict, ex::Expr, file::Symbol, mod::Module, path)
     newmod
 end
 
+function watch_files_via_dir(dirname)
+    watch_file(dirname)  # this will block until there is a modification
+    latestfiles = String[]
+    wf = watched_files[dirname]
+    for file in wf.trackedfiles
+        path = joinpath(dirname, file)
+        if mtime(path) > wf.timestamp
+            push!(latestfiles, path)
+        end
+    end
+    updatetime!(wf)
+    latestfiles
+end
+
 function watch_package(modsym::Symbol)
     if modsym ∈ dont_watch_pkgs
         if modsym ∉ silence_pkgs
@@ -413,23 +435,33 @@ function watch_package(modsym::Symbol)
         return nothing
     end
     files = parse_pkg_files(modsym)
+    udirs = Set{String}()
     for file in files
-        @schedule revise_file_queued(file)
+        dir, basename = splitdir(file)
+        haskey(watched_files, dir) || (watched_files[dir] = WatchList())
+        push!(watched_files[dir], basename)
+        push!(udirs, dir)
+    end
+    for dir in udirs
+        updatetime!(watched_files[dir])
+        @schedule revise_dir_queued(dir)
     end
     return nothing
 end
 
-function revise_file_queued(file)
-    if !isfile(file)
+function revise_dir_queued(dirname)
+    if !isdir(dirname)
         sleep(0.1)   # in case git has done a delete/replace cycle
-        if !isfile(file)
-            warn(file, " is not an existing file, Revise is not watching")
+        if !isfile(dirname)
+            warn(dirname, " is not an existing directory, Revise is not watching")
             return nothing
         end
     end
-    watch_file(file)  # will block here until the file changes
-    push!(revision_queue, file)
-    @schedule revise_file_queued(file)
+    latestfiles = watch_files_via_dir(dirname)  # will block here until file(s) change
+    for file in latestfiles
+        push!(revision_queue, file)
+    end
+    @schedule revise_dir_queued(dirname)
 end
 
 function revise_file_now(file0)
@@ -621,5 +653,14 @@ function __init__()
         end
     end
 end
+
+## WatchList utilities
+function updatetime!(wl::WatchList)
+    tv = Libc.TimeVal()
+    wl.timestamp = tv.sec + tv.usec/10^6
+end
+Base.push!(wl::WatchList, filename) = push!(wl.trackedfiles, filename)
+WatchList() = WatchList(Dates.datetime2unix(now()), Set{String}())
+Base.in(file, wl::WatchList) = in(file, wl.trackedfiles)
 
 end # module
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,1 +1,1 @@
-GSL
+Example
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -300,26 +300,26 @@ end
     end
 
     @testset "Pkg exclusion" begin
-        if !(is_windows() && Int == Int32) # GSL install fails on 32-bit Windows
-            @eval import GSL
-            for k in keys(Revise.file2modules)
-                if contains(k, "GSL")
-                    error("Should not track files in GSL")
-                end
+        push!(Revise.dont_watch_pkgs, :Example)
+        push!(Revise.silence_pkgs, :Example)
+        @eval import Example
+        for k in keys(Revise.file2modules)
+            if contains(k, "Example")
+                error("Should not track files in Example")
             end
-            # Ensure that silencing works
-            sfile = Revise.silencefile[]  # remember the original
-            try
-                sfiletemp = tempname()
-                Revise.silencefile[] = sfiletemp
-                Revise.silence("GSL")
-                @test isfile(sfiletemp)
-                pkgs = readlines(sfiletemp)
-                @test contains(==, pkgs, "GSL")
-                rm(sfiletemp)
-            finally
-                Revise.silencefile[] = sfile
-            end
+        end
+        # Ensure that silencing works
+        sfile = Revise.silencefile[]  # remember the original
+        try
+            sfiletemp = tempname()
+            Revise.silencefile[] = sfiletemp
+            Revise.silence("GSL")
+            @test isfile(sfiletemp)
+            pkgs = readlines(sfiletemp)
+            @test contains(==, pkgs, "GSL")
+            rm(sfiletemp)
+        finally
+            Revise.silencefile[] = sfile
         end
     end
 end


timholy:
Also CC @sbromberger in case it helps his baffling need to call `revise` manually.

timholy:
Hmm, this seems to have made things *worse* on OSX. I might need help from an OSX user to debug this.

samuelpowell:
I am @samuelpowell not @sampowell as above, so didn't see your ping. Is there anything I can do to help?

timholy:
Sorry about the messed-up ping! (To both of you.)

And thanks for the offer of help! If you can `Pkg.checkout("Revise", "teh/watch_dirs")` and give it a whirl. Assuming it fails in "normal" usage, then perhaps insert some debugging statements in various spots. (e.g., `@show revision_queue` [here](https://github.com/timholy/Revise.jl/blob/master/src/Revise.jl#L462), obviously other places might be necessary).

If in contrast it works in normal usage but only fails tests, then perhaps it's just a timing thing?

samuelpowell:
Okay I'll give this a shot tomorrow morning (UK)

cstjean:
I just tried it, 

```julia
using Revise
using OrdinaryDiffEq
using GSL
```

works now, whereas it failed before. `Pkg.test` has the same errors as on Travis. There's probably a logic error somewhere; will debug when I get a change.

cstjean:
You've probably considered this already, but instead of launching `N` `watch_file` processes, wouldn't it be easier to just check the modification date of each tracked file whenever the user enters a command at the REPL? `@btime stat("/Applications/Julia-0.6.app/Contents/Resources/base/sysimg.jl").mtime` takes less than a microsecond on OSX, so you could check a lot of files before it's a significant slowdown (granted, this benchmark may be misleading due to caching, but `@elapsed` on a random file is 30 microseconds).

timholy:
It's worth considering. But I wouldn't be entirely surprised if folks sometimes track 500-1000 files, and 30ms may not be so negligible.

timholy:
One data point: after running the GLVisualize tests, I'm tracking 452 files. Running `mtime` on them all is ~35ms. That's not huge, but you can just notice it.

sbromberger:
@timholy I'm happy to help test out on OSX - just let me know what you need me to do (more specific is better).

timholy:
See https://github.com/timholy/Revise.jl/pull/22#issuecomment-315451496; at a minimum, just checking out this branch and reporting your experience will help. From the Travis failures there is some debugging to do, but from https://github.com/timholy/Revise.jl/pull/22#issuecomment-315455044 it sounds like this has some "in practice" promise.

cstjean:
It's a timing bug. Changing the `sleep` calls in runtests' `yry() = (sleep(0.1); revise(); sleep(0.1))` to 2 seconds each fixed all but 1 failure, and changing it to 5 seconds fixed all of them.

samuelpowell:
@timholy I've been using this for a few hours now and it seems robust.

timholy:
Glad to hear this is working in practice.

From the standpoint of the tests, 5 seconds seems *crazy*, the OS shouldn't need that long. And if it does, we'll get bugs in interactive usage because folks will frequently execute a command within less than 5 seconds of saving a file. Finally, we didn't need that long back when we were watching files rather than directories.

What about adding `flush(io)` as the last line of the `open` blocks, e.g., just above [lines like these](https://github.com/timholy/Revise.jl/blob/06ed964297071f2be80bfb5a0ed5b98f5c016555/test/runtests.jl#L154)?

cstjean:
> What about adding flush(io)

Didn't help.

samuelpowell:
This doesn’t appear to have any effect.

On 15 Jul 2017, at 14:09, Tim Holy <notifications@github.com<mailto:notifications@github.com>> wrote:


Glad to hear this is working in practice.

From the standpoint of the tests, 5 seconds seems crazy, the OS shouldn't need that long. And if it does, we'll get bugs in interactive usage because folks will frequently execute a command within less than 5 seconds of saving a file. Finally, we didn't need that long back when we were watching files rather than directories.

What about adding flush(io) as the last line of the open blocks, e.g., just above lines like these<https://github.com/timholy/Revise.jl/blob/06ed964297071f2be80bfb5a0ed5b98f5c016555/test/runtests.jl#L154>?

cstjean:
Tried this to quantify the delay:
```julia
> while true
>     watch_file("/Users/cedric/Programa")
>     @show time() - mtime("/Users/cedric/Programa/dummy.txt")
> end

# `touch` the file repeatedly in a terminal
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.5564489364624023
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.17713594436645508
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.8093059062957764
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.9283499717712402
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.5245048999786377
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.08036303520202637
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.6614151000976562
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.3394160270690918
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.8979310989379883
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.5021030902862549
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.3166849613189697
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.40420007705688477
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.18465399742126465
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.7014529705047607
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.2500929832458496
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.970628023147583
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.5295419692993164
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.05443406105041504
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.6506540775299072
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.7773680686950684
```

timholy:
OMG. For comparison, here's Linux:
```julia
julia> while true
           watch_file("/tmp/timing.txt")
           @show time() - mtime("/tmp/timing.txt")
       end
time() - mtime("/tmp/timing.txt") = 0.008331775665283203
time() - mtime("/tmp/timing.txt") = 0.0017702579498291016
time() - mtime("/tmp/timing.txt") = 0.005357027053833008
time() - mtime("/tmp/timing.txt") = 0.002613544464111328
time() - mtime("/tmp/timing.txt") = 0.003486156463623047
time() - mtime("/tmp/timing.txt") = 0.0032646656036376953
time() - mtime("/tmp/timing.txt") = 0.002134561538696289
time() - mtime("/tmp/timing.txt") = 0.004884958267211914
time() - mtime("/tmp/timing.txt") = 0.00482630729675293
time() - mtime("/tmp/timing.txt") = 0.004861354827880859
time() - mtime("/tmp/timing.txt") = 0.0039081573486328125
time() - mtime("/tmp/timing.txt") = 0.004782676696777344
time() - mtime("/tmp/timing.txt") = 0.002710580825805664
time() - mtime("/tmp/timing.txt") = 0.0056591033935546875
time() - mtime("/tmp/timing.txt") = 0.002579212188720703
time() - mtime("/tmp/timing.txt") = 0.003469228744506836
time() - mtime("/tmp/timing.txt") = 0.0025243759155273438
```

timholy:
Oh, you were watching the dir not the file. (On Linux they give you the same result.)

For comparison, what are the timings like if you watch the file directly?

cstjean:
There's a weird effect: each `touch` triggers `watch_file` twice, hence why it takes longer on even lines. Still way slower than Linux.

```
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.949307918548584
time() - mtime("/Users/cedric/Programa/dummy.txt") = 1.9624691009521484
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.417741060256958
time() - mtime("/Users/cedric/Programa/dummy.txt") = 1.6081271171569824
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.9227490425109863
time() - mtime("/Users/cedric/Programa/dummy.txt") = 2.1249399185180664
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.487515926361084
time() - mtime("/Users/cedric/Programa/dummy.txt") = 1.5650100708007812
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.861565113067627
time() - mtime("/Users/cedric/Programa/dummy.txt") = 2.007297992706299
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.6241199970245361
time() - mtime("/Users/cedric/Programa/dummy.txt") = 1.7327840328216553
time() - mtime("/Users/cedric/Programa/dummy.txt") = 0.8215749263763428
time() - mtime("/Users/cedric/Programa/dummy.txt") = 2.565164089202881
```

cstjean:
I wonder why the delay doesn't cause testing failures on Revise's `master`.

timholy:
Curious to know if the latency is due to OSX, libuv, or (unlikely) julia. If you change the above to `@profile watch_file("/Users/cedric/Programa")` does the output of `Profile.print()` reveal anything interesting?

If this long poll latency isn't something we can easily fix, then perhaps we should close this and go back to the current method...except that on OSX we don't do any polling at all, and adopt your suggestion above of checking the `mtime` on every file.

cstjean:
```
23 ./event.jl:430; (::Base.##300#301{IJulia.#send_std...
 23 ./event.jl:357; wait(::Timer)
  23 ./stream.jl:42; stream_wait(::Timer, ::Condition,...
   23 ./event.jl:27; wait(::Condition)
    21 ./event.jl:216; wait()
     21 ./libuv.jl:82; process_events
      1  ./event.jl:394; uv_timercb(::Ptr{Void})
      12 ./event.jl:399; uv_timercb(::Ptr{Void})
       4 ./event.jl:47; notify(::Condition, ::Void, ::B...
       4 ./event.jl:48; notify(::Condition, ::Void, ::B...
       1 ./event.jl:49; notify(::Condition, ::Void, ::B...
       3 ./event.jl:54; notify(::Condition, ::Void, ::B...
    1  ./event.jl:234; wait()
    1  ./event.jl:241; wait()
     1 ./libuv.jl:80; process_events
1  ./task.jl:335; (::IJulia.##11#14)()
 1 ...IJulia/src/eventloop.jl:8; eventloop(::ZMQ.Socket)
  1 ...src/execute_request.jl:196; execute_request(::ZMQ.Socket, ::I...
   1 ...v0.6/IJulia/src/msg.jl:48; send_ipython(::ZMQ.Socket, ::IJul...
    1 ...a/v0.6/ZMQ/src/ZMQ.jl:393; ZMQ.Message(::String, ::Ptr{UInt...
     1 ...a/v0.6/ZMQ/src/ZMQ.jl:345; gc_protect_handle(::String)
      1 ./dict.jl:421; setindex!(::Dict{Ptr{Void},Any}...
       1 ./dict.jl:348; ht_keyindex2(::Dict{Ptr{Void},...
        1 ./dict.jl:210; hashindex
         1 ./hashing.jl:5; hash
          1 ./hashing.jl:10; hash
```

timholy:
Nothing too surprising there. Just to check the Julia side a little more, do you see something like this?
```julia
julia> c = Condition()
Condition(Any[])

julia> function mywait(c)
           wait(c)
           Δt = time() - mytime[]
           @show Δt
           @schedule mywait(c)
       end
mywait (generic function with 1 method)

julia> const mytime = Ref(time())
Base.RefValue{Float64}(1.500131620310978e9)

julia> @schedule mywait(c)
Task (runnable) @0x00007f41ba2671f0

julia> mytime[] = time(); notify(c)
Δt = 0.00275397300720214841


julia> mytime[] = time(); notify(c)
Δt = 9.202957153320312e-51


julia> mytime[] = time(); notify(c)
Δt = 6.29425048828125e-51


julia> mytime[] = time(); notify(c)
Δt = 6.198883056640625e-51


julia> mytime[] = time(); notify(c)
Δt = 0.000104188919067382811


julia> mytime[] = time(); notify(c)
Δt = 7.295608520507812e-51
```

samuelpowell:
```
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.0 (2017-06-19 13:05 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> c = Condition()
Condition(Any[])

julia> function mywait(c)
                  wait(c)
                  Δt = time() - mytime[]
                  @show Δt
                  @schedule mywait(c)
              end
mywait (generic function with 1 method)

julia> const mytime = Ref(time())
Base.RefValue{Float64}(1.50013191502657e9)

julia> @schedule mywait(c)
Task (runnable) @0x00000001238786d0

julia> mytime[] = time(); notify(c)
Δt = 0.0020501613616943361


julia> mytime[] = time(); notify(c)
Δt = 7.414817810058594e-51


julia> mytime[] = time(); notify(c)
Δt = 7.104873657226562e-51


julia> mytime[] = time(); notify(c)
Δt = 7.295608520507812e-51


julia> mytime[] = time(); notify(c)
Δt = 7.009506225585938e-51


julia> mytime[] = time(); notify(c)
Δt = 7.319450378417969e-51
```

cstjean:
```julia
julia> c = Condition()
Condition(Any[])

julia> function mywait(c)
                  wait(c)
                  Δt = time() - mytime[]
                  @show Δt
                  @schedule mywait(c)
              end
mywait (generic function with 1 method)

julia> const mytime = Ref(time())
Base.RefValue{Float64}(1.500131953417719e9)

julia> @schedule mywait(c)
Task (runnable) @0x0000000111073a90

julia> mytime[] = time(); notify(c)
Δt = 0.0024440288543701171


julia> mytime[] = time(); notify(c)
Δt = 7.510185241699219e-51


julia> mytime[] = time(); notify(c)
Δt = 6.604194641113281e-51


julia> mytime[] = time(); notify(c)
Δt = 6.29425048828125e-51


julia> mytime[] = time(); notify(c)
Δt = 6.794929504394531e-51


julia> mytime[] = time(); notify(c)
Δt = 6.198883056640625e-51
```

timholy:
OK, so it's unlikely to be on the Julia side. I will file a Julia issue in case there's something that can be done. Meanwhile, I suspect this should be closed and we try again with a check-every-file interface.

vtjnash:
I recommend going with the directory polling approach. This is recommended by Apple as being more reliable and scaling better (https://developer.apple.com/library/content/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html).

But anyways, here's a more accurate timing for `watch_file` in each case:
```
julia> for i = 1:10
              t1 = tic()
              @async touch("VERSION")
              watch_file(".")
              toc()
              end
elapsed time: 0.014973798 seconds
elapsed time: 0.01163164 seconds
elapsed time: 0.011636839 seconds
elapsed time: 0.011765657 seconds
elapsed time: 0.011677741 seconds
elapsed time: 0.011554242 seconds
elapsed time: 0.011660417 seconds
elapsed time: 0.011168355 seconds
elapsed time: 0.011687557 seconds
elapsed time: 0.010673983 seconds

julia> for i = 1:10
              t1 = tic()
              @async (yield(); touch("VERSION"))
              watch_file("VERSION")
              toc()
              end
elapsed time: 0.061740413 seconds
elapsed time: 0.00016992 seconds
elapsed time: 9.1868e-5 seconds
elapsed time: 9.3964e-5 seconds
elapsed time: 8.5777e-5 seconds
elapsed time: 0.000101364 seconds
elapsed time: 7.6415e-5 seconds
elapsed time: 9.2019e-5 seconds
elapsed time: 7.9775e-5 seconds
elapsed time: 7.7759e-5 seconds
```

I'm not certain, but maybe it may be _good_ to have a short delay between observing the file being modified and trying to read it? AFAIK, most editor will write to a temporary file, then move it over the old file, to give a fairly atomic view of the change operation. But perhaps not all or in all cases?

cstjean:
I just spent some time looking at the test results again. When I put `sleep(1)` in `yry()`, the only "File Paths" tests that can fail (over >10 trials) are the `pc1()` and `npc1()` tests. And the `pc1()` tests either all fail, or all succeed, which is odd given the multiple `sleep` calls, no? PC.jl never seems to get pushed onto `latestfiles`. I tried adding

```julia
        if mtime(path) > wf.timestamp
            push!(latestfiles, path)
        else
            @show path mtime(path) wf.timestamp
        end
```

in `watch_files_via_dir`, and the testing results are:

```julia
path = "/Users/cedric/.julia/v0.6/Revise/test/testing/PC/src/file2.jl"
mtime(path) = 1.500434186e9
wf.timestamp = 1.500434187293732e9
path = "/Users/cedric/.julia/v0.6/Revise/test/testing/PC/src/PC.jl"
mtime(path) = 1.500434187e9
wf.timestamp = 1.500434187293732e9
path = "/Users/cedric/.julia/v0.6/Revise/test/testing/PC/src/file5.jl"
mtime(path) = 1.500434186e9
wf.timestamp = 1.500434187293732e9
File paths: Test Failed
  Expression: pc1() == -1
   Evaluated: 1 == -1
...
```

It's late, will look more into it tomorrow, but does any of this ring a bell? 

I wonder if `Dates.datetime2unix(now())` and `mtime()` are using exactly the same clock.

cstjean:
I think I got most of it: the resolution of `mtime()` on OSX is at the second level. There's only a `sleep(0.1)` call between the first and second writes to `PC.jl`, so nine times out of ten, that file's `mtime` doesn't change at all. If I change it to `sleep(1)` (in addition to having `sleep(1)` in `yvr`), then all tests pass consistently.

`datetime2unix(now())` is four hours off from `mtime()` on my system (EST), though that's probably overwritten in `watch_package`, so it's irrelevant. Libc.TimeVal looks accurate, but the modification time on OSX seems to be rounded _down_ to the second, as shown by this code.

```julia
t = Dates.datetime2unix(now())
tv = Libc.TimeVal()
tvt = tv.sec + tv.usec/10^6
open("tmpp.tmp", "w") do f
    write(f, "hi")
end
m = mtime("tmpp.tmp")
@show m - t tvt - m
```

`tvt - m` is always positive (and <1), even though logically it should be negative. I think that's what's happening: the `watched_files[dir]` time is greater than the `mtime()` of the file, even though it happened first.

vtjnash:
On most platforms, you can mostly avoid needing to inspect `mtime`, but instead just get the file change list directly. Note however that since this is edge-triggered, it may drop events (which is why `watch_file` doesn't report filenames). It might be useful for someone to change this to use an infinite non-blocking Channel.
```julia
julia> fm = Base.Filesystem.FileMonitor(".");

julia> @async begin
                  Base.Filesystem.start_watching(fm)
                  while true
                      filename, events = wait(fm.notify)
                      @show filename, events
                  end
              end
Task (runnable) @0x00007fdd2832f1f0

shell> touch VERSION
(filename, events) = ("VERSION", Base.Filesystem.FileEvent(false, true, false))

shell> touch base
(filename, events) = ("base", Base.Filesystem.FileEvent(true, true, false))
```

src/Revise.jl
```diff
@@ -405,6 +413,20 @@ function parse_module!(md::ModDict, ex::Expr, file::Symbol, mod::Module, path)
     newmod
 end
 
+function watch_files_via_dir(dirname)
+    watch_file(dirname)  # this will block until there is a modification
+    latestfiles = String[]
+    wf = watched_files[dirname]
+    for file in wf.trackedfiles
+        path = joinpath(dirname, file)
+        if mtime(path) > wf.timestamp
```
cstjean: If I change this line to `if mtime(path) >= floor(wf.timestamp)`, then all tests pass, even with `sleep(0.1)`. It might trigger a few needless checks into the files, but that should be rather minor.

timholy: Nice. Unfortunately it seems a bit flaky, see the Travis tests (passed on 50% of the runs).

timholy: I'll try again with adding 1s of tolerance.

cstjean:
> One data point: after running the GLVisualize tests, I'm tracking 452 files. Running mtime on them all is ~35ms. That's not huge, but you can just notice it.

FWIW, on OSX, touching a file modifies the directory update time, so we'd only have to check directories. This might bring another set of problems though...

PR #25, opened 2017-07-15
Support type redefinition (WIP)

timholy:
This is work towards supporting redefinition of types. It relies on https://github.com/JuliaLang/julia/pull/22721, so it will only be available on 0.7 and higher. AFAICT the only missing part is to populate [method2rex](https://github.com/timholy/Revise.jl/blob/89e9104ccfcb838f145d98fa1ba597be7f5bd07f/src/Revise.jl#L332). (It would be nicer if we could just get the source-expression directly from the `Method`, but currently that's not cached.)

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -240,22 +240,96 @@ end
 #     error("definition for module $modsym not found in $s")
 # end
 
+const oldbinding_counter = Ref(0)
+
 function eval_revised(revmd::ModDict)
+    to_delete = RelocatableExpr[]
     for (mod, exprs) in revmd
+        empty!(to_delete)
+        # Any type (re)definitions must come first
         for rex in exprs
-            ex = convert(Expr, rex)
-            try
-                eval(mod, ex)
-            catch err
-                warn("failure to evaluate changes in ", mod)
-                println(STDERR, ex)
+            if istypedef(rex)
+                typsym = rex.args[2]
+                typ = getfield(mod, typsym)
+                # Get the list of methods that we're going to have to
+                # re-bind to the new name
+                meths = invalid_methods(typ)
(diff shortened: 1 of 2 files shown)


ExpandingMan:
Is there anything blocking this?

timholy:
> Is there anything blocking this?

Um, yes. In the OP I linked to a Julia PR for which @vtjnash points out significant problems. AFAIK those barriers remain.

ExpandingMan:
Ah, forgive my ignorance, it was not apparent to me that either of type renaming or method deletion were needed to make this work.

On a less ambitious note, I wonder if the Julia devs would be open to having a `replinit.jl` in addition to a `startup.jl`?  I imagine this running just *after* the REPL initailizes rather than just *before* like `startup.jl`.  I'd imagine there'd be lots of use for this., even though ideally the `atreplinit()` function would provide that functionality.  Certainl a `replinit.jl` would solve the problem of only initializing Revise (or any other package) with the REPL.  I of course realize that even if the devs are open to it, it would have to wait until 1.1.  Any thought on that, is it a bad idea?

ExpandingMan:
Sorry, I feel silly, I actually posted that question on the wrong PR!  Nevermind!

timholy:
No problem! I'd like this to work too. I'm confident we'll get there eventually, but it's difficult work and there still remains quite a lot of low-hanging fruit.

Nosferican:
Any updates on this?

timholy:
Nope. Anyone interested in tackling it might take pointers from https://discourse.julialang.org/t/manually-reloading-a-module-in-1-1/25572/11?u=tim.holy.

serenity4:
@timholy Should we close this?

timholy:
Superseded by #894

PR #33, opened 2017-08-02
Use OrderedSet

cstjean:
We could then add a user-facing function that forces reloading every rex in a module. This would be a work-around for the macro/generated function issues.

On both master and this branch, I'm getting a failure, four times out of five:

```julia
Cleanup: Test Failed
  Expression: contains(readstring(warnfile), "is not an existing directory")
Stacktrace:
 [1] macro expansion at /Users/cedric/.julia/v0.6/Revise/test/runtests.jl:457 [inlined]
 [2] macro expansion at ./test.jl:860 [inlined]
 [3] macro expansion at /Users/cedric/.julia/v0.6/Revise/test/runtests.jl:445 [inlined]
 [4] macro expansion at ./test.jl:860 [inlined]
```

I haven't investigated, but I printed `warnfile`, and it's empty.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -121,7 +121,7 @@ __precompile__($pcflag)
 
 module $modname
 
-export $(fbase)1, $(fbase)2, $(fbase)3, $(fbase)4, $(fbase)5
+export $(fbase)1, $(fbase)2, $(fbase)3, $(fbase)4, $(fbase)5, using_mac$(fbase)
 
 $(fbase)1() = 1
 
@@ -130,6 +130,10 @@ include("subdir/file3.jl")
 include(joinpath(@__DIR__, "subdir", "file4.jl"))
 otherfile = "file5.jl"
 include(otherfile)
+macro mac$(fbase)()
+    return $(fbase)1()
+end
+using_mac$(fbase)() = @mac$(fbase)()
 
 end
 """)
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -129,6 +130,10 @@ include("subdir/file3.jl")
 include(joinpath(@__DIR__, "subdir", "file4.jl"))
 otherfile = "file5.jl"
 include(otherfile)
+macro mac$(fbase)()
```
timholy: I confess I don't quite follow what's happening here.

timholy:
I like the plan of force-evaling every `rex` as a workaround---great idea.

test/runtests.jl
cstjean: The first code update changes the `using_mac()` definition, and the `$(fbase)1()` definition, which is called in the macroexpansion. If the former is updated before the latter (which had a 50% chance before this PR), then the macroexpansion will return 1, and the test will fail.

It's not a great test, admittedly, since it might have passed before this PR anyway. Do you have a better suggestion?

timholy: Let's start with the basics: the point of the test is to exercise a macro? I spent some time trying to figure out whether this was intended to be OSX-specific, since we've had trouble with OSX. ("mac" = "macro" and not "macintosh"). OK, the purpose is much clearer now. 

Since we can't control hash order, and in any case you've fixed that now anyway, I'm fine with a test that might have passed before but *will* pass now.

test/runtests.jl
```diff
@@ -150,24 +159,30 @@ end
...
 $(fbase)1() = -1
 include("file2.jl")
 include("subdir/file3.jl")
 include(joinpath(@__DIR__, "subdir", "file4.jl"))
 otherfile = "file5.jl"
 include(otherfile)
+macro mac$(fbase)()
+    return $(fbase)1()
+end
+using_mac$(fbase)() = @mac$(fbase)()+0
```
timholy: How does this allow you to detect a change in the definition of `using_mac$(fbase)`? Wouldn't you rather add 1 or something?

cstjean:
Sorry, you're right, that code wasn't very clear. I simplified it somewhat.

test/runtests.jl
cstjean:
```julia
macro some_macro()
    return 6
end
using_macro() = @some_macro()

# Bad update order
using_macro() = @some_macro()+0
macro some_macro()
    return -6
end

# Eval
using_macro()  # 6 with the bad order, -6 if the order of the definitions is flipped
```

Hopefully that wasn't me misunderstanding how Revise.jl works...

timholy:
Thank you so much for tackling this! A very nice enhancement.

PR #34, opened 2017-08-03
Fix depwarn on latest master

eveydee:
Split it into two commits, since I'm not sure if the version check should be there or not with version tagging, etc.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -319,7 +319,7 @@ function parse_source!(md::ModDict, file::AbstractString, mod::Module, path)
         warn("omitting ", file, " from revision tracking")
         return false
     end
-    parse_source!(md, readstring(file), Symbol(file), 1, mod, path)
+    parse_source!(md, read(file, String), Symbol(file), 1, mod, path)
 end
 
 function parse_source!(md::ModDict, src::AbstractString, file::Symbol, pos::Integer, mod::Module, path)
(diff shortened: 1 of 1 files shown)


src/Revise.jl
```diff
@@ -319,7 +319,11 @@ function parse_source!(md::ModDict, file::AbstractString, mod::Module, path)
         warn("omitting ", file, " from revision tracking")
         return false
     end
-    parse_source!(md, readstring(file), Symbol(file), 1, mod, path)
+    if VERSION >= v"0.7.0-DEV"
```
timholy: Thanks!

It's likely you'll have to be more specific about this version number (just using "DEV" has generated complaints in the past). If you can find the commit sha at which this change occurred (check the julia git log), then the `contrib/commit-name.sh` script in julia should give you the version number.

ararslan: Alternatively, you could use the Compat package, which provides a `read(x, String)` method that calls `readstring` on older versions.

eveydee: I can't find the `read` method in the latest release of the `Compat` package 🤷‍. And `@macroexpand @compat read(x, String)` returns `:(read(x, String))` on both 0.6 and master. I found the commit, though, so will update the version check.

ararslan: Sorry, my bad—the Compat PR that includes `read(x, String)` hasn't been merged yet.

timholy:
Thanks!

Issue #31, opened 2017-07-31
Generated functions and world age


cstjean:
Generated functions only use the functions as they were in the world age where they were defined. 

> Some operations that should not be attempted include:
> Calling any function that is defined after the body of the generated function. 

This means:
```julia
module Tmp
f() = 3
@generated g() = f()
end
```

```julia
julia> using Revise, Tmp

julia> Tmp.f()
3

julia> Tmp.g()
3

# Change f
julia> Tmp.f()
WARNING: Method definition f() in module Tmp at /Users/cedric/Programa/RL/Tmp.jl:2 overwritten at /Users/cedric/Programa/RL/Tmp.jl:2.
4

julia> Tmp.g()
3
```

Furthermore, because the rexes are stored in an unordered set, I believe that changing both `Tmp.f` and `Tmp.g` (to `g() = f()+0`, say) would have a 50% chance of returning 3 or 4. Same with macros.

timholy:
I don't think there's anything that can be done here about the core problem---all this package (currently) does is put a convenient wrapper around capabilities in Base. Since that doesn't work when you do it manually, it won't work with Revise either.

You're right about the unsorted Set though. We should rebase on OrderedSet from DataStructures. Quite busy now so I wouldn't mind a PR, but I will likely get to it eventually.

timholy:
Unless generated functions start keeping track of their backedges to functions called from their generator, there's nothing we can do here.

Demo:
```julia
julia> module Tmp
       f() = 3
       g() = 5
       @generated function h()
           x = f()
           :($x, g())
       end
       k() = f()
       l() = g()
       end
Tmp

julia> Tmp.h()
(3, 5)

julia> Tmp.k()
3

julia> Tmp.l()
5

# Functions called from the "body" (quoted expression) do work
julia> m = first(methods(Tmp.g))
g() in Tmp at REPL[1]:3

julia> mi = m.specializations.func
MethodInstance for g()

julia> mi.backedges
2-element Array{Any,1}:
 MethodInstance for h()
 MethodInstance for l()

# Functions called from the generator are not tracked
julia> m = first(methods(Tmp.f))
f() in Tmp at REPL[1]:2

julia> mi = m.specializations.func
MethodInstance for f()

julia> mi.backedges
1-element Array{Any,1}:
 MethodInstance for k()
```

#41 is as close as we can come to fixing this now, the rest is a Base issue.

Issue #20, opened 2017-07-10
Changes to macros are ignored


cstjean:
I've been poking at Revise and found an issue with macros, that you're probably already aware of:

```julia
module B
macro some_val()
    4   # change here
end

f() = @some_val()   # f() doesn't get updated
end
```

timholy:
I don't think it's Revise-specific:
```julia
julia> module B
       macro some_val()
           4   # change here
       end

       f() = @some_val()   # f() doesn't get updated
       end
B

julia> B.f()
4

julia> eval(B, quote
       macro some_val()
           5
       end
       end)
WARNING: Method definition @some_val() in module B at REPL[1]:3 overwritten at REPL[3]:3.
@some_val (macro with 1 method)

julia> B.f()
4
```

@vtjnash, I presume this is a known limitation of the fix of 265? Should I file an issue or not?

vtjnash:
No, this isn't a bug, that Julia is not a pure functional language.

timholy:
Understood. I'll add a caveat to the README.

cstjean:
> No, this isn't a bug, that Julia is not a pure functional language.

That macros are expanded at "parse-time" (or shortly thereafter) instead of compile-time is a Lisp hold-over, no? 

> Understood. I'll add a caveat to the README.

If there was a function to force reloading every function in the module, it might be a good work-around. In ClobberingReload, any change to the file reloads the entire module, which is wasteful, but also solves this issue in the single-module case (and for multiple modules, it's possible to specify dependencies explicitly).

vtjnash:
That said, Revise.jl can track any additional information it would like, to (partially) cover additional cases. So it could keep track of macros and for direct redefinition know which uses to re-expand.

vtjnash:
> That macros are expanded at "parse-time" (or shortly thereafter) instead of compile-time is a Lisp hold-over, no?

No, Julia (like lisp) expand them at expansion time. That's just the definition of what it means to be a macro. A function that is expanded a runtime is known simply as "a function".

timholy:
> So it could keep track of macros and for direct redefinition know which uses to re-expand.

True. I'll reopen. I don't plan on implementing this anytime soon, but for macro definitions that occur after Revise is loaded (i.e., not ones in Base), this doesn't sound too hard.

timholy:
[...]

PR #41, opened 2017-09-09
revise(module) to force reevaluating every expression

cstjean:
As suggested in https://github.com/timholy/Revise.jl/pull/33#issue-247406423

Issues:

- I like `revise(module)` as a verb, but it's not entirely obvious what it's going to do.  `forcerevise(module)`? 
- Both here and in TraceCalls.jl, I get the module files from `parse_pkg_files`, which seems a tad wasteful given that those files were already parsed. We could maintain `module2files`.
- `revise(module)` triggers a lot of redefinition warnings. In ClobberingReload I use [this code](https://github.com/cstjean/ClobberingReload.jl/blob/master/src/scrub_stderr.jl). We could move over `scrub_stderr` to Revise.jl

This is a work-around for #20 and #31, I'll leave you to decide if you want to close these issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,10 @@ There are some kinds of changes that Revise cannot incorporate into a running Ju
 
 - changes to type definitions
 - function or method deletions
-- changes to macros that affect method definitions
 - file or module renames
+- changes to macros that affect method definitions, or to functions that affect generated
+function expansion. To work around this issue, you may explicitly call `revise(module)`
+to force reevaluating every definition in `module`.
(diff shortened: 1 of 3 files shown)


cstjean:
Sorry - accidental click.

timholy:
If you're up for the `module2files` route, I think I'd prefer that.

timholy:
Thanks a bunch!

The OSX tests routinely fail at the "cleanup" step. I don't suppose you can reproduce that yourself, can you? If so, does adding a sleep or something fix it?

cstjean:
I've seen it a few times, I'll try to fix it.

PR #44, opened 2017-09-15
Add module2files

cstjean:
This is the easiest definition to support #41. I wish that `parse_pkg_files` returned the files in evaluation order, but I can't quite follow its logic to fix it. In any case, I don't think the order matters for #41: if I modify a macro's definition in MyModule.jl and call `revise(MyModule)`, the macro will be updated by the regular `revise()` call before `revise(MyModule)` is executed.

Alternatively, we could define `module2files` as an inverse of `file2modules`, but it's not as straight-forward, and I didn't see any upside.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -266,6 +266,7 @@ function eval_revised(revmd::ModDict)
 end
 
 const file2modules = Dict{String,FileModules}()
+const module2files = Dict{Symbol,Vector{String}}()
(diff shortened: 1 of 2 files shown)


timholy:
> I wish that `parse_pkg_files` returned the files in evaluation order, but I can't quite follow its logic to fix it.

Hopefully addressed in #45. Please review and request changes if needed for clarity.

PR #45, opened 2017-09-15
module2files and a bunch of comments

timholy:
Added on top of #44

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -266,6 +266,7 @@ function eval_revised(revmd::ModDict)
 end
 
 const file2modules = Dict{String,FileModules}()
+const module2files = Dict{Symbol,Vector{String}}()
 const new_files = String[]
 
 function parse_pkg_files(modsym::Symbol)
@@ -285,6 +286,7 @@ function parse_pkg_files(modsym::Symbol)
         parse_source(mainfile, Main, dirname(mainfile))
         files = new_files
     end
+    module2files[modsym] = copy(files)
     files
 end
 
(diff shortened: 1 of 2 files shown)


cstjean:
LGTM. My hesitation was over the [first branch in `parse_pkg_files`](https://github.com/timholy/Revise.jl/blob/e10a80e981500337b3511313e4320402effb1c92/src/Revise.jl#L277). Why is there an `if` at all? Couldn't both cases be handled by the else branch? If `!isempty(paths)`, then `module2files[modsym] = map(first, files_mtimes)`. Does `Base.cache_dependencies` return files in the correct order?

timholy:
Revise can't parse tricky include statements like `for file in files; include(file); end`, but if the package is precompiled then the list of included files is available in the `*.ji` file. From my limited experiments, the order is correct (but I don't know if it's guaranteed).

timholy:
I added more comments to explain this to anyone else poking at the source.

timholy:
@cstjean, thanks for kicking this off!

cstjean:
Thank you for helping out with my PR!

> Revise can't parse tricky include statements like for file in files; include(file); end,

I was thinking about this recently... If, for example, [DataFrames' include statements](https://github.com/JuliaData/DataFrames.jl/blob/master/src/DataFrames.jl#L74) were rewritten as a macro

```julia
@include_in_dir(("other", "utils.jl"),
                ("other", "index.jl"),
                ...)
```

then we could perhaps macroexpand top-level macro statements and get the include statements from there.

Getting the files from the .ji is a very nice work-around, though.

PR #47, opened 2017-09-16
fix internal name change of JLOptions field

fredrikekre:
xref https://github.com/JuliaLang/julia/commit/d2041509ad94759c0fa61426f6b194473a6c8669


timholy:
Merging despite the Travis bottleneck.

Thanks, @fredrikekre!

PR #48, opened 2017-09-27
Fix Revise on 0.7

timholy:
https://github.com/JuliaLang/julia/pull/23579 changed how modules are loaded, and for Revise's purposes the important change appears to be that the module isn't externally available while its `__init__` method is being run or when `Base.package_callbacks` are being run. This `@schedule`s Revise's workflow until after module definition is completely finished. @JeffBezanson, is this correct (and intended)? The first commit is the one that implements the fix.

Also fixes some deprecations, and attempts to fix test failures on OSX (I'm just guessing here). CC @cstjean

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -517,10 +517,16 @@ module is "parented" by `mod`. Source-code expressions are added to
 `md` under the appropriate module name.
 """
 function parse_module!(md::ModDict, ex::Expr, file::Symbol, mod::Module, path)
+    mod = moduleswap(mod)
     newmod = getfield(mod, _module_name(ex))
     md[newmod] = OrderedSet{RelocatableExpr}()
     parse_source!(md, ex.args[3], file, newmod, path)  # recurse into the body of the module
     newmod
+end
+if VERSION >= v"0.7.0-DEV.1877"
+    moduleswap(mod) = mod == Base.__toplevel__ ? Main : mod
+else
+    moduleswap(mod) = mod
 end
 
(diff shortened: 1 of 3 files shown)


cstjean:
On OSX, `Revise.watch_files_via_dir("/Users/cedric/nodir")` doesn't return when I `rm -r nodir`. Does it on Linux?

timholy:
```julia
julia> Revise.watch_files_via_dir("/tmp/nodir")
ERROR: KeyError: key "/tmp/nodir" not found
Stacktrace:
 [1] getindex at ./dict.jl:474 [inlined]
 [2] watch_files_via_dir(::String) at /home/tim/.julia/v0.6/Revise/src/Revise.jl:521
```
(That's triggered when I delete the directory.)

cstjean:
Then isn't that where the test failure comes from? `revise_dir_queued` blocks on `watch_files_via_dir`, so it never gets to `warn(dirname, " is not an existing directory, Revise is not watching")`.

timholy:
I see, thanks for explaining. Curiously, it did pass tests on some OSX installations, but not others. Let's just skip that test on OSX, then.

Windows was failing for an independent reason (only on 0.7), let's see if the latest works on all platforms.

cstjean:
> Curiously, it did pass tests on some OSX installations, but not others. 

Might be timing-dependent: `rm dir/file` still triggers `watch_files_via_dir`, so if the order of events is `rm dir/file`, `rm dir`, `revise_dir_queued(dir)`, then we'll see the warning.

PR #49, opened 2017-09-28
Eliminate need to parse include paths

### LINKED ISSUE #40: Packages with fancy `include` mechanisms
Followup to https://github.com/timholy/Revise.jl/pull/10#issuecomment-313971103. Both [`DataFrames.jl`](https://github.com/JuliaStats/DataFrames.jl/blob/master/src/DataFrames.jl#L74-L99) and [`Plots.jl`](https://github.com/JuliaPlots/Plots.jl/blob/b585a0ef61b1af1308cf7ebf0f746c42b39d3869/src/backends.jl#L28) include files through complicated mechanisms. Plots.jl might be a challenge to handle reliably.

> I wondered when this would come up. One option is to put a @require Revise, or a isdefined(Main, :Revise) in the __init__ file, and then having DataFrames add the files "manually" with Revise.track(DataFrames, filename).

Revise.jl is already hooking into `using` to detect that a package is loaded. Is there a way to do the same with `include`?

timholy:
This is based on ~~the not-yet-merged~~ https://github.com/JuliaLang/julia/pull/23898. Locally it passes all tests ~~until it gets to the `Revise.track` tests, because there I haven't expunged the dependence on parsing. I'm posting this now because there seems to be room to discuss how this should work:~~

- ~~`Revise.track(Base)`: perhaps the best approach here would be for `base/sysimg.jl` to keep track of the files it includes, and Revise could just look that information up.~~
- ~~More interesting is `Revise.track(filename)`, which I'm thinking of in the context of "scripts" for test or analysis code. Currently we recursively track any file `included` by that file, based on parsing. If we wanted to replicate that, note that the new `Revise.included_files` list includes the entire set, once `filename` has been included once. The trouble is that if the user has included several files, there doesn't seem to be a great way to know which files got recursively included by which operation. One approach would be to temporarily empty the list and re-include `filename`, but that has some obvious downsides. Overall, here I favor restricting it to be literal, and require users to list each file they want to track.~~

Once merged this will fix #40.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
   - nightly
 notifications:
   email: false
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Revise
 
+**NOTE: this page is for Julia 0.7-DEV and higher. For Julia 0.6 [see this branch](https://github.com/timholy/Revise.jl/tree/v0.6)**
+
 [![Build Status](https://travis-ci.org/timholy/Revise.jl.svg?branch=master)](https://travis-ci.org/timholy/Revise.jl)
 [![Build status](https://ci.appveyor.com/api/projects/status/e1xnsj4e5q9308y6/branch/master?svg=true)](https://ci.appveyor.com/project/timholy/revise-jl/branch/master)
 [![codecov.io](http://codecov.io/github/timholy/Revise.jl/coverage.svg?branch=master)](http://codecov.io/github/timholy/Revise.jl?branch=master)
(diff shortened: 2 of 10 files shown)


cstjean:
How will this interact with `@require`? AFAICT `@require A include("foo.jl")` will correctly track `foo.jl`, but `@require A function foo() ... end` will not track `foo`. Is that right?

timholy:
Not sure, to be honest.

FYI once I finish closing a couple of other issues, I'm planning to tag a v0.1.0 release, fork a 0.6 branch, and then make master 0.7-only. https://github.com/JuliaLang/julia/pull/23898 makes Revise work so much more reliably, I think we really should just focus on making Revise as good as it can be for Julia 1.0.

When tagged, the new one will be (at least) v0.2.0, so that if needed we can continue to fix 0.6 on that 0.6 branch.

Seem OK to you?

cstjean:
Sure, I will most likely do the same with TraceCalls.

timholy:
Great. Just tagged 0.1.

I may have just figured out in principle how to handle deleted specializations. For me I think that's the main reason for Revise to "fail", so I'm tempted to see if I can get that in before the fork. Won't be quick, though, I have other deadlines (the PRs were a "study break" :smile).

timholy:
This should work on sufficiently-recent master, but until CI passes I'm reluctant to merge since I can only test Linux. https://github.com/JuliaLang/julia/issues/23846

PR #57, opened 2017-10-21
Support method deletion

### LINKED ISSUE #46: handling stale functions
When changing a source file, sometimes you change the signature of functions. Revise defines the new functions, but the old functions still pollute the run-time. Many times, the source still includes calls to the old function signature (because nobody remembers to change all calls), so they silently run. Having just encountered such a potential pitfall, I'm putting it out here.

In any case, if functions from a Module / file could be identified in the run-time, then all functions from a "revised" file about to be run, can be redefined to indicate them being stale. And then the redefinitions would overwrite them again.

This problem doesn't necessarily have a simple solution, or one which should be under Revise.jl.

timholy:
This fixes #46. This is built on top of #49 (not yet merged), so only the last commit will be in the final version of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,14 +99,14 @@ doing it manually.
 
 To accomplish this, Revise uses the following overall strategy:
 
-- add a callback to Base so that Revise gets notified when new
+- add callbacks to Base so that Revise gets notified when new
   packages are loaded or new files `include`d
 - prepare source-code caches for every new file. These caches
   will allow Revise to detect changes when files are updated. For precompiled
   packages this happens on an as-needed basis, using the cached
   source in the `*.ji` file. For non-precompiled packages, Revise parses
-  the source for each `include`d file immediately so that the "starting point" is
-  known before you start saving changes.
+  the source for each `include`d file immediately so that the initial state is
+  known and changes can be detected.
 - monitor the file system for changes to any of the dependent files;
   it immediately appends any updates to a list of file names that need future
   processing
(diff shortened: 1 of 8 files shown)


vtjnash:
Corrupting the system in this way isn't a good idea. Inference will start spitting out assertion failures that the world bounds are wrong and, in a debug build, gf.c will start aborting when it detects the world bounds are inconsistent.

timholy:
Could we update the world age of the affected methods?

vtjnash:
Probably. It just needs to be implemented in base first.

timholy:
> It just needs to be implemented in base first.

In what sense? (Meaning, "what does the following miss?")
```julia
julia> foo(x) = 1
foo (generic function with 1 method)

julia> foo(3)
1

julia> m = first(methods(foo))
foo(x) in Main at REPL[1]:1

julia> m.specializations
TypeMapEntry(nothing, Tuple{typeof(foo),Int64}, nothing, svec(), 22822, -1, MethodInstance for foo(::Int64), true, true, false)

julia> m.specializations.min_world = 42
42

julia> m.specializations
TypeMapEntry(nothing, Tuple{typeof(foo),Int64}, nothing, svec(), 42, -1, MethodInstance for foo(::Int64), true, true, false)
```

timholy:
Now that I think about it a little bit more, I'm confused. Presumably you noted that I [removed cached specializations](https://github.com/timholy/Revise.jl/blob/c6046e0c11a2c99c6e73245d39da647d858a9a17/src/delete_method.jl#L45-L79) that depend (even indirectly) on the method being deleted.  Why doesn't that solve the problem? My thought is that this will then force the system to re-infer and recompile those methods using "current" code. I would have thought that would fix the world age problem.

One thing I haven't fixed is the `TypeMapLevel`. Jeff's explanation on discourse was very helpful but I just haven't had the time to implement it. Current code here is a debugging stub trying to figure out what specifically needs doing.

I'd be happy to contribute `delete_method(m::Method)` and dependent code to Base, but I think it makes sense to get it working here first and then move it (since it's marginally easier to work outside of Base). I just offered you push access in case you feel like playing with it yourself. But guiding me would be very helpful, too.

vtjnash:
There's a number of tables that a MethodInstance can get linked from. Some of them can only be accessed from C and they should only be accessed while holding a lock on the backedges array. I think the ambiguity list also may need to be recomputed when a method get removed, since we accelerate it currently by pre-filtering entries.

vtjnash:
And there's a few other places in src (such as jl_specializations_get_linfo) where we assume that Methods can't be deleted, that may need to be identified and updated. I suppose the primary question to be answered is whether Method deletion should also imply the method _can't_ be used, or simply isn't found in the MethodTable anymore. In theory, I think the former is more correct, but in practice, we're often doing the later, since it is a bit more symmetric with other usages of world age.

timholy:
Nightly builds are still to old to merge this.

timholy:
This apparently just needs to wait for the new Compat 0.51 release and this is good to merge (OSX nightly is too old for `lastindex`).

pablosanjose:
Amazing. Just amazing. It works! :-D

PR #59, opened 2017-11-16
Fix some warnings on julia 0.7-DEV

garrison:


timholy:
Thanks. If you're on 0.7, I highly recommend #49. If you're daring and willing to encounter bugs, perhaps even #57.

PR #61, opened 2017-11-20
More 0.7 Compat

Keno:


cstjean:
Why?

Keno:
Good question. I've seen a bunch of errors where Revise gets confused in loading itself. However, trying to reproduce that circumstance now doesn't reproduce the problem, so I'll get rid of that commit. I also see that one of the compat definitions was already added, so I'll adjust this PR accordingly also.

timholy:
Oh hooray, this revealed that nightlies have caught up on all platforms other than Windows. Are people willing to have a short-term test "hole" on Windows? If so, I'm going to close this and merge an updated #49.

timholy:
Oh wow, the windows nightly is 0.6.1! That seems like a mistake.

timholy:
Obviated by #49.

PR #65, opened 2017-11-27
scan for the right .ji file containing the target uuid

vtjnash:


timholy:
Just to check, the main change here is checking that the UUID matches that provided by `Base.module_uuid`. Seems like a good thing to do. Once it gets through CI, merge at will.

vtjnash:
Yeah (much easier to see if you add `?w=1` to the url

timholy:
Sounds good. Merge whenever you're ready.

timholy:
Is this ready to merge?

vtjnash:
Yes

timholy:
Thanks!

PR #74, opened 2018-01-21
fix Base.find_all_in_cache_path call after 0.7-3235

iblis17:
Ref: https://github.com/JuliaLang/julia/pull/25358/files#diff-a03d5ce5729a81495000698e643fce27L107


timholy:
By the time I got to this, the interface had changed again. I incorporated this into #79 (and gave you credit), hope you don't mind.

iblis17:
👌

PR #75, opened 2018-01-21
implement revise_file_queued for FreeBSD

### LINKED ISSUE #66: Behaviour of watch_file on FreeBSD is different
https://github.com/timholy/Revise.jl/blob/5754f39e17133b384913a7d05a44f377fcc174a6/src/Revise.jl#L188-L200

On FreeBSD modifying a existed file which located at `dir` will not being caught by `watch_dir(dir)`.

iblis17:
fix #66 

---

Basic functionality works for me, but I cannot get test passed :p

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -241,6 +241,10 @@ function process_parsed_files(files)
         haskey(watched_files, dir) || (watched_files[dir] = WatchList())
         push!(watched_files[dir], basename)
         push!(udirs, dir)
+
+        @static if Sys.KERNEL == :FreeBSD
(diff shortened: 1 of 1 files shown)


timholy:
Now that I have tests passing on nightly, is there a plausible way to run this through CI?

iblis17:
I rebased and ran it at local. It passed!

timholy:
That's great. I can merge this as-is, and if you want we can think about whether it's possible to set up FreeBSD CI testing for Revise.

timholy:
Many thanks for fixing this!

PR #79, opened 2018-01-31
Get tests passing on recent 0.7

timholy:
This incorporates #74 (this interface has changed multiple times now, and some of the nightly builds are behind).

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -12,11 +12,11 @@ else
     throwing_function(bt) = bt[1]
 end
 
-const rseed = Ref(Base.GLOBAL_RNG)  # to get new random directories (see #24445)
+const rseed = Ref(Random.GLOBAL_RNG)  # to get new random directories (see #24445)
 function randtmp()
     srand(rseed[])
     dirname = joinpath(tempdir(), randstring(10))
-    rseed[] = Base.GLOBAL_RNG
+    rseed[] = Random.GLOBAL_RNG
     dirname
 end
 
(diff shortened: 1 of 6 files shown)


timholy:
The Windows nightly seems unfixable because of https://github.com/JuliaLang/julia/issues/25279.

test/runtests.jl
```diff
@@ -12,6 +12,14 @@ else
     throwing_function(bt) = bt[1]
 end
 
+const rseed = Ref(Base.GLOBAL_RNG)  # to get new random directories (see #24445)
+function randtmp()
+    srand(rseed[])
+    dirname = joinpath(tempdir(), randstring(10))
+    rseed[] = Base.GLOBAL_RNG
```
rfourquet: I don't understand the point of this line: isn't rseed[] already equal to Base.GLOBAL_RNG?

test/runtests.jl
```diff
@@ -12,6 +12,14 @@ else
     throwing_function(bt) = bt[1]
 end
 
+const rseed = Ref(Base.GLOBAL_RNG)  # to get new random directories (see #24445)
```
rfourquet: Any reason not to use a `RandomDevice` object here? it doesn't need seeding, and is used internally in a call to `srand(::MersenneTwister)`. Actually, `randtmp() = joinpath(tempdir(), randstring(RandomDevice(), 10))` should work (unless called too often and too quickly, as `RandomDevice()` opens a file ("/dev/urandom") which is not freed immediately).

test/runtests.jl
timholy: I can't remember anymore.

test/runtests.jl
timholy: Didn't know about RandomDevice, that seems useful.

PR #80, opened 2018-01-31
Fix some deprecations

timholy:
I'll add more commits, but just tackling the `warn` deprecation is giving me some trouble. @c42f, any tips? It's the final testset that's causing trouble, somehow the message are appearing in the console instead of being captured to the file. Note that higher up in `runtests` the same strategy seems to work, so it's weird...

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -6,6 +6,7 @@ VERSION >= v"0.7.0-DEV.2359" && using FileWatching
 VERSION >= v"0.7.0-DEV.2575" && using Dates
 using Compat
 using Compat.REPL
+using Compat: @warn
 
 if VERSION < v"0.7.0-DEV.3483"
     register_root_module(m::Module) = Base.register_root_module(Base.module_name(m), m)
@@ -113,7 +114,7 @@ function eval_revised(revmd::ModDict)
                     eval(mod, ex)
                 end
             catch err
-                warn("failure to evaluate changes in ", mod)
+                @warn "failure to evaluate changes in $mod"
                 println(STDERR, ex)
             end
         end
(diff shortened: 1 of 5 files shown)


c42f:
I can't see anything obviously wrong with the source, I'll have to try this out for myself.

Just so you know, there's a `Test.@test_logs` macro in 0.7 for testing log records in a more precise way than string matching. I'm not sure it helps you here due to compatibility with 0.6 but it should be useful in the future.

timholy:
This is already 0.7-only. 0.6 is a separate branch and not getting more changes from me (though I'll merge reasonable PRs).

c42f:
I had a look into the logging issue.  The problem is that you're scheduling `revise_dir_queued` in a background task, but `with_logger` only captures logs from the current task (and any child tasks).

So, what to do about this... The easiest option is to install a global logger temporarily while you remove the directories, and hope that all the background tasks finish in the allotted time slot.  It's a bit messy.

An alternative is to call `track()` with a particularly task local logger installed so that the background task which is spawned to run `revise_dir_queued` inherits that logger.  Something like

```julia
logger1 = SimpleLogger(io)
with_logger(logger1) do
    Revise.track(srcfile1)
end
```

Now all logs pertaining to the async operations on `srcfile1` go to `logger1`, so you can disentangle them from the other invocations of `track()` if you want to.

c42f:
It's quite possible that the logging system should have further refinement in the way that it deals with tasks - I didn't have a lot of use cases to inform the design there.

timholy:
Thanks for checking. The large majority of uses are not from an explicit `track` but come from `using SomePkg`, many of which are in different `@testset`s. I'd rather not redirect the display of messages except in the places where I'm doing it now, as I fear it could make debugging harder (assuming that some messages would get swallowed).

timholy:
I seem to have fixed this in a separate effort, closing.

PR #84, opened 2018-02-06
Support distributed revision (fixes #82)

### LINKED ISSUE #82: Revising modules on multiple workers/processes
Hi, and thank you for Revise.jl.

Really there are two issues here.

The thing I really want is to have mypkg Revised on all of the Julia workers as well.  That doesn't happen with either of my below scenarios.  I have tried `@everywhere using mypkg` and `@everywhere using Revise` but no dice.    This makes for a long restart process.  Any suggestions on how to achieve this multi-process revise?

As far as process 1 is concerned, it appears that Revise will not update a local module when Julia is started with -p procs.  Interestingly Revise will notice updates when procs were added with addprocs().

Successful operation:
- Start `julia`
- run `using mypkg` 
- run `f()` from mypkg and get baseline output
- edit source code  to mypkg - add an `info("test1")`
- run `f()` from mypkg and see the new info line test1
- run `addprocs(2)` to add two worker processes
- edit source code  to mypkg - add an `info("test2")`
- run `f()` from mypkg and see the new info line test2

Error replication:
- Start `julia -p 2`
- run `using mypkg` 
- run `f()` from mypkg and get baseline output
- edit source code  to mypkg - add an `info("test1")`
- run `f()` from mypkg and fail to see the new info line test1

My Julia versioninfo, running on latest Arch:

```
Julia Version 0.6.2
Commit d386e40 (2017-12-13 18:08 UTC)
Platform Info:
	OS: Linux (x86_64-pc-linux-gnu)
	CPU: AMD Opteron(TM) Processor 6276
	WORD_SIZE: 64
	BLAS: libopenblas (DYNAMIC_ARCH NO_AFFINITY Bulldozer)
	LAPACK: libopenblas
	LIBM: libm
	LLVM: libLLVM-3.9.1 (ORCJIT, bdver1)
```

timholy:
One caveat is that this doesn't yet support method deletion in workers.
That seems a bit harder, and might require splitting parts of Revise into a sub-package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ If you're using polling, you may have to wait several seconds before changes tak
 
 Revise can handle many kinds of changes to Julia code, but a few may require special treatment:
 
-### Method deletion
+#### Method deletion
 
 Sometimes you might wish to change a method's type signature or number of arguments,
 or remove a method specialized for specific types.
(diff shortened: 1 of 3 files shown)


timholy:
Closing in favor of #85

PR #87, opened 2018-03-07
STDERR -> stderr

fredrikekre:


timholy:
Thanks for this! Sorry it took me so long to notice.

I merged #92 and then re-ran the tests, LGTM.

Should we make this depend on `Compat v0.55` just for safety? In case anyone is running an old version of master.

timholy:
Included in #94. Thanks!

PR #90, opened 2018-04-13
Fix Juno integration (master)

pfitzseb:
https://github.com/timholy/Revise.jl/pull/89


cstjean:
Am I missing something, or is this a duplicate PR?

pfitzseb:
Yeah, but against `master` and not `v0.6`.

PR #103, opened 2018-06-21
Add user callbacks (fixes #93)

### LINKED ISSUE #93: Callbacks for changes [feature suggestion]
It would be useful to allow registering callbacks so that other parts of the app would be notified of changes. This would allow the reuse of Revise's functionality for other features, without the need to set up extra file watchers.

timholy:
This allows one to recalculate a result whenever the underlying code changes. CC @essenciary


timholy:
I'll close this for reasons discussed in #203.

PR #106, opened 2018-06-29
return status of module reevaluation

gaika:


timholy:
Good idea. Can you add to the tests to make sure this doesn't break in the future?

timholy:
Thanks!

PR #108, opened 2018-07-03
Switch to OrderedCollections and eliminate ambiguities

timholy:
This cuts loose of DataStructures and switches to the more focused OrderedCollections. I also add a test for ambiguities (and stomp on those that still remained after the switch).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ git:
 
 ## uncomment the following lines to override the default test script
 script:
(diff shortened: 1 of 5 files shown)


timholy:
OK, bug (?) reported: https://github.com/JuliaLang/Pkg.jl/issues/465. To keep this useful for their purposes I will leave this PR open and start a new branch & PR that strips out the debugging components of this PR.

timholy:
Ref #111

PR #116, opened 2018-07-28
Redesign Revise for bidirectional `expr`<-->`method` capabilities

### LINKED ISSUE #51: Line numbers
One longstanding issue is the fact that if you load a package, and then add some code at line 100, code for later (unmodified) functions will show up in backtraces with the wrong line numbers. Aside from re-evaluating the entire module (#41), I'm not sure there's a good solution, but I've wondered a bit about whether there could be some extra layer of indirection to look up "living" line numbers. Since this is of relevance for debuggers too, CCing @Keno to see if he has any ideas.

timholy:
This is a major redesign of Revise's internals. It is now organized around two "maps" (Dicts), one which connects expressions to Julia types and the other which goes in the opposite direction. Together the two can be leveraged for an accurate and efficient connection between methods and the source-expressions that define them.

One nice consequence of this change is that we can finally fix #51. It depends on some cooperation from Base, however: https://github.com/JuliaLang/julia/pull/28328.

### DIFF
--- a/docs/src/internals.md
+++ b/docs/src/internals.md
@@ -1,7 +1,8 @@
 # How Revise works
 
 Revise is based on the fact that you can change functions even when
-they are defined in other modules. Here's an example showing how you do that manually (without using Revise):
+they are defined in other modules.
+Here's an example showing how you do that manually (without using Revise):
 
 ```julia
 julia> convert(Float64, π)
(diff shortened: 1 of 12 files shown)


docs/src/internals.md
```diff
@@ -50,4 +51,195 @@ To accomplish this, Revise uses the following overall strategy:
...
+                     :(indent(::UInt8) = 4)  => ([Tuple{typeof(indent),UInt8}], 0)
+                     ),
+              SigtMap(Tuple{typeof(indent),UInt16} => :(indent(::UInt16) = 2),
+                      Tuple{typeof(indent),UInt8}  => :(indent(::UInt8) = 4)
+                      ))
+```
+The `lineoffset`s are initially set to 0 when the code is first compiled, but these
+may be updated if the source file is changed.
+
+`Items.jl` is reprsented with a bit more complexity,
```
tkoolen: 'represented' (just browsing)

timholy: Thanks!

timholy:
Test failure is https://github.com/JuliaLang/julia/issues/28384, works locally with https://github.com/JuliaLang/julia/pull/28386 and in any case is independent of this PR.

PR #119, opened 2018-08-01
Remove __precompile__, as it is now the default.

tpapp:
Elimitates the warning message after https://github.com/JuliaLang/julia/pull/26991


timholy:
Working on it. Note that tests fail, though not due to this PR. See https://github.com/JuliaLang/julia/issues/28384.

PR #123, opened 2018-08-03
fixed document instructions on starting Revise by default

ExpandingMan:
Currently there is no function as `Revise.async_steal_repl_backend()`, it seems that what was intended was `@async Revise.steal_repl_backend()`.  Was there an intention of adding a function istead?


timholy:
The docs are for 0.6, which is not released yet. But it should be any hour now, see https://github.com/JuliaLang/METADATA.jl/pull/16263.

PR #126, opened 2018-08-07
Add includet to simplify script-tracking

timholy:
Plus another pile of fixes. The new signature parsing places many more demands on accuracy, it seems, so the v0.6 rewrite has led to several problems. However, in several respects it's already better than ever, so I am pretty confident this is moving in the right direction.

### DIFF
--- a/docs/src/user_reference.md
+++ b/docs/src/user_reference.md
@@ -1,12 +1,13 @@
 # User reference
 
-There are really only two functions that a user would be expected to call manually:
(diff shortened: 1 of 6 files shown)


timholy:
One glitch on windows and the usual test failures on OSX. In other words, looks fine.

(Would be great if an OSX user would chime in and tell me whether tests pass locally: I can't tell if this is Travis-specific or not.)

PR #127, opened 2018-08-07
Greatly improve performance with IdDict

timholy:
Motivated by the slowness of `track(LinearAlgebra)`, I profiled and found a surprising result: almost all the time was spent on type-inference for inserting `sigtmap[sigt] = def`. Given that each key has a different type, perhaps this shouldn't have been surprising. By switching to an IdDict it got faster.

Before:
```julia
julia> using LinearAlgebra

julia> @time Revise.track(LinearAlgebra)
 89.868081 seconds (233.75 M allocations: 11.436 GiB, 3.34% gc time)
```
After:
```julia
julia> @time Revise.track(LinearAlgebra)
  3.469887 seconds (5.75 M allocations: 214.248 MiB, 4.76% gc time)
```

CC @andreasnoack

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -371,7 +371,7 @@ function instantiate_sigs!(fmm::FMMaps, def::RelocatableExpr, sig::RelocatableEx
     try
         sigts = Any[Core.eval(mod, s) for s in sigtexs]
     catch err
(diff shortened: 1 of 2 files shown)


andreasnoack:
That is a really good speedup.

PR #132, opened 2018-08-08
Update build.jl

raminammour:
For non super users (like myself) you get:
```
 sysctl
Absolute path to 'sysctl' is '/usr/sbin/sysctl', so running it may require superuser privileges (eg. root).
```
And thus the build fails, this achieves the same end result, and circumvents the error... I hope :)


timholy:
Wonderful, thanks for doing this!

PR #133, opened 2018-08-09
More robustly find `base` (and the stdlibs)

timholy:
This is designed to support Revise.track(Base) on a wider array of machines. This is becoming important because of [Rebugger](https://github.com/timholy/Rebugger.jl). This may fix https://github.com/timholy/Rebugger.jl/issues/1 ~and the longstanding Travis failures on OSX.~ (EDIT: those were fixed in #136)

### DIFF
--- a/docs/src/dev_reference.md
+++ b/docs/src/dev_reference.md
@@ -17,6 +17,7 @@ Revise.tracking_Main_includes
 ```@docs
 Revise.juliadir
 Revise.basesrccache
+Revise.basebuilddir
 ```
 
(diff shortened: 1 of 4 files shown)


timholy:
OK, this is weird: as you can see above, the tests pass on OSX (finally!). But only in the presence of that debugging commit, and that's really weird because as you can see for yourself all it does is add some `@show` statements. Does anyone understand this? CC @andreasnoack.

timholy:
Yay! Finally. :relieved:

PR #136, opened 2018-08-09
Fix test failures on OSX

timholy:
This makes no sense, but it reproducibly fixes the test failures. Thanks to @andreasnoack and @vtjnash.

Andreas also mentioned that on OSX there are times where methods don't update until you insert a `print` statement (which yields) into the method body. This suggests that we may need to insert more `yield`s in more places. But it makes sense to go stepwise, so let's start here.


timholy:
Back to green on all platforms, yay! Thanks so much for helping out.

PR #151, opened 2018-08-14
Document the steps needed to fix path problems

timholy:
As Revise gets more ambitious about tracking more and more of Julia, it seems likely that path-related problems will be the number one source of trouble for users. Rather than just respond to issues as they come up, let's have some documentation that walks through the process of diagnosing and fixing path problems.


timholy:
CC @jlapeyre

PR #154, opened 2018-08-15
Parse properly when macros redefine methods

### LINKED ISSUE #148: StatsPlots
We get an error parsing a `@recipe`

timholy:
Boy, this was really dumb. I was implicitly assuming that macros don't modify the method signature. :disappointed_relieved: 

Fixes #148 . CC @mkborregaard, @daschw.


mkborregaard:
Well, in that case it was certainly dumb at a rather higher level than I'm usually dumb.
Anyway, I can confirm that checking out this branch resolves the issue.

PR #160, opened 2018-08-16
Make get_def more modular

timholy:
This is work that's designed to support Rebugger getting a better handle on "generated" kwarg methods.


timholy:
Just a random error in cleanup.

PR #162, opened 2018-08-19
Remove the build script and issue a warning instead

timholy:


timholy:
In case of trouble, here's a breadcrumb:
```
(File Monitor): no space left on device (ENOSPC)
```
If you're getting that error, it's possible I got the error type or code wrong, see https://github.com/timholy/Revise.jl/blob/99d250a00ff13094ee0838f3726e9f4f40a01083/src/Revise.jl#L32-L46.

PR #163, opened 2018-08-19
Add a logging framework

timholy:
This may help catch weird/rare errors. CC @pfitzseb, @vtjnash.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -11,6 +11,7 @@ makedocs(
         "Home" => "index.md",
         "config.md",
         "limitations.md",
+        "debugging.md",
         "internals.md",
         "user_reference.md",
         "dev_reference.md",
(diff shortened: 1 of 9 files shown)


vtjnash:
+1

Would it also be useful to sprinkle some `@debug` statements around? That allows someone to turn this on via the ENV (JULIA_DEBUG=Revise)

timholy:
OK, I completely restructured this around Base's logging framework---not sure why I rolled my own when that was available.

However, I am not sure how to usefully turn on logging using the `JULIA_DEBUG` command. As you'll see if you inspect this, I exploit the `Test.TestLogger` and that's not the default. In the docs I show a more cumbersome way to turn it on, but if you have any other suggestions I'm all ears. (I am new to the logging system.)

vtjnash:
Yeah, that’s a known problem (https://github.com/JuliaLang/julia/issues/25404). You can also control that variable at runtime (via ENV)

timholy:
Passed on Appveyor previously, but my AV queue is long (all the Images packages are running).

PR #170, opened 2018-08-26
Optionally return the list of modified files from Revise.track(Base)

### LINKED ISSUE #168: Make Revise.track(Base) fast for everyone
With Rebugger, presumably it's now much more common for people to trigger `Revise.track(Base)` simply by stepping into Base methods. Consequently, it's become a major negative to have `Revise.track(Base)` be slow. On my machine it's not too bad:
```julia
julia> @time Revise.track(Base); @time revise()
  0.335863 seconds (761.32 k allocations: 38.150 MiB, 2.98% gc time)
  0.942074 seconds (955.82 k allocations: 43.453 MiB, 1.04% gc time)
```
However, https://github.com/timholy/Revise.jl/issues/161#issuecomment-414616409 reported times on the order of ~15s, which really isn't acceptable.

The timing will be greatly affected by the amount of work that needs to be done, and this comes down to the `mtime` criterion. On my machine, I happen to have one file (`"InteractiveUtils.jl"`) that I must have touched at some point after building Julia:
```julia
julia> rlogger = Revise.debug_logger()
Revise.ReviseLogger(Revise.LogRecord[], Debug)

julia> Revise.track(Base)

julia> filter(r->r.group == "Watching", rlogger.logs)
1-element Array{Revise.LogRecord,1}:
 Revise.LogRecord(Debug, Recipe for Base, Watching, Revise_61343d83, "/home/tim/.julia/dev/Revise/src/recipes.jl", 26, (filename=/home/tim/src/julia-1.0/usr/share/julia/stdlib/v1.0/InteractiveUtils/src/InteractiveUtils.jl, mtime=1.5344475727586915e9, mtimeref=1.5337688806937075e9))
```
My current working hypothesis is that it's fairly fast on my machine because it only needs to examine one file, and that it's slow on some other machines because they must have many more files that trigger the `mtime` criterion.

If I had to guess, I suspect a major division might be between people who build from source and people who download binaries:

- if you build from source, you first have the files sitting on your hard drive and then you build julia. Consequently the build times in `base.cache` should be newer than the `mtime` on the file.
- if you download a binary, the build time in `base.cache` is from the buildbot, and the `mtime` for the files corresponds to the time Julia was downloaded and installed on the user's machine. Hence every file in Base has a fresher `mtime` which forces parsing by Revise.

There's just one hitch with that theory: if I download a binary on Linux, it's really fast:
```julia
julia> @time Revise.track(Base); @time revise()
  0.206909 seconds (734.45 k allocations: 36.860 MiB, 2.92% gc time)
  0.030500 seconds (21.35 k allocations: 1022.261 KiB)
```
So this theory requires an additional element: perhaps on Linux systems, the `mtime` is preserved in the download. Indeed:
```sh
tim@diva:~/bin/julia-1.0.0/share/julia/base$ ls -l
total 2120
-rw-r--r-- 1 tim holy 67233 Aug  8 20:27 abstractarray.jl
-rw-r--r-- 1 tim holy 10068 Aug  8 20:27 abstractarraymath.jl
...
```
which was the date of Julia's release (and I downloaded the binary weeks later).
However, perhaps on Windows systems that's not true?

Anyone want to test? (@pfitzseb?) This is a useful patch:

```diff
diff --git a/src/recipes.jl b/src/recipes.jl
index 53d53e4..9c15552 100644
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -21,6 +21,7 @@ function track(mod::Module)
             filename = fixpath(filename)
             push!(fileinfos, filename=>FileInfo(submod, basesrccache))
             push!(files, filename)
+            @show filename mtime(filename)-mtcache
             if mtime(filename) > mtcache
                 with_logger(_debug_logger) do
                     @debug "Recipe for Base" _group="Watching" filename=filename mtime=mtime(filename) mtimeref=mtcache
```
Any time difference that shows as positive will trigger parsing.

If this guess is correct (and it may not be), I see two potential paths forward:

- modify Julia so that `base.cache` also includes an MD5SUM or similar. Presumably this would be much faster to compute. Obviously this can't happen overnight (1.x feature?)
- write a file `Revise/deps/base.mtimes` that records an updated `mtime`, perhaps the most recent time at which the diff was computed and no difference was found.

While the latter feels more "hacky," it doesn't require modifications to Julia and would probably be even faster. So I find myself tempted by the latter.

timholy:
This should fix #168 by ~~basically having Revise ignore changes to Base files unless the user has a source-build of Julia. Since we'd like to have them commit their improvements back to Base anyway, this is just incentive for good behavior :smile:.~~
allowing the caller to choose whether to process the files that trigger the `mtime` criterion.

@pfitzseb, does this fix #168 for you?

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -592,7 +592,7 @@ Return the RelocatableExpr defining `method`.
 The source-file defining `method` must be tracked.
 If it is in Base, this will execute `track(Base)` if necessary.
 """
-function get_def(method::Method)
+function get_def(method::Method; modified_files=revision_queue)
     filename = String(method.file)
     yield()   # magic bug fix for the OSX test failures. TODO: figure out why this works (prob. Julia bug)
     startswith(filename, "REPL") && error("methods defined at the REPL are not yet supported")
@@ -623,7 +623,7 @@ function get_def(method::Method)
     filename isa AbstractString || return nothing
     if !haskey(fileinfos, filename)
         @info "tracking $recipemod"
-        track(recipemod)
+        track(recipemod; modified_files=modified_files)
     end
     fi = fileinfos[filename]
     maybe_parse_from_cache!(fi, filename)
--- a/src/exprutils.jl
+++ b/src/exprutils.jl
@@ -75,7 +75,7 @@ julia> Revise.get_signature(quote
 """
 function get_signature(ex::E) where E <: ExLike
     if ex.head == :macrocall
-        error("macros must be handled externally")
+        error("macros must be handled externally, got $ex")
     end
     if is_trivial_block_wrapper(ex)
         return get_signature(ex.args[end])
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -6,7 +6,7 @@
 Track updates to the code in Julia's `base` directory, `base/compiler`, or one of its
 standard libraries.
 """
-function track(mod::Module)
+function track(mod::Module; modified_files=revision_queue)
     if mod == Base
         # Test whether we know where to find the files
         if !isdir(joinpath(juliadir, "base"))
@@ -25,18 +25,18 @@ function track(mod::Module)
                 with_logger(_debug_logger) do
                     @debug "Recipe for Base" _group="Watching" filename=filename mtime=mtime(filename) mtimeref=mtcache
                 end
-                push!(revision_queue, filename)
+                push!(modified_files, filename)
             end
         end
         # Add the files to the watch list
         init_watching(files)
     elseif mod == Core.Compiler
         compilerdir = joinpath(juliadir, "base", "compiler")
-        track_subdir_from_git(Core.Compiler, compilerdir)
+        track_subdir_from_git(Core.Compiler, compilerdir; modified_files=modified_files)
     elseif nameof(mod) ∈ stdlib_names
         stdlibdir = joinpath(juliadir, "stdlib")
         libdir = joinpath(stdlibdir, String(nameof(mod)), "src")
-        track_subdir_from_git(mod, normpath(libdir))
+        track_subdir_from_git(mod, normpath(libdir); modified_files=modified_files)
     else
         error("no Revise.track recipe for module ", mod)
     end
@@ -60,7 +60,7 @@ function fixpath(filename; badpath=basebuilddir, goodpath=juliadir)
 end
 
 # For tracking subdirectories of Julia itself (base/compiler, stdlibs)
-function track_subdir_from_git(mod::Module, subdir::AbstractString; commit=Base.GIT_VERSION_INFO.commit)
+function track_subdir_from_git(mod::Module, subdir::AbstractString; commit=Base.GIT_VERSION_INFO.commit, modified_files=revision_queue)
     # diff against files at the same commit used to build Julia
     repo, repo_path = git_repo(subdir)
     if repo == nothing
@@ -84,7 +84,7 @@ function track_subdir_from_git(mod::Module, subdir::AbstractString; commit=Base.
             rethrow(err)
         end
         if src != read(fullpath, String)
-            push!(revision_queue, fullpath)
+            push!(modified_files, fullpath)
         end
         fmod = get(juliaf2m, fullpath, Core.Compiler)  # Core.Compiler is not cached
         fi = FileInfo(fmod)


pfitzseb:
It does indeed:
```
julia> @time Revise.track(Base); @time Revise.revise()
  0.179792 seconds (769.17 k allocations: 38.294 MiB, 5.98% gc time)
  0.025312 seconds (23.42 k allocations: 1.256 MiB)
```

---
I think I'd much prefer 
```diff
diff --git a/src/recipes.jl b/src/recipes.jl
index 805e4e4..f8c3256 100644
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -27,7 +27,7 @@ function track(mod::Module)
             filename = fixpath(filename)
             push!(fileinfos, filename=>FileInfo(submod, basesrccache))
             push!(files, filename)
-            if issourcebuild && mtime(filename) > mtcache
+            if mtime(filename) > mtcache + (issourcebuild ? 0 : 600)
                 with_logger(_debug_logger) do
                     @debug "Recipe for Base" _group="Watching" filename=filename mtime=mtime(filename) mtimeref=mtcache
                 end
@@ -35,7 +35,7 @@ function track(mod::Module)
             end
         end
         # Add the files to the watch list
-        issourcebuild && init_watching(files)
+        init_watching(files)
     elseif mod == Core.Compiler
         compilerdir = joinpath(juliadir, "base", "compiler")
         track_subdir_from_git(Core.Compiler, compilerdir)

```
or something similar, because there's no really compelling reason to completly disable this imho.

timholy:
I should have asked first: in practice, do you edit Base files using a binary build? If the answer is "yes" then I should reconsider. If the answer is "no" then to me it just seems safer to have the behavior be completely predictable, rather than based on some constant that we pick today and might be argued about in the future.

timholy:
I suppose another option would be to add a keyword argument, e.g., Rebugger would use `Revise.track(Base; watch_for_changes=false)` or something. Might take a little thinking about how to handle if it if the user explicitly says `Revise.track(Base)` later in the same session.

pfitzseb:
> in practice, do you edit Base files using a binary build?

On Windows, yes (because it's annoying to build from source there), but for anything more than a quick two line change I'd use a source build on Linux. So I personally wouldn't miss this all that much (and for small changes I could just use Juno anyways).

It just feels like a shame to basically cut this functionality.

> rather than based on some constant that we pick today and might be argued about in the future.

Another alternative would be a user-adjustable offset (`Revise.mtimeoffset = Ref(0)`). That makes it slightly nicer to use Revise for Base, but keeps the same default behaviour.

timholy:
I agree we could make it adjustable. But then the docs need to have a whole section about "if Revise.track(Base) is slow for you, here's the procedure for tuning..."

Is there anything obviously wrong with the keyword arg solution? (https://github.com/timholy/Revise.jl/pull/170#issuecomment-416033178) That seems less likely to fail due to vagaries of a subset of user's systems.

timholy:
OK, I simply made it possible for the caller to intercept the list of files that trigger the `mtime` criterion. Here are the implications:

- if you manually call `Revise.track(Base)` from your session before using Rebugger, it will behave as usual.
- after this merges and is tagged, I'll change Rebugger so that it simply drops the list of modified files. This is actually more correct than prior behavior, because up to now you could get into a situation in which (1) the user has local modifications but hasn't issued `Revise.track(Base)` so until they Rebug something the code they are running is the old code, but then (2) as soon as they step in they are running the new code because Rebugger called `track(Base)`. That could be really confusing, so it's actually an improvement to have Rebugger "fail" to process earlier revisions. (Future revisions to Base files do not pose a problem because once the files are being watched, both "ordinary" and Rebugged calls will trigger the new version.) 
- if you really want to make sure you always process "relevant" revisions to Base, in your `config/startup.jl` file insert some lines like the following:

```julia
modfiles = String[]
Revise.track(Base; modified_files=modfiles)
recheck(modfiles, mtime(Revise.basesrccache))
```
where `recheck` is a function you write using whatever `mtime` adjustment you want, and which just pushes any files that trigger the criterion onto `Revise.revision_queue`.

This design lets the user do whatever s/he wants without introducing yet more tuning parameters into Revise. As is probably obvious, I really dislike tuning parameters that have an element of magic to them, and this issue is pure magic: any slight change to the way Julia's Windows binary builder works, and "poof" all prior settings become irrelevant. But it doesn't prevent the user from issuing the incantations directly.

pfitzseb:
This seems like a very clean solution, I like it! :)

PR #188, opened 2018-09-07
Change recommendations on startup.jl

### LINKED ISSUE #184: how to `steal_repl_backend` using `atreplinit`
Doing `using Revise` causes latency in scripts that I'm running outside of the repl.  I can call `using Revise` with `atreplinit` and that seems to work fine, however I can't seem to do this with `steal_repl_backend`.  Is there a way to do this?  This seems like a common use case that should get documentation.

On second thought, after having looked into this a bit more it seems like this *may* not even be possible right now?

timholy:
Fixes #184. The code changes are designed to be backwards-compatible. However, I'd like to understand how to work around behavioral changes this will cause for IJulia.

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -6,12 +6,21 @@ If you like Revise, you can ensure that every Julia session uses it by
 adding the following to your `.julia/config/startup.jl` file:
 
 ```julia
+atreplinit() do repl
+    try
+        @eval using Revise
+        @async Revise.wait_steal_repl_backend()
+    catch
+    end
+end
+```
+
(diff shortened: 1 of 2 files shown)


ExpandingMan:
Sorry *this* is the one!  Is there anything blocking this? (Forgive my absent-mindedness, this is where I intended to ask that in the first place.)

timholy:
I need to submit an IJulia PR (not hard, I just haven't gotten to it yet). See https://github.com/JuliaLang/IJulia.jl/issues/738

ExpandingMan:
Awesome, that's been on my wish-list for a while.  Thanks for your tireless work on this great package!

PR #211, opened 2018-11-25
Create a logo

timholy:
I am considering adding Revise to the "Editors and ~~IDEs~~Development Environment" section of https://julialang.org/; for that to happen and for it to fit in with the others, Revise would need a logo. See https://github.com/timholy/Revise.jl/blob/teh/splash/images/logo.png if you want to see what this candidate logo looks like.

Thoughts? CC @ViralBShah, @StefanKarpinski.


ViralBShah:
The logo link does not work. I see there's a logo up there, but I usually ping @cormullion for logos. :-)

timholy:
Did not know about @cormullion's many talents!

timholy:
That branch got deleted, now you can see it on the home page https://github.com/timholy/Revise.jl

ViralBShah:
Yes! The DynamicalSystems.jl one is one of my favourties.

https://github.com/JuliaDynamics/DynamicalSystems.jl

PR #229, opened 2019-01-05
Allow Base.@​propagate_inbounds etc.

tkf:
It would be nice if Revise.jl notices `Base.@​propagate_inbounds` etc. are the "annotation" macro.  Is this the right approach?

### DIFF
--- a/src/exprutils.jl
+++ b/src/exprutils.jl
@@ -3,7 +3,10 @@
 using Core: MethodInstance
 using Base: MethodList
 
-const poppable_macro = (Symbol("@inline"), Symbol("@noinline"), Symbol("@propagate_inbounds"), Symbol("@eval"))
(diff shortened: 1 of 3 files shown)


src/exprutils.jl
```diff
@@ -3,7 +3,10 @@
 using Core: MethodInstance
 using Base: MethodList
 
-const poppable_macro = (Symbol("@inline"), Symbol("@noinline"), Symbol("@propagate_inbounds"), Symbol("@eval"))
+const poppable_macro = (Symbol("@inline"), Symbol("@noinline"), Symbol("@propagate_inbounds"), Symbol("@eval"), Symbol("@pure"))
+
+is_poppable_macro(ex) = ex ∈ poppable_macro ||
+    ex.head == :. && ex.args[1] == :Base && ex.args[2].value ∈ poppable_macro
```
timholy: What if `ex` isn't an expression? Add `if ex isa Expr &&` to the front of this line. Also add parenthesis to disambiguate the grouping of logical operators.

timholy:
Thanks!

PR #230, opened 2019-01-05
Major enhancements to parsing

timholy:
With this set of changes, Revise successfully extracts signatures from about 85% of Base's methods (including the imported and non-exported ones). This should make Rebugger work better.


timholy:
FYI I'm not planning on merging this immediately because I'm playing around with a much more ambitious alternative that in theory should get 100% of the signatures. But I think this branch can be used locally; I'm happy to answer questions or fix small issues if folks encounter them.

timholy:
Superseded by #243

PR #241, opened 2019-02-22
Smaller logo and center it

musm:
Currently the logo is so big it dominates the page, I've made it smaller and centered it in the readme.

Feel free to close if you prefer to current style!


timholy:
Much better, thank you!

PR #243, opened 2019-02-25
Rewrite based on lowered code

timholy:
This PR rewrites the package around JuliaInterpreter, which it uses to do much deeper analysis of the methods that get defined by code. A particularly noteworthy change is that it can now dive into methods defined via `@eval`, and it is finally able to handle kwarg functions robustly. The goal is to make Revise better at understanding the origin of methods and therefore better at deleting expunged ones and managing CodeTracking.jl's state for consumption by others.

This is a "developer preview" for upcoming changes. You need the following packages (probably in dev mode, and updated frequently):
- [JuliaInterpreter.jl](https://github.com/JuliaDebug/JuliaInterpreter.jl)
- [LoweredCodeUtils.jl](https://github.com/JuliaDebug/LoweredCodeUtils.jl)
- [CodeTracking.jl](https://github.com/timholy/CodeTracking.jl)

Automatic revision is currently disabled; just manually execute `revise()` whenever you want to update. (It's much easier to debug failures if they're not running in a separate task.)

Still to do:
- [x] add tests that check the fraction of extracted signatures from Base
- [x] restore and/or improve the catching of exceptions generated by Revise
- [x] restore automatic revision
- [ ] update the documentation
- [x] insert bounds in the registered version of Rebugger (this will break current versions)
- [x] work on startup and other latencies


timholy:
CC @KristofferC, @pfitzseb

src/lowered.jl
```diff
@@ -0,0 +1,173 @@
...
+end
+
+function methods_by_execution!(@nospecialize(recurse), methodinfo, docexprs, mod::Module, ex::Expr; define=true)
+    try
+        frame = prepare_thunk(mod, ex)
+        frame === nothing && return nothing
+        return methods_by_execution!(recurse, methodinfo, docexprs, frame; define=define)
+    catch err
+        isa(err, InterruptException) && rethrow(err)
+        @warn "error evaluating in module $mod: $ex"
```
oxinabox:
```suggestion
        @warn "error evaluating in module $mod" exception=ex
```

Loggers implement special handling/rendering for the `exception` kwarg.
Can probably also remove the `showerror` below, because of this.

timholy:
...and we have green tests. Removing the WIP, this is getting close to liftoff.

pfitzseb:
Not sue if this is to be expected, but I'm getting
```
julia> Revise.track("/home/pfitzseb/Documents/julia-1.1.0/share/julia/base/intfuncs.jl")
┌ Warning: error evaluating in module Main
│   exception = :(const HWNumber = Union{HWReal, Complex{<:HWReal}, Rational{<:HWReal}})
└ @ Revise ~/.julia/dev/Revise/src/lowered.jl:43
# loads more warnings
```

timholy:
Do you get the warnings with `Revise.track(Base)`? You shouldn't track individual files in Base, I am not sure what would happen (probably, lots of bad stuff).

pfitzseb:
I don't.

KristofferC:
I get the following warnings:

```
julia> using Revise; using Debugger

julia> let
           f_inv(x::Real) = x^2;
           f_inv(x::Integer) = 1 + invoke(f_inv, Tuple{Real}, x)
           @enter f_inv(2)
       end
┌ Warning: error evaluating in module Main
│   exception =
│    :(let
│          #= REPL[6]:2 =#
│          f_inv(x::Real) = begin
│                  #= REPL[6]:2 =#
│                  x ^ 2
│              end
│          #= REPL[6]:3 =#
│          f_inv(x::Integer) = begin
│                  #= REPL[6]:3 =#
│                  1 + invoke(f_inv, Tuple{Real}, x)
│              end
│          #= REPL[6]:4 =#
│          #= REPL[6]:4 =# @enter f_inv(2)
│      end)
```

timholy:
It's probably a world-age issue in that you're defining methods and then doing stuff with them. But commenting out the warning seems to let things run properly anyway...

Not quite sure what to do here.

timholy:
Oh wait, that warning had a mistake (from https://github.com/timholy/Revise.jl/pull/243#discussion_r266256208), it's probably leading us to a real bug...

cstjean:
Am I correct that this means that for a `__precompile__(false)` module, this means that all macros are expanded by the interpreter? I've been on Revise 0.7.14 for a long time, and Julia 1.2 forced me to upgrade, but now my package just won't load (or rather, `__init__` is called, but then it hangs for a very long time, seemingly proportional to the size of my macroexpansions).

timholy:
Sorry about the troubles. First, there is the Revise 1 branch, currently at 1.1.1. That does *not* use JuliaInterpreter.

Second, if building your package involves a lot of work that's done outside of an `__init__`, it might be slow. See #300, https://github.com/JuliaAstro/AstroBase.jl/pull/31. If you let me know which package you're working on I might be able to provide more insight.

cstjean:
It's a private package. I just filed an issue without seeing this comment, I'll try 1.1.1. Thank you for the help offer, and for all the work on Revise!

PR #245, opened 2019-03-02
Integrate with CodeTracking

timholy:
This sets up [CodeTracking.jl](https://github.com/timholy/CodeTracking.jl) as the new "query interface" of Revise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
   - osx
 julia:
   - 1.0
+  - 1.1
   - nightly
 notifications:
   email: false
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 1.0
 OrderedCollections
+CodeTracking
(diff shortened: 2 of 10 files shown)


oxinabox:
I am now using this branch and it is (so far) working for me. <3

timholy:
> It would be cool if you could push your CodeTracking.jl branch at the sametime you push to this branch. Alternatively I could just be patient.

There's a pretty complicated dance going on right now with this branch, #243, and work in JuliaInterpreter (e.g., https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/88). I fear you're going to have to expect some breakage for a little while yet.

Some more breaking changes are coming. I need `method_at(file, lineno)` for work in JuliaInterpreter, and that will likely change the internal representation that Revise and/or CodeTracking use.

timholy:
But it should be working as of this moment :smile:

timholy:
The most recent commit requires https://github.com/timholy/CodeTracking.jl/pull/11

PR #251, opened 2019-03-21
Disable breakpoints before executing code

timholy:
Fixes https://github.com/timholy/CodeTracking.jl/issues/27

CC @pfitzseb


pfitzseb:
Yep, can confirm this works.

timholy:
My AV queue is very long. I'll just merge.

Issue #203, opened 2018-10-15
entr-like functionality?


mlhetland:
A (hopefully small) feature request…

For quite some time, I've been using [`entr`](http://www.entrproject.org) as part of my coding workflow. For example, I have code generating TeX code, which is then compiled by `latexmk`, so all I do is save, and the PDF is updated. Or I use `entr` to run my tests, for example. Needless to say, for Julia, this can (with current compile times) sometimes lead to a bit of waiting. `Revise` would be ideal to cut down on that wait time, and I thought about setting up some wrapper/connection between a process (e.g., REPL) running with `Revise` and `entr` monitoring files, calling the appropriate code if something changes.

But … `Revise` is *already* monitoring files, of course, so this would probably be a huge waste! In general, I guess it's even monitoring the *right* files. So I guess my request boils down to the two following features, in order of descending priority:

1. **Registering a callback** or delegate, or some similar mechanism, to trigger specified user code after revision (à la `entr`); and
2. **Tracking non-Julia files** (e.g., in one of my use-cases, TeX-related files that are read and used as string constants at compile-time).

As the current version already uses strings for `track`, dispatch would be a bit of work there, but it could perhaps just be a keyword flag? E.g., have `julia=true` or something as the default?

timholy:
See #93 and #103. This is the third request for this functionality (CC @essenciary and @yurivish), so I don't doubt this would be useful. But no one has ever tested it and reported back if it's useful. I'm not convinced it is all that useful without changing the mode for looking for updates, which is currently dependent upon REPL input.

mlhetland:
Oh, sorry about that! I thought I looked through the issues, but missed these :-O No, I realized that it's not based on file changes. I guess maybe the functionality is a candidate for a separate library/tool, then!

mlhetland:
But, yeah, a callback isn't really useful in itself for me – what I was hoping to avoid was having to go into a REPL or the like and execute something every time the file changes. I guess I'll look into alternatives!

mlhetland:
For future seekers, this seems to do the trick, assuming the code you wish to run is in `main`:
```
julia> using Revise, FileWatching
julia> includet("myfile.jl")
julia> while true
           watch_file("myfile.jl")
           revise()
           main()
           sleep(.1)
       end
```
The `sleep` is to avoid triggering multiple watch events with a single change. (It seems that happens otherwise; there may be a more elegant solution to that, though.)

timholy:
[...]

PR #269, opened 2019-04-03
 Add `entr` to do work when files update 

timholy:
This is an oft-requested feature whose solution was pioneered in discussion in #203. I initially felt it was orthogonal to Revise, but one important thing changed: Revise's ability to follow `dev->add` and back means that a low-level implementation in terms of absolute paths would be painful from the user side. So, at the risk of growing this package, it now seems to belong here.

I also have the sense that usage of `includet` is getting out of hand. I suspect some of its users are really after `entr`.

CC @mlhetland, @essenciary, @yurivish

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,7 +16,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
-CodeTracking = "0.4, 0.5"
+CodeTracking = "~0.5.1"
 JuliaInterpreter = "0.2, 0.3, 0.4"
 LoweredCodeUtils = "0.3"
 
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 1.0
 OrderedCollections
-CodeTracking 0.4 0.6
+CodeTracking 0.5.1 0.6
 JuliaInterpreter 0.2 0.5
 LoweredCodeUtils 0.3 0.4
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -119,6 +119,22 @@ With Revise, you can
 
 all without restarting your Julia session.
 
+## Other Revise workflows
+
+Revise can be used to perform work when files update.
+For example, let's say you want to regenerate a set of web pages whenever your code changes.
+Suppose you've placed your Julia code in a package called `MyWebCode`,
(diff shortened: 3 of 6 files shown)


mlhetland:
Great stuff!

yurivish:
Tim, this looks fantastic. I'm not currently deep in the same mindspace I was when working on the package where I originally wanted this functionality, but I think that your example captures that use case perfectly.

Another one off the top of my head is an eventual future integration of Julia into [Observable Notebooks](https://observablehq.com), which are visualization-focused interactive notebooks. I recently asked Mike Bostock, creator of Observable and d3 about his plans to open it up to non-Javascript languages eventually, and it is possible but will require their runtime to support distributed computation (currently all notebook code is JS, and runs inside your browser). 

Observable notebooks are reactive, where the cells define a dependency graph. One of the challenges there is keeping the displayed values in the notebook internally consistent and, for a Julia version, consistent with any edits to outside files that affect the values of variables in the notebook. Some of this might be in the realm of Revise.

mlhetland:
🎉

essenciary:
I didn't get to say thank you for this, so thank you! I'll use it for live reload in genie.

essenciary:
@timholy 
Seems to be broken in v2.1.4 (I'll open a separate issue) so I can't comment much on functionality yet, but in terms of early feedback: 
1 - it shouldn't block (I'm thinking I can circumvent this by using `@spawn` or `@async` but I need to run some tests to be sure)
2 - ideally the callback would receive the path to the changed file as an argument, for more granular logic

#### Error:  
```julia
julia> Revise.entr(["/Users/adrian/Dropbox/Projects/testapp/app/resources/books"]) do 
       println("Updated")
       end
ERROR: type Pair has no field renamed
getproperty(::Any, ::Symbol) at ./sysimg.jl:18
macro expansion at /Users/adrian/.julia/packages/Revise/UtBAC/src/Revise.jl:692 [inlined]
(::getfield(Revise, Symbol("##66#67")){getfield(Main, Symbol("##5#6"))})() at ./task.jl:259
Stacktrace:
 [1] sync_end(::Array{Any,1}) at ./task.jl:226
 [2] macro expansion at ./task.jl:245 [inlined]
 [3] entr(::getfield(Main, Symbol("##5#6")), ::Array{String,1}, ::Nothing) at /Users/adrian/.julia/packages/Revise/UtBAC/src/Revise.jl:687
 [4] entr(::Function, ::Array{String,1}) at /Users/adrian/.julia/packages/Revise/UtBAC/src/Revise.jl:675
 [5] top-level scope at none:0
```

PR #287, opened 2019-04-13
update docs for Juno compatibility

pfitzseb:


timholy:
Thanks! I'm guessing [these lines](https://github.com/timholy/Revise.jl/blob/1a739c61dcf2892225af9538f62f7221c69f1775/src/Revise.jl#L1038-L1040) aren't really working? Is it just a loading-order thing or something that would qualify as "broken"?

pfitzseb:
Hm, actually the loading order should be something like
1. `startup.jl`
2. `using Atom`
3. `atreplinit`
4. `juno_startup.jl`

so the current way should work fine (and actually does for me in my very limited tests). 

Maybe @mkborregaard, @ssfrr, and @oxinabox can chime in here, because you reported stuff not working with Juno+Revise :)

ssfrr:
Yeah, previously Juno+Revise wasn't working for me because Revise was loaded in my `startup.jl` (before Juno) and so `Revise.setup_atom()` was never getting called. My solution was to make sure that Revise got loaded inside `juno_startup.jl` instead of `startup.jl` so it would successfully detect Atom was loaded. This PR's approach of running `setup_atom()` inside your `juno_startup.jl` seems like it should work fine as well.

Incidentally this was the issue that led me to write [Hooks.jl](https://github.com/ssfrr/hooks.jl) so that this setup code wouldn't have to live in both places to handle both package load orderings. LMK if you're interested in integrating it and I'll fix up the docs and actually register the package.

Basically Atom would add this code to its `__init__()`:

```julia
using Hooks
notify(hook"atom-init", @__MODULE__)
```

So it will pass a reference to the module to the hook.

Then Revise would have in its `__init__()`:

```julia
using Hooks
handle(hook"atom-init") do Atom
    setup_atom(Atom)
end
```

Or even more concisely:

```julia
using Hooks
handle(setup_atom, hook"atom-init")
```


Then `setup_atom` would get called whatever order the two packages get loaded in.

timholy:
Presuming you intend to keep Hooks lightweight, I'd merge a PR here that makes it a dependency and handles the integration with other packages.

ssfrr:
I've realized that the extra code in `juno_startup.jl` might not be necessary after all. I think (though I'm not 100% sure) when I was having trouble before I wasn't actually loading Revise within an `atreplinit` block. I agree with @pfitzseb's [comment above](https://github.com/timholy/Revise.jl/pull/287#issuecomment-482926298) that the currently-documented method should be OK.

timholy:
OK, I will close this and wait for the Hooks-based solution.

PR #288, opened 2019-04-17
fix appveyor julia version

johnnychen94:
`julia_version: 1` actually calls the latest stable version, i.e., `julia v1.1.0`


fredrikekre:
Cool, that is pretty annoying though.

PR #297, opened 2019-04-30
fix Revise breakpoint toggling with new breakpoint implementation in JuliaInterpreter

KristofferC:
Will be needed for https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/259.


timholy:
Since this apparently works now, no reason to wait to merge & tag, right?

KristofferC:
Yeah, if we have enough confidence that the API in that breakpoint PR is good enough. But I'm not really planning on changing it.

PR #306, opened 2019-05-12
Use correct error type for inotify warning

mortenpi:
Unless there is something I am missing, `watch_file` throws a `Base.IOError`, not `SystemError`:

```
julia> using FileWatching; try watch_file("test.jl") catch e e end
Base.IOError("FileMonitor (start): no space left on device (ENOSPC)", -28)
```

So currently the warning about used up inotify capacity actually never gets shown and Revise just silently refuses to update definitions.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -39,7 +39,7 @@ function wait_changed(file)
     try
         polling_files[] ? poll_file(file) : watch_file(file)
     catch err
-        if Sys.islinux() && err isa SystemError && err.errnum == 28  # ENOSPC
+        if Sys.islinux() && err isa Base.IOError && err.code == -28  # ENOSPC
             @warn """Your operating system has run out of inotify capacity.
             Check the current value with `cat /proc/sys/fs/inotify/max_user_watches`.
             Set it to a higher level with, e.g., `echo 65536 | sudo tee -a /proc/sys/fs/inotify/max_user_watches`.


mortenpi:
I haven't checked the `poll_file` branch though. But since it's polling, I guess the inotify error would never be thrown from there?

timholy:
Definitely not worried about `poll_file`, for the reasons you describe.

Did you have a reliable way to trigger this? I just tried creating more watches than I have capacity, but it didn't error.
```julia
julia> using FileWatching; try @sync for i = 1:10^4; @async watch_file("/tmp/test.jl"); end; catch e; e; end
```

mortenpi:
At least on my system, I can actually set `max_user_watches` to a really low value like 100 or something, so I can reproduce it on a system that isn't having inotify problems otherwise.

By the way, based on the stack trace, [this is where the `IOError` gets thrown](https://github.com/JuliaLang/julia/blob/80516ca20297a67b996caa08c38786332379b6a5/stdlib/FileWatching/src/FileWatching.jl#L411-L420), with [`uverror` etc defined in `libuv.jl`](https://github.com/JuliaLang/julia/blob/v1.1.0/base/libuv.jl#L68-L85) (and as far as I can tell the code hasn't changed much between 0.7 and current master).

timholy:
Merged in #307

Issue #298, opened 2019-05-01
entr should not break on rename


mlhetland:
I was puzzled for a while about why I couldn't get `Revise.entr` to work – every time I modified a file, it just quit. Then I realized it was because my editor (`vim`) uses two-stage saving, where it writes to a temp file and then moves that to overwrite the original. And I don't think that is a situation where `entr` should conclude that the file is no longer relevant; indeed, this works just fine with the originale `entr` [command-line tool](http://eradman.com/entrproject/). 

I'm not sure exactly what the behavior *should* be – e.g., if `renamed` is also true if the original file is renamed to something else (in which case we might *not* want to keep watching it), but at least for my use-case (editing with `vim`), the current behavior is not helpful.

timholy:
Nice debugging. Perhaps try putting in a short sleep and see if that fixes it?

Code for `entr` is quite small so I suspect you'll have little trouble playing with it, but don't hesitate to ask if you have questions.

mlhetland:
Indeed – it's still quite close to [my version](https://github.com/timholy/Revise.jl/issues/203#issuecomment-429913111), so that should be fine. I started looking at it, but was (1) a bit uncertain about what the behavior should be, and (2) a bit uncertain about how to add the appropriate tests :-}

I'm not sure how a sleep would work, though. The returned event will still indicate that the file has been renamed, no? That's information we get from the `Node.js` library regardless?

As far as I can see, the event returned is identical, regardless of whether a new file overwrites the one we're tracking or if the tracked file is renamed to something else (i.e., it's a `FileEvent(true, false, false)` either way).

The behavior of [entr](http://eradman.com/entrproject/) is:
1. Trigger once initially, unconditionally (as opposed to `Revise.entr` – maybe something to consider);
2. If the file is overwritten, trigger as usual;
3. If the file is moved elsewhere, quit with an error.

I'm not sure what the right thing is in case 3. One possibility is to not do anything particular – we'd end up with an error once we try to keep tracking changes after it's gone, anyway, and the resulting `IOError` is clear enough.

So: My suggestion would actually be simply to [the conditional break](https://github.com/timholy/Revise.jl/blob/31344123a6911e5af470bc099ef424bb925231f5/src/Revise.jl#L692) when `ret.renamed` is true.

[...]

PR #308, opened 2019-05-13
Rewrite, postpone and pause

mlhetland:
This PR has three minor modifications to `entr` (cf. #298):

- By default, `entr` will now execute `f()` initially, unless the `postpone` keyword is used;
- `entr` sleeps for the amount given by `pause` between being triggered and executing `f()`, to handle clusters of modifications, such as text editor saves; and
- Rewrites *to* a given file/file name are now accepted as modifications; rewrites *from* the file name cause an error, as the file with the supplied name no longer exists.

The first and last are in line with the `entr` command line tool. The second is an added convenience, to avoid redundant execution when (e.g.) generating output based on saved files, etc.

The modifications are independent – no need to accept them all.

(Note: No new tests have been added for these features.)

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -692,8 +692,7 @@ function entr(f::Function, files, modules=nothing; postpone=false)
                 waitfor = isdir(file) ? watch_folder : watch_file
                 @async while active
                     ret = waitfor(file, 1)
-                    ret.renamed && break
-                    if active && ret.changed
+                    if active && (ret.changed || ret.renamed)
                         revise()
                         f()
                     end
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -662,7 +662,7 @@ Execute `f()` whenever files listed in `files`, or code in `modules`, updates.
 `entr` will process updates (and block your command line) until you press Ctrl-C.
 Unless `postpone` is `true`, `f()` will be executed also when calling `entr`,
 regardless of file changes. The `pause` is the period (in seconds) that `entr`
-will wait between it has been triggered and actually calling `f()`, to handle
+will wait between being triggered and actually calling `f()`, to handle
 clusters of modifications, such as those produced by saving files in certain
 text editors.
 
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -655,11 +655,13 @@ end
 includet(file::AbstractString) = includet(Main, file)
 
 """
-    entr(f, files)
-    entr(f, files, modules)
+    entr(f, files; postpone=false)
+    entr(f, files, modules; postpone=false)
 
 Execute `f()` whenever files listed in `files`, or code in `modules`, updates.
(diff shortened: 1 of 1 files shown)


src/Revise.jl
```diff
@@ -655,11 +655,16 @@ end
...
-    entr(f, files)
-    entr(f, files, modules)
+    entr(f, files; postpone=false, pause=0.02)
+    entr(f, files, modules; postpone=false, pause=0.02)
 
 Execute `f()` whenever files listed in `files`, or code in `modules`, updates.
 `entr` will process updates (and block your command line) until you press Ctrl-C.
+Unless `postpone` is `true`, `f()` will be executed also when calling `entr`,
+regardless of file changes. The `pause` is the period (in seconds) that `entr`
+will wait between it has been triggered and actually calling `f()`, to handle
```
timholy:
```suggestion
will wait between being triggered and actually calling `f()`, to handle
```

mlhetland:
> Feel free to add any tests that you want. If some behaviors are difficult to test, I'll leave it up to your judgement to decide where to draw the line; let me know when you want this merged and I'll do it.

There should probably be a couple of added tests in the `"entr"` test set. I've got it on my todo list, but if anyone is waiting for this fix/functionality, feel free to add/suggest some – not sure when I'll get to it.

timholy:
If you think it will be a long time, like I said I'm willing to merge this now and risk breakage later. I kind of think of `entr` as "a community supported extra" to Revise, so I'm fine if the community is willing to risk breakage by going without thorough tests.

Just let me know what you want me to do.

mlhetland:
I'm fine with that. I guess if there's a bug that needs fixing, we could add a test provoking that bug, anyway :-)

timholy:
Technically this is breaking (previously we effectively had `postpone=true`), but I think of the old behavior as a bug, so I'll release this as Revise 2.2 rather than Revise 3.0.

mlhetland:
Yes, I thought about that, too (specifically the `postpone` thing) – but I guess breaking changes don't depend on actual behavior, but *documented* behavior (if you're going with semantic versioning), so I guess the question here is how clear an expectation one might have, from the documentation, that there would be no call until the file changes. It states that it'll be called whenever the file is updated, of course, so I guess it would be reasonable to infer that it wouldn't be called until the file is updated post calling `entr` – but given the rather high-level description, it's not something I'd have relied on. (Then again, to find the specific behavior, I probably *would* have consulted the source, so … :-})

But despite being (*mildly*) breaking, the fix (`postpone=true`) is very easy to add. And, as you say, given the naming (from the command-line tool), it's not unreasonable to see this as a bug.

That reminds me: I didn't modify any documentation (beyond the docstring itself). If it's mentioned elsewhere, that should probably also be changed?

timholy:
Improvement to the documentation would be most welcome. Note #311, which I think is due to JuliaInterpreter not working on nightly. (I guess fixing that should be my next priority.)

PR #313, opened 2019-05-28
update to 2.1.6

pfitzseb:
for compatibility with JuliaInterpreter 0.6.


KristofferC:
Restarting since LoweredCodeUtils was just tagged so we can run tests with JuliaInterpreter 0.6.

KristofferC:
Passes with JuliaInterpreter 0.6 on everything except nightly (where it is the known GC corruption bug).

Issue #311, opened 2019-05-23
Documentation is not building. 


timholy:
https://travis-ci.org/timholy/Revise.jl/builds/536209996

Is this because of failures on nightly?

fredrikekre:
Yea, the next build stage does not start unless the previous one was successful.

PR #314, opened 2019-05-28
allow failures on nightly

pfitzseb:


timholy:
There's part of me that prefers to see the failure so I don't forget about it. Is the failure causing a problem? (E.g., are you thinking of #311?)

KristofferC:
The GC bug got fixed so maybe things will work on nightly again?

pfitzseb:
> There's part of me that prefers to see the failure so I don't forget about it. Is the failure causing a problem? (E.g., are you thinking of #311?)

Fair enough. I was mainly bothered by the red badges in the readme :)

PR #329, opened 2019-08-15
Mention autoloading at beginning of docs

pearlzli:
Suggested on [Discourse](https://discourse.julialang.org/t/more-basic-revise-usage-with-module/27517/8).


tpapp:
I think this is important enough to deserve a [`!!! note`](https://docs.julialang.org/en/v1/stdlib/Markdown/#Admonitions-1) box.

pearlzli:
@tpapp Sure, I've added that.

timholy:
Thanks @pearlzli!

PR #331, opened 2019-08-19
add compatibility with JuliaInterpreter 0.7

KristofferC:
Tested locally with JuliaInterpreter 0.7 and passes. For this to run on CI, the PR to General for 0.7 needs to be merged and LoweredCodeUtils also needs new compat + tag.


timholy:
Thanks. Just FYI I have a local version updated to nightly that mostly works (requires https://github.com/JuliaLang/julia/pull/32961, but latest master segfaults near the end...).

PR #332, opened 2019-08-19
Update tests for Julia 1.3

timholy:
Needs https://github.com/JuliaLang/julia/pull/32961, but let's see if there are any other failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ os:
 julia:
   - 1.0
   - 1.1
+  - 1.2
   - nightly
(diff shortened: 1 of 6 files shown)


KristofferC:
You could sneakily just define the method here if it doesn't exist... If you want to be less sneaky, only define it in the tests.

timholy:
This ended up being a bigger cleanup---I rethought the `mtime` issues that have been causing sporadic test failures on OSX. I think the new approach is much cleaner, and will hopefully end those problems.

PR #336, opened 2019-08-23
See if increasing delay helps entr tests pass on OSX

timholy:


timholy:
The only failures here were travis timeouts, so I think this solves the specific problem. Timeouts will be a separate PR.

PR #338, opened 2019-08-24
A usage cookbook

timholy:
In response to https://discourse.julialang.org/t/trying-and-trying-with-revise/27915

### DIFF
--- a/docs/src/cookbook.md
+++ b/docs/src/cookbook.md
@@ -1,0 +1,148 @@
+# Revise usage: a cookbook
+
+## Package-centric usage
+
+For code that might be useful more than once, it's often a good idea to put it in
+a package.
+For creating packages, the author recommends [PkgTemplates.jl](https://github.com/invenia/PkgTemplates.jl).
+
+!!! note
+    If you've never developed code before, this approach might require you to do some configuration.
+    (Once you get things set up, you shouldn't have to do this part ever again.)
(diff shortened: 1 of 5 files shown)


docs/src/cookbook.md
```diff
@@ -0,0 +1,148 @@
...
+
+```julia
+julia> MyPkg.greet()
+Hello, revised World!
+```
+
+From this point forward, revisions should be fast. You can modify `MyPkg.jl`
+quite extensively without quitting the Julia session, although there are some [Limitations](@ref).
+
+## includet usage
```
fredrikekre:
```suggestion
## `includet` usage
```

docs/src/cookbook.md
```diff
@@ -0,0 +1,148 @@
...
+    (Some instructions on configuration are [here](https://help.github.com/en/articles/set-up-git)
+    and [here](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup).)
+    It's also helpful to sign up for a [GitHub account](https://github.com/)
+    and set git's `github.user` variable.
+    The [PkgTemplates documentation](https://invenia.github.io/PkgTemplates.jl/stable/)
+    may also be useful.
+
+!!! note
+    If the current directory in your Julia session is itself a package folder, PkgTemplates
+    with use it as the parent environment (project) for your new package.
```
visr: with use it -> will use it

PR #345, opened 2019-08-31
Preserve revision_queue items that errored

### LINKED ISSUE #317: Optional continuous warnings after revise failure resp. on inconsistent state
A frequent problem I've encountered with less experienced users is that they often overlook a revise error message (e.g. when revise failed because of errors in the changed source code). Since revise reports this only once, the users are often unaware that they're still using the old version of the code afterwards.

It would be nice if Revise could (as an option) warn that on-disk source code and active code are inconsistent on every REPL/Jupyter-cell evaluation, to keep users aware that they're not running the code they think they are.

### LINKED ISSUE #321: Remind user that file is skipped due to parsing error
I think it could be helpful for `Revise` to remind the user that certain files are being skipped due to parsing errors.

For instance, every so often I might skip a parsing error message when multiple errors exist, and on subsequent method runs miss the fact that a file is being omitted... which leads me off into a fruitless debug loop. I try to remember that this can happen with Revise, so restart the session, but it's rarely the first thing I think of.

timholy:
This issues a continuous reminder about files that are out-of-sync. Fixes #317, fixes #321.

I may not merge this immediately, this seems like the kind of change that would be good to try out before making it the standard. @oschulz and @ianshmean, can you try out this branch and see how you like it?

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -95,6 +95,13 @@ This list gets populated by callbacks that watch directories for updates.
 const revision_queue = Set{Tuple{PkgData,String}}()
 
 """
+    Revise.queue_errors
+
+Global variable, lists `(pkgdata, filename)` pairs that errored upon last revision
+"""
+const queue_errors = typeof(revision_queue)()
+
+"""
     Revise.pkgdatas
 
 `pkgdatas` is the core information that tracks the relationship between source code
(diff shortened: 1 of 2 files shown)


ianshmean:
This is great! 
This tweak might help clarify for the warning message:
`Running code does not match saved version for the following files:\n$str`
to
`Due to a previously reported error, the running code does not match saved version for the following files:\n$str`

oschulz:
Just tried it - very nice, thanks a lot! I also like your suggested message format, @ianshmean .

timholy:
I've just rebased this on current master (which is being released as Revise 2.1.10). If problems don't crop up, I'll merge this for the next release.

PR #349, opened 2019-09-01
Allow Revise to be loaded in -e/E arguments

### LINKED ISSUE #202: Revise loaded in -E/-e for -i not updating definitions
When loading a Julia script as follows:

```sh
julia -i -e 'using Revise; includet("script.jl")'
```

Revise will not update the function definitions etc. afterwards. The original version of the script gets loaded without issues though.

Tested with Revise master (dd6e970a23e0b998883f814742723bf53b88f57f) on Julia 1.0.0 and the normal Revise workflow (`using Revise; includet("script.jl")` in the REPL) works without problems.

mortenpi:
@timholy Did you mean like this?

I have been running Revise with essentially this patch (sans the `Base.isinteractive` check) for quite a while now and everything has been working fine.

At some point, I did see an issue when I would get some warnings with scripts (i.e. running `julia script.jl`). I don't remember exactly, but I think it was about stealing the REPL when there isn't one when I had `using Revise` in the startup file. But I wasn't able to reproduce that anymore. And hopefully the `Base.isinteractive` check will get around that?

As far as I can tell, this definitely fixes #202, i.e. `julia -i -e 'using Revise; includet("script.jl")'` now works as expected.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -1045,7 +1045,9 @@ function __init__()
         elseif isdefined(Main, :IJulia)
             Main.IJulia.push_preexecute_hook(revise)
         elseif Base.isinteractive()
-            @async Revise.wait_steal_repl_backend()
+            atreplinit() do repl
+                @async Revise.wait_steal_repl_backend()
+            end
         end
         if isdefined(Main, :Atom)
             setup_atom(getfield(Main, :Atom)::Module)
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -1044,6 +1044,8 @@ function __init__()
             steal_repl_backend(Base.active_repl_backend::REPL.REPLBackend)
         elseif isdefined(Main, :IJulia)
             Main.IJulia.push_preexecute_hook(revise)
+        elseif Base.isinteractive()
+            @async Revise.wait_steal_repl_backend()
         end
         if isdefined(Main, :Atom)
             setup_atom(getfield(Main, :Atom)::Module)


mortenpi:
Also, with this patch, I think the recommended way to load Revise in the `startup.jl` file would become just `using Revise`. If I have this

```julia
atreplinit() do repl
    try
        @eval using Revise
    catch
    end
end
```

then the following happens

```
$ julia -i -e'includet("foo.jl")'
ERROR: UndefVarError: includet not defined
Stacktrace:
 [1] top-level scope at none:1
```

But if you just have `using Revise`, then `julia -i -e'includet("foo.jl")'` works fine. And the definitions in `foo.jl` will be kept up to date in the REPL session.

mortenpi:
Ah, the warning showed up. It was the same one that @mlhetland mentioned in https://github.com/timholy/Revise.jl/issues/202#issuecomment-518343458. I had `using Revise` in `startup.jl` and there was some compilation necessary, so this happened:

```
$ julia -i --project analysis/v4.jl
[ Info: Recompiling stale cache file /home/mortenpi/.julia/compiled/v1.2/DataFrames/AR9oZ.ji for DataFrames [a93c6f00-e57d-5684-b7b6-d8193f3e46c0]
┌ Warning: REPL initialization failed, Revise is not in automatic mode. Call `revise()` manually.
└ @ Revise ~/.julia/dev/Revise/src/Revise.jl:986
```

timholy:
Can we wrap this in an `atreplinit`?That might solve the "how long a delay?" issue.

mortenpi:
Ok, I've tested this wrapped version and it seems to work great. The warning is gone and loading Revise in either `startup.jl` or `-e/E` still works.

Are there any tests that should/could be added for this?

timholy:
> Are there any tests that should/could be added for this?

Yes please. I suspect the best approach would be in the `.travis.yml` file. You'll note there are already some tests with different startup options, this would be kind of the same thing.

I haven't yet played with this myself, but any preliminary thoughts about whether this affects people who have been adding the recommended lines to their `startup.jl` files?

timholy:
This works in current Revise, let's close this and the associated issue.

PR #353, opened 2019-09-03
 Improve error message in case of parsing problems (fixes #301) 

### LINKED ISSUE #301: Handle error in lowering
Having a referenced undefined label causes an error during lowering that makes Revise throw an error:

```
lowering returned an error, $(Expr(:error, "label \"err\" referenced but not defined"))
Stacktrace:
 [1] error(::String, ::Expr) at .\error.jl:42
 [2] prepare_thunk(::Module, ::Expr, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:342
 [3] prepare_thunk(::Module, ::Expr, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:353
 [4] prepare_thunk at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:339 [inlined]
 [5] #methods_by_execution!#9(::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:define,),Tuple{Bool}}}, ::Function, ::Any, ::Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::Module, ::Expr) at C:\Users\Kristoffer\Debugging\Debugger\dev\Revise\src\lowered.jl:47
 [6] #methods_by_execution! at .\none:0 [inlined]
 [7] #eval_with_signatures#59(::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Module, ::Expr) at C:\Users\Kristoffer\Debugging\Debugger\dev\Revise\src\Revise.jl:344
 [8] eval_with_signatures at C:\Users\Kristoffer\Debugging\Debugger\dev\Revise\src\Revise.jl:342 [inlined]
 [9] (::getfield(Revise, Symbol("##57#58")){OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}},OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}},Module})() at C:\Users\Kristoffer\Debugging\Debugger\dev\Revise\src\Revise.jl:278
 [10] with_logstate(::getfield(Revise, Symbol("##57#58")){OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}},OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}},Module}, ::Base.CoreLogging.LogState) at .\logging.jl:395
```

Perhaps this should be "handled" in similar way as syntax errors?

timholy:
This should generate error messages much more similar to Base's. The biggest change here is to avoid calling `unwrap` on the stored expressions, which produced better location info in some cases but necessitated some changes elsewhere.

Also, @mlhetland and @essenciary, I have disabled the `entr` tests on OSX. They have been the cause of [many failures](https://travis-ci.org/timholy/Revise.jl/builds) (note I went through a period where I allowed failures on OSX, so some of those green checks are a bit misleading). I have tried several approaches (e.g., 77cfdcb3f1c71eeec2bacebcbfdb43a3cb5dba38, e22ec82548c103f97cccb6a5d7b024e65fda759c, f1645112a28833dd9ad20376a794dc367ae33160, 8c60017f70824a264e37367fecf1cf89eb80c578) without completely nailing the problem. Because `entr` is a bit specialized, I don't want to fail to test the core functionality of the package, so disabling those specific tests seems best. But of course things could decay, so I wanted you to know.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -291,7 +291,7 @@ function eval_new!(exs_sigs_new::ExprsSigs, exs_sigs_old, mod::Module)
             else
                 sigs = exs_sigs_old[rexo]
                 # Update location info
-                ln, lno = firstline(rex), firstline(rexo)
+                ln, lno = firstline(unwrap(rex)), firstline(unwrap(rexo))
                 if sigs !== nothing && !isempty(sigs) && ln != lno
                     @debug "LineOffset" _group="Action" time=time() deltainfo=(sigs, lno=>ln)
                     for sig in sigs
@@ -563,12 +563,13 @@ function revise(mod::Module)
         for (mod, exsigs) in fi.modexsigs
             for def in keys(exsigs)
                 ex = def.ex
-                isexpr(ex, :call) && ex.args[1] == :include && continue
+                exuw = unwrap(ex)
+                isexpr(exuw, :call) && exuw.args[1] == :include && continue
                 try
                     Core.eval(mod, ex)
                 catch err
                     @show mod
-                    display(def)
+                    display(ex)
                     rethrow(err)
                 end
             end
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -205,7 +205,7 @@ function methods_by_execution!(@nospecialize(recurse), methodinfo, docexprs, fra
                             showerror(io, err)
                         end
                         fl, ln = whereis(frame)
-                        println(stderr, "\nin expression starting at ", fl, ": ", ln)
+                        println(stderr, "\nin expression starting at ", location_string(fl, ln))
                         badstmt = lookup_callexpr(frame, stmt)
                         @warn "omitting call expression $badstmt"
                         assign_this!(frame, nothing)
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -190,7 +190,7 @@ function methods_by_execution!(@nospecialize(recurse), methodinfo, docexprs, fra
                     # A :call Expr we don't want to intercept
                     try
                         pc = step_expr!(recurse, frame, stmt, true)
-                    catch
+                    catch err
                         # This can happen with functions defined in `let` blocks, e.g.,
                         #     let trynames(names) = begin
                         #         return root_path::AbstractString -> begin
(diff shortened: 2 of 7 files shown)


mlhetland:
Per your request in [#354](https://github.com/timholy/Revise.jl/issues/354#issuecomment-529093945), I have now run the (original) `entr` tests 100 times on macOS, and have not triggered any errors…

timholy:
That's very kind of you. Weird that it barfs on Travis so often. Well, let's just leave those tests disabled then!

mlhetland:
Does it work (consistently) in *any* of the Mac environments? Maybe we could do a more fine-grained disabling? Then again, the new test I added in #355 is not (yet) disabled – if it works, it does exercise the basic `entr` functionality, which is at least something.

mlhetland:
Is [this](https://travis-ci.org/timholy/Revise.jl/jobs/580169351) an example of the kind of failure you're talking about? It seems that's what (sometimes) happened for my test, too – i.e., that `entr` wasn't triggered, which meant that the test would hang. In that case, at least, it was a matter of waiting for even longer (and I added a `touch`, just in case…), to ensure `entr` was up and running by the time the file was modified. I guess the test-waits in macOS are already a couple of seconds, but … maybe this could be extended further? (Or maybe modify the file repeatedly in a loop, until `entr` reacts?)

mlhetland:
Might not be worth it, though. It would be just to get the test to run – not really to do any valuable testing. 🤷‍♂

timholy:
We could experiment with waiting longer. The thing I found surprising is that it's not random, it's *always* the `entr` tests that fail. (There was another one earlier but I fixed that one, it was missing a file-aging sleep.) The fact that it's so consistent makes me suspect that waiting another second won't really help. But I'm game to try if you want to.

Issue #354, opened 2019-09-05
entr triggered, but no revision


mlhetland:
I apologize in advance for this (at least for now) rather vague issue description … but it's triggered in a rather large, currently unpublished code base, and I'm not entirely sure how to reproduce more generally. I'll make some attempts, but I thought maybe I could get some pointers on what to look for, unless the cause happens to be obvious.

I'm using `entr` in a script, more or less like so:

```
using Revise

push!(LOAD_PATH, "demo") # demo directory contains Demos.jl
using Demos, MyModule

entr(["demo/rundemos.jl", "src/path/somefile.tex"], [Demos, MyModule]) do
    print("recompiling and rerunning ...")
    Demos.run_some_demo("some argument")
    println(" done")
end
```

I'm in the project directory, with `MyModule.jl` in `src` (along with lots of included stuff).

Now, if I edit any of the files that make up `Demos` or `MyModule`, `entr` is triggered, and the code is run. However, the modifications are only actually incorporated if they are in the files belonging to `MyModule`; if I edit any of the files of `Demos`, `entr` is triggered, *but it runs the old code*.

Now … I do have a possible cause – that `Demos` runs `include` statements in a loop over `readdir(@__DIR__)`, which I guess is probably a rather bad idea, anyway. Actually, I'm almost surprised that the functions defined in the `include`d files aren't lost, as they're defined in the loop body … but since they do work, I guess maybe `Revise` should handle them? Or maybe I should just find a better way of doing the inclusion :-> (I do wish I could have it be automated, though…)

Either way, if this is either a clear “to-don't” or there is some workaround (or, indeed, there is an issue that can be fixed in `Revise`) it might be worth documenting or something… (If nothing else, I guess it'll be documented here ;-D)

mlhetland:
OK, can confirm that with hard-coded includes (i.e., not in a loop) things do work as they should.

mlhetland:
No, scratch that … that actually *didn't* help. So it seems it's some other part of the setup that's causing it? Well, I'll see if I can narrow it down, but if anyone has suggestions, that'd be helpful.

timholy:
Just to check, `Demos` is precompiled? If so it shouldn't matter whether they are hard-coded or in a loop, because the list of files is taken from the precompile header (see `src/pkgs.jl`) which is a comprehensive list of all dependent files.

[...]

Issue #344, opened 2019-08-31
Segfault on sigtest


timholy:
The sequence
```julia
using LibGit2
include("sigtest.jl")
```
results in a segfault. More directly:
```julia
julia> using LibGit2, Revise

julia> file = "/home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/LibGit2/src/consts.jl"
"/home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/LibGit2/src/consts.jl"

julia> mexs = Revise.parse_source(file, LibGit2)
OrderedCollections.OrderedDict{Module,OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}}} with 2 entries:
  LibGit2        => ExprsSigs(<0 expressions>, <0 signatures>)
  LibGit2.Consts => ExprsSigs(<178 expressions>, <0 signatures>)

julia> Revise.instantiate_sigs!(mexs)
```
triggers it.

Uncommenting some debug lines in JuliaInterpreter yields the crucial bit of information
```julia
LibGit2.Consts
  doc!(__module__::Module, b::Base.Docs.Binding, str::Base.Docs.DocStr, sig) in Base.Docs at docs/Docs.jl:217
    handle_message(logger::Logging.ConsoleLogger, level, message, _module, group, id, filepath, line) in Logging at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Logging/src/ConsoleLogger.jl:100
      #handle_message#2(maxlog, kwargs, ::typeof(Base.CoreLogging.handle_message), logger::Logging.ConsoleLogger, level, message, _module, group, id, filepath, line) in Logging at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Logging/src/ConsoleLogger.jl:100
        displaysize(io::Base.TTY) in Base at stream.jl:458
          iolock_end() in Base at libuv.jl:49, pc = 1
node = :(($(QuoteNode(JuliaInterpreter.CompiledCalls.##compiledcall#425)))())
```
meaning that a runtime call to `displaysize` from the Doc module triggered this problem, which failed on on `iolock_end()`.

Adding `displaysize(::Base.TTY)` to JuliaInterpreter's `compiled_methods` fixes the problem.

timholy:
Fixed by https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/335

PR #355, opened 2019-09-07
Test and fix for issue #354

mlhetland:
Incorporates @timholy's fix and adds my test.

Note: The test does not properly exercize the added `yield()` statement.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2189,6 +2189,10 @@ const A354_result = Ref(0)
     @async begin
         sleep(mtimedelay)
         setvalue(2)
+        # belt and suspenders -- make sure we trigger entr:
+        sleep(mtimedelay)
+        touch(srcfile)
+        sleep(mtimedelay)
     end
 
     try
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2162,6 +2162,40 @@ end
     end
 end
 
+# issue #354
+@testset "entr with modules" begin
+
+    testdir = newtestdir()
+    modname = "A354"
+    srcfile = joinpath(testdir, modname * ".jl")
+
+    function setvalue(x)
+        open(srcfile, "w") do io
+            print(io, "module $modname test() = $x end")
+        end
+    end
+
+    setvalue(1)
+    @eval using A354
+
+    result = 0
+    t = @async begin
+        entr([], [A354], postpone=true) do
+            result = A354.test()
+        end
+    end
+    sleep(mtimedelay)
+
+    setvalue(2)
+    sleep(mtimedelay)
+
+    @test result == 2
+
+    t.state = :done
+    rm_precompile(modname)
+
+end
+
 println("beginning cleanup")
 GC.gc(); GC.gc()
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2162,6 +2162,8 @@ end
     end
 end
 
+const A354_result = Ref(0)
+
 # issue #354
 @testset "entr with modules" begin
 
@@ -2182,11 +2184,11 @@ end
     @eval using A354
     sleep(mtimedelay)
 
-    result = Ref(0)
+    A354_result[] = 0
 
     @async begin
         entr([], [A354], postpone=true) do
-            result[] = A354.test()
+            A354_result[] = A354.test()
             throw(InterruptException())
         end
     end
@@ -2195,7 +2197,7 @@ end
     setvalue(2)
     sleep(mtimedelay)
 
-    @test result[] == 2
+    @test A354_result[] == 2
 
     rm_precompile(modname)
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2176,14 +2176,18 @@ end
     end
 
     setvalue(1)
+
+    # these sleeps may not be needed...
     sleep(mtimedelay)
     @eval using A354
     sleep(mtimedelay)
 
     result = 0
-    t = @async begin
+
+    @async begin
         entr([], [A354], postpone=true) do
             result = A354.test()
+            throw(InterruptException())
         end
     end
     sleep(mtimedelay)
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -2162,6 +2162,40 @@ end
...
+    modname = "A354"
+    srcfile = joinpath(testdir, modname * ".jl")
+
+    function setvalue(x)
+        open(srcfile, "w") do io
+            print(io, "module $modname test() = $x end")
+        end
+    end
+
+    setvalue(1)
```
timholy: I think you want `sleep(mtimedelay)` before & after the `@eval using` for reasons we discussed in #354. I know these tests don't run on OSX by default but we might as well get them as close to passing as possible.

mlhetland:
Hm. It seems the checks are failing…? Both warnings about running out of inotify capacity and an undefined `rseed`…?

test/runtests.jl
mlhetland: Right. This one actually *does* run on macOS at the moment, as it seemed to work just fine on my machine – but I can add the same conditional around it as the other `entr` stuff, if you'd like. As for the sleep – I didn't think it applied at that point in the code? (I'm just importing the module for the first time?) But no problem, I'll add that.

mlhetland:
(Oops, it seems my use of `@eval` in the commit msg. tagged a user. Sorry!)

test/runtests.jl
timholy: Maybe you were right, didn't notice the sleep after the `@async` loop. Still can't hurt.

mlhetland:
Well, I can't really make any sense of the check failures, here.

timholy:
Does it reproduce locally if you do `]test Revise`?

timholy:
The `inotify` text is expected, part of tests that only run on Linux & Travis. See old passing builds in the Travis history.

mlhetland:
It works locally – tested that before creating the PR. Also, I don't quite see the link between the errors and the new code/test?

timholy:
It fails for me locally. Are you sure you tested it with `]test Revise`? (The `]` meaning pkg-mode.)

mlhetland:
Yes, I understood. I was sure it worked at some point, but I can see it doesn't, now. (It did work when run from the command-line without `]`.)

mlhetland:
The new test succeeds, though, and the failure seems unrelated to `entr`.

mlhetland:
Oh, well, I'll do some local debugging :-)

mlhetland:
Well, at least it seems to be specifically the use of `entr` that fails – not the call to `A354.test()` nor any other part of the test (such as the `@async` stuff). And it fails with both the new and the old version of `entr`. However, if I run `julia test/runtests.jl`, it works just fine. And all I get from `Pkg` is:
```
ERROR: Package Revise errored during testing
```
Is there no way of getting the full stack trace?

timholy:
Definitely a weird error. When I run at the REPL, I see it starting the cleanup phase and then Julia just quits outright, dumping me back to the Linux prompt. Might be a Julia bug. I suspect it's a task problem.

timholy:
Read about the complexities at https://github.com/JuliaLang/julia/issues/6283 and then change the test so it doesn't try to stop the task.

mlhetland:
Ah, right. I actually came across that in my quest for how to stop the task without using exceptions. And I didn't look at it thoroughly enough :-}

The reason I didn't want to use exceptions was that I seemed to end up with a compound exception that I had to unpack (similarly to in the other test), though I thought that throwing an `InterruptException` should work in just terminating `entr`.

And now, somehow, it does… So unless there's some reason not to, I'll just go with that.

mlhetland:
Whoah! Seemed to more or less for the early checks, but then triggered a segfault in Julia nightly! Any ideas?

timholy:
Yay! The segfault may be #344

But I get the same error that shows up in one of the appveyor tests, https://ci.appveyor.com/project/timholy/revise-jl/builds/27252479/job/d20rpoxf78awybn4. Looking at the code, I am a bit surprised it works; I would have thought your implementation would suffer from scope problems. Maybe make `result` a `Ref`? I.e., `result = Ref(0)` and then access it with `result[]`.

mlhetland:
Hm. Could there still be scope issues, even when using a `Ref`? I.e., could there be some copying going on, or something? Maybe I'll try to have a global result, instead…

mlhetland:
OK, trying to move the `entr` and result assignment out of the `@async` (beyond what's in `entr` to begin with; I guess nesting that was an issue to begin with), moving the file modification to a separate task instead. Hopefully a bit cleaner.

And now I remember why the `InterruptException` didn't work with `entr` before … it's because I didn't have it in a separate task. So now, once again, that doesn't work. Given that we should only get the correct result assignment if nothing goes wrong, I'm just catching my own `error` with an indiscriminate `catch` here. Could do more checking and `rethrow`ing, if you think that's necessary.

mlhetland:
Hm. And now it stalls (sometimes) – I guess the `entr`wasn't triggered, then. Maybe I'll try another `sleep`…? :-P

mlhetland:
Hm. The stall was on macOS, where you've set `mtimedelay` to 2.1, so the wait before `setvalue` should be plenty of time to get `entr` running? I dunno. Adding a bit more poking at the file, just to see if it helps.

mlhetland:
Welp, at least now the only issues are the Julia nightly segfaults, as far as I can see. So I guess things are perhaps good to go? (Unless you want to turn off this new test on macOS, like the other `entr` test; it works, for now, but…)

timholy:
This indeed looks fine. Thanks for your patience! While the final result is fairly simple, it's pretty amazing how many factors come into place when you combine watching file systems & recompilation.

mlhetland:
That's fine! Yes, it's indeed a bit tricky – and here we had issues with `@async` mixed up with it all, I guess :-) And the resulting test is sort of a test we might have wanted for `entr` to begin with, I guess.

PR #357, opened 2019-09-09
check staleness before trying to verify the header

### LINKED ISSUE #356: Invalid header of precompile cache file on master
Using Revise I rarely get

```
julia> using Revise, Pkg
[ Info: Precompiling Revise [295af30f-e4ad-537b-8983-00126c2a3abe]
ERROR: InitError: ArgumentError: Invalid header in cache file /Users/kristoffer/.julia/compiled/v1.4/JuliaInterpreter/PliIn_OohgU.ji.
Stacktrace:
 [1] parse_cache_header(::String) at /Users/kristoffer/.julia/packages/Revise/JCJHO/src/pkgs.jl:50
 [2] pkg_fileinfo at /Users/kristoffer/.julia/packages/Revise/JCJHO/src/pkgs.jl:62 [inlined]
 [3] parse_pkg_files(::Base.PkgId) at /Users/kristoffer/.julia/packages/Revise/JCJHO/src/pkgs.jl:115
 [4] __init__() at /Users/kristoffer/.julia/packages/Revise/JCJHO/src/Revise.jl:1068
 [5] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:692
 [6] _require_from_serialized(::String) at ./loading.jl:743
 [7] _require(::Base.PkgId) at ./loading.jl:1034
 [8] require(::Base.PkgId) at ./loading.jl:922
 [9] require(::Module, ::Symbol) at ./loading.jl:917
during initialization of module Revise
```

My guess is that this is somehow interacting with https://github.com/JuliaLang/julia/pull/32651 but I cannot see how.

KristofferC:
Should fix #356

### DIFF
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -58,7 +58,9 @@ function pkg_fileinfo(id::PkgId)
     uuid, name = id.uuid, id.name
     # Try to find the matching cache file
     paths = Base.find_all_in_cache_path(id)
+    sourcepath = Base.locate_package(id)
     for path in paths
+        Base.stale_cachefile(sourcepath, path) === true && continue
         provides, includes_requires = parse_cache_header(path)
         mods_files_mtimes, _ = includes_requires
         for (pkgid, buildid) in provides


src/pkgs.jl
```diff
@@ -58,7 +58,9 @@ function pkg_fileinfo(id::PkgId)
     uuid, name = id.uuid, id.name
     # Try to find the matching cache file
     paths = Base.find_all_in_cache_path(id)
+    sourcepath = Base.locate_package(id)
     for path in paths
+        Base.stale_cachefile(sourcepath, path) === true && continue
```
KristofferC: `stale_cachefile` also, of course, parses the cache header internally but perhaps cheap enough to not worry?

timholy: On my machine it's of order 1ms. I'm not worried about it, thanks for doing this.

KristofferC:
It does fix it locally. AFAIU, we only want the data from `parse_cache_header` for the precompilation file that actually matches what we load so this seems like possibly the correct fix.

vchuravy:
Are we doing backports to Revise 1? If so it would be lovely to have this backported.

KristofferC:
Someone who wants that could set up a branch with this backported and the release version bumped and then it should be simple enough to just tag a new Revise 1 from that.

PR #367, opened 2019-09-25
Make Revise.juliadir more robust

### LINKED ISSUE #365: Making `Revise.juliadir` more robust
Hi,

We are using a shared server where we build `julia` from the source. The building happens under a privileged user account (e.g., `root` or some other admin account). This results in problems when trying to use `Revise`. The offending line, in our (cc: @martinbiel) opinion, is the following

https://github.com/timholy/Revise.jl/blob/568af230364f97bfc4c859aae7fb7cb5dd5d3dcf/src/Revise.jl#L149

where `isdir` throws `EACCESS`. Rewriting the `isdir` part in a `try ... catch` block should solve the problem, but we do not know whether there are other places that need a similar fix.

aytekinar:
Closes #365.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -146,7 +146,9 @@ This should be reliable even for local builds, cross-builds, and binary installs
 """
 const juliadir = begin
     local jldir = basebuilddir
-    if !isdir(joinpath(jldir, "base"))
+    try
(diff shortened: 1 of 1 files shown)


timholy:
The test error seems unrelated, so I will merge. Thanks for the PR!

PR #369, opened 2019-10-01
update to https://github.com/JunoLab/Atom.jl/pull/177

aviatesk:
update to https://github.com/JunoLab/Atom.jl/pull/177

Actually the `evalrepl` handler hasn't been used in recent Juno versions, thus I think we even don't need version compat checks for this change.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -1115,11 +1115,13 @@ end
 
 function setup_atom(atommod::Module)::Nothing
     handlers = getfield(atommod, :handlers)
-    for x in ["eval", "evalall", "evalrepl"]
-        old = handlers[x]
-        Main.Atom.handle(x) do data
-            revise()
-            old(data)
+    for x in ["eval", "evalall", "evalshow", "evalrepl"]
+        if haskey(handlers, x)
+            old = handlers[x]
+            Main.Atom.handle(x) do data
+                revise()
+                old(data)
+            end
         end
     end
     return nothing


aviatesk:
I'm not sure why `evalshow` hasn't been added to this, but AFAICT the handler _can be called_ in the current Juno version, thus I added that instead.

But maybe I'm missing some point.

src/Revise.jl
```diff
@@ -1115,7 +1115,7 @@ end
 
 function setup_atom(atommod::Module)::Nothing
     handlers = getfield(atommod, :handlers)
-    for x in ["eval", "evalall", "evalrepl"]
+    for x in ["eval", "evalall", "evalshow"]
         old = handlers[x]
```
timholy: For safety, can you change this to
```julia
        if haskey(handlers, "x")
            old = handlers[x]
            ...
```
Since `"evalrepl"` still exists, is there any reason to exclude it from the list?

aviatesk: Sure.

> Since "evalrepl" still exists, is there any reason to exclude it from the list?

The reasoning was that `evalrepl` isn't used in recent Juno versions.
But someone might use Revise in the old Juno versions, so this safety check would be better.

***

Well, why didn't you included `evalshow` within the previous implementation ?
Any reasoning behind that ?

pfitzseb: I think I just forgot to update Revise when I added the `evalshow` handler :)

timholy:
Thanks @aviatesk! Great to have your help with this.

PR #370, opened 2019-10-04
[DOCS] Fixed default startup script recommendation.

rapus95:
According to function reference this seems like a forgotton change. Also, without that change the JULIA_REVISE flag is probably not obeyed.

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -10,7 +10,8 @@ atreplinit() do repl
     try
         @eval using Revise
         @async Revise.async_steal_repl_backend()
-    catch
+    catch err
+        @error err
     end
 end
 ```
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -9,7 +9,7 @@ adding the following to your `.julia/config/startup.jl` file:
 atreplinit() do repl
     try
         @eval using Revise
-        @async Revise.wait_steal_repl_backend()
+        Revise.async_steal_repl_backend()
     catch
     end
 end
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -9,7 +9,7 @@ adding the following to your `.julia/config/startup.jl` file:
 atreplinit() do repl
     try
         @eval using Revise
-        Revise.async_steal_repl_backend()
+        @async Revise.async_steal_repl_backend()
     catch
     end
 end


timholy:
Thanks for the PR! I suppose this is probably better, though now I regret that function isn't named `async_maybe_steal_repl_backend` since it's not certain that it will steal the backend (depending on the setting of `JULIA_REVISE`). I am going to think about this a bit before merging.

timholy:
Actually, I don't think this works. If you comment out the `try/catch`, do you get the following error?
```julia
MethodError: no method matching async_steal_repl_backend()
The applicable method may be too new: running in world age 26056, while current world is 26061.
Closest candidates are:
  async_steal_repl_backend() at /home/tim/.julia/dev/Revise/src/Revise.jl:1020 (method too new to be called from this world context.)
```
The issue is that `atreplinit` takes a closure, and that closure gets defined before Revise is loaded. Hence the world age problem. One could use `invokelatest` but I think `@async` may be easier.

rapus95:
Damn I indeed get that problem aswell.
```julia
    try
        @eval using Revise
        @async Revise.async_steal_repl_backend()
    catch
    end
```
works as you predicted. Thus changed to that now.

pfitzseb:
```
    try
        @eval using Revise
        @async Revise.async_steal_repl_backend()
    catch err
        @error err
    end
```
would make sense imho.

timholy:
I've had it print errors locally in the past and found I didn't like it. But I will run locally with this for a while and see what I think.

ExpandingMan:
I have this
```julia
atreplinit() do repl
    try
        @eval using Revise
    catch e
        @warn("unable to load Revise", e)
    end
end
```
and to me, that's the ideal Revise startup script.

m-wells:
Maybe a slight tweak?
```julia
atreplinit() do repl
    try
        @eval using Revise
    catch e
        @warn(e.msg)
    end
end
```

Gives
```
┌ Warning: Package Revise not found in current path:
│ - Run `import Pkg; Pkg.add("Revise")` to install the Revise package.
└ @ Main ~/.julia/config/startup.jl:6
```

rather than
```
┌ Warning: unable to load revise
│   e =
│    ArgumentError: Package Revise not found in current path:
│    - Run `import Pkg; Pkg.add("Revise")` to install the Revise package.
│
└ @ Main ~/.julia/config/startup.jl:6
```

timholy:
At this point most users don't even need to steal the REPL.

PR #372, opened 2019-10-08
2.2.1

pfitzseb:
To [ensure compatibility with Juno 0.11](https://discourse.julialang.org/t/ann-juno-0-11/29644/2?u=pfitzseb).


timholy:
Oh shoot, I didn't notice this before seeing the discourse post. I already added this in 8a0ffe086e3a8bcc3752f644d81360fb69aecbb9 (though it's not "taking").

timholy:
BTW, super excited about the new Juno! Just keeps getting better.

PR #376, opened 2019-10-20
Update config to include auto install of Revise.jl

m-wells:
I am using [juliavm](https://github.com/pmargreff/juliavm) to manage multiple versions of `julia`.  When I switch versions the above script will auto install `Revise.jl`.  Using the auto install as above does not noticeably increase my subsequent startup times.

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -8,6 +8,8 @@ adding the following to your `.julia/config/startup.jl` file:
 ```julia
 atreplinit() do repl
     try
+        @eval using Pkg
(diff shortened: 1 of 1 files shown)


timholy:
I was a little skeptical about messing with packages, but upon further reflection this doesn't seem unreasonable. Let's try it and see what happens. Thanks for the contribution!

m-wells:
Excellent!  It is easy to forget to install Revise when switching julia versions and I find the above addition makes it a true drop it in and forget about it solution.

Issue #383, opened 2019-11-08
Revise doesn't work if I get a dependency warning


ExpandingMan:
With Revise 2.2.0 and Julia 1.3-rc4, Revise seemed to not be working at all.  After some experimenting, I find that it only seems to not work if I get the following warning
```julia
┌ Warning: Package Arrow does not have Debugger in its dependencies:
│ - If you have Arrow checked out for development and have
│   added Debugger as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Arrow
└ Loading Debugger into Arrow from project dependency, future warnings for Arrow are suppressed.
```
This warning is of course generated because I have a package loading which calls `using Debugger` when `Debugger` is not listed in its dependencies in its `Manifest.toml`.  It's pretty inconvenient for Revise not to work when I get this warning, because of course I deliberately do not have Debugger in my dependencies because it's only for development.  Is it possible to get Revise to work even though this warning is thrown?

ianshmean:
Thanks for confirming this. I had a suspicion it was happening!

timholy:
Duplicate of #371.

PR #385, opened 2019-11-09
Add fallback for stale cachefiles

### LINKED ISSUE #371: Revise stops working when using modules that are not listed as dependencies
Revise stops working if modules use a package that is not explicitly listed as a dependency in `Project.toml`.

Example (julia v1.1, started with `julia --project` from `MyPackage` folder):

```
module MyPackage

# this line breaks Revise!!
# ]add Colors solves the problem
using Colors

function foo()
    println("foo")
end

export foo

end # module
```
If `Colors` (or any other package) is not in the dependencies of `MyPackage`, then Revise stops working (silently, without raising an error or warning). Removing the line `using Colors` or adding it to the dependencies fixes the issue.

If it's not worth fixing, at least putting this information somewhere in the Docs would be very helpful.

### LINKED ISSUE #381: Revise >= v2.2.0 does not detect changes
Hello, Revise works fine until v2.1.10

After this version it doesn't take into account the changes in my projects

I am working with Julia 1.2.

I have attached ~/.julia/environments/v1.2/Manifest.toml and an example of project's Manifest.toml

[~/.julia/environments/v1.2/Manifest.toml](https://github.com/timholy/Revise.jl/files/3816150/environments_v1.2_Manifest.toml.txt)
[project's Manifest.toml](https://github.com/timholy/Revise.jl/files/3816156/project_Manifest.toml.txt)

timholy:
Fixes #371. See also #383. Might fix #381

### DIFF
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -59,17 +59,32 @@ function pkg_fileinfo(id::PkgId)
     # Try to find the matching cache file
     paths = Base.find_all_in_cache_path(id)
     sourcepath = Base.locate_package(id)
-    for path in paths
-        Base.stale_cachefile(sourcepath, path) === true && continue
-        provides, includes_requires = parse_cache_header(path)
-        mods_files_mtimes, _ = includes_requires
-        for (pkgid, buildid) in provides
-            if pkgid.uuid === uuid && pkgid.name == name
-                return path, mods_files_mtimes
-            end
+    if length(paths) > 1
+        fpaths = filter(path->Base.stale_cachefile(sourcepath, path) !== true, paths)
+        if isempty(fpaths)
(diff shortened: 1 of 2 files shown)


timholy:
xref #357. @KristofferC, do you think this will retrigger #356?

KristofferC:
If it tries to call ‘parse_cache_header‘ on a stale cache file, then yes.

timholy:
They can be stale for many reasons, here the reason is https://github.com/JuliaLang/julia/blob/72126021e612e91289fddb4b9aa01d4d9ec96ba9/base/loading.jl#L1445.

I've wrapped the `parse_cache_header` in a `try`/`catch`, hopefully that will solve the problem. The only other thing I can think to do is modify the Manifest, and that seems like a bad idea.

PR #232, opened 2019-01-08
Support adding new source files to a package

timholy:
This fixes both https://discourse.julialang.org/t/revise-with-requires/19347 and #107. While #107 is "transitory" and therefore not a huge problem, the case in the discourse post is a permanent state. Since I could fix both with the same infrastructure, this seemed worth prioritizing.

For usage with `Requires.jl`, the following is taken from the new documentation:

### docs

[Requires.jl](https://github.com/MikeInnes/Requires.jl) allows you to declare conditional dependencies for a package. When those dependencies result in the inclusion of new source files, Revise will not automatically track them. However, you can force Revise to track them by registering the file(s). An example registration might look like this:

```julia
function __init__()
    @require SomePackage="xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" begin
        include("newcode.jl")
        # Register newcode.jl with Revise
        @require Revise="295af30f-e4ad-537b-8983-00126c2a3abe" begin
            import .Revise
            Revise.add_file(TrackRequires, "src/newcode.jl")
        end
    end
end
```

The inner `@require` statement makes registration conditional on Revise, allowing your package to be used with and without Revise. In the `add_file` call (which is what performs the registration), note that you must use the relative path from the package top-level directory (here, `"src/newcode.jl"` rather than just `"newcode.jl"`).

!!! note
    In principle, Revise could parse the `__init__` function and perform this
    registration itself. However, this would lead to a performance problem:
    for this to work, Revise would have to "defensively" parse the top-level source file
    of every loaded package.
    Since Revise takes pains to parse only source files that
    have changes,
    we instead opt for manual registration of `@require` dependencies.

PR #387, opened 2019-11-11
Support new files added by Requires

### LINKED ISSUE #244: Compatibility with Requires.jl
Hi, I noticed that Revise does not track changes to code loaded by Requires.jl. This would be code defined as follows:
```
module MyModule
…
include("somecode.jl")    # This code is revised as expected
function __init__()
    @require SomeOptionalPackage="…" begin
        include("optional_code.jl")     # This code is not tracked by Revise
    end
end
end
```
I understand that the code being defined inside a macro inside `__init__`, this must be much more tricky, but still I'm not sure to properly understand:
- would that be technically possible, or is it known to be a limitation?
- If it is not possible, is there a way to reload the code manually without restarting julia? I tried `revise(MyModule)` as suggested on the "Limitations" page of the doc, but it does not work.

timholy:
Fixes #244. In contrast with #232, there are no manual interventions needed. CC @Djoop, @oschulz, @rafaqz

To work this requires a PR to Requires, https://github.com/MikeInnes/Requires.jl/pull/69.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,11 @@
 This file describes only major changes, and does not include bug fixes,
 cleanups, or minor enhancements.
 
+## Revise 2.4
+
+* Automatic tracking of methods and included files in `@require` blocks
+  (needs Requires 1.0.0 or higher)
+
 ## Revise 2.3
 
 * When running code (e.g., with `includet`), execute lines that "do work" rather than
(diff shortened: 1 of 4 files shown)


oschulz:
Awesome, thanks a lot!

timholy:
There are probably a bunch of packages that need their `[compat]` entries bumped for Requires `1`. Until that happens, you may not see this working as expected.

oschulz:
Oh, good to know, thanks!

PR #406, opened 2019-12-11
Wait for module initialization when tracking files

### LINKED ISSUE #403: "Revise is not watching" warning when using Plots.jl and __init__-function
Following module
```julia
module debug_plotsrevise
using Plots
function __init__()
	pyplot()
end
function hey()
	plot(rand(10))
end
end # module
```
gives a warning
```
julia> using debug_plotsrevise
julia> ┌ Warning: C:\Users\jaakkor2\.julia\packages\Plots\2KhB2\src\Plots.jl\ is not an existing directory, Revise is not watching
└ @ Revise C:\Users\jaakkor2\.julia\packages\Revise\2NCCI\src\Revise.jl:488
```
Versions I have are Julia v1.3.0, Revise v2.3.2, Plots v0.28.2, PyPlot v2.8.2.

Even an empty `__init__`-function in the module causes the warning.

I think the warning started with Plots v0.28.0 (but I could be mistaken).

timholy:
Fixes #403 

This PR also improves the precompile statements.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -677,6 +677,12 @@ function track(mod::Module, file::AbstractString; kwargs...)
         instantiate_sigs!(fm; kwargs...)
         id = PkgId(mod)
         if !haskey(pkgdatas, id)
+            # Wait a bit to see if `mod` gets initialized
+            # This can happen if the module's __init__ function
+            # calls `track`, e.g., via a @require. Ref issue #403.
+            sleep(0.1)
+        end
+        if !haskey(pkgdatas, id)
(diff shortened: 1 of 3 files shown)


lewisl:
This is not fixed for Julia 1.3.  Should it be?  Or only for Julia 1.4 (or something later than 1.3)?

Thanks.

timholy:
Doubtful that the Julia version is particularly relevant. For me the demo in #403 (the specific warning mentioned) is fixed with Revise 2.6.0. What happens for you?

lewisl:
Yes, it is.  Thank you.

PR #416, opened 2020-02-02
Modify recommended startup

### LINKED ISSUE #415: Suggested startup.jl gives a deprecation warning
The startup.jl as suggested here:
https://github.com/timholy/Revise.jl/blob/4297e4524accc47b31f5c8735e2b28f12f108b59/docs/src/config.md#using-revise-by-default

Gives a deprecation warning from Pkg since Pkg.installed is deprecated:
https://github.com/JuliaLang/Pkg.jl/blob/54eae6c1b591070918a65a581b7ab3a0c6d9d90c/src/Pkg.jl#L530-L540

This could be tweaked to use the new Pkg API on Julia 1.4. @KristofferC wondered, why not simply use something like

```julia
try
    using Revise
catch
    @warn "Revise is not installed"
end
```

Ref #370 which also suggests a startup.jl modification.

timholy:
Fixes #415

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,4 +2,4 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
-Documenter = "~0.23"
+Documenter = "~0.24"
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -8,14 +8,10 @@ adding the following to your `.julia/config/startup.jl` file:
 ```julia
 atreplinit() do repl
     try
-        @eval using Pkg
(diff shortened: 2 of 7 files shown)


timholy:
Passes for me locally (with https://github.com/MikeInnes/Requires.jl/pull/78). Let's see if just trying the latest nightly fixes this.

timholy:
#%^*@! Can anyone replicate the failure locally?

timholy:
Ah, it was triggered only by `coverage=true` in `Pkg.test` and due to https://github.com/JuliaLang/julia/pull/34254. This *should* fix it.

Awesome change BTW @vtjnash, look forward to seeing it change our coverage stats!

PR #425, opened 2020-02-15
Use composable AST transform on Julia 1.5

### LINKED ISSUE #423: use REPL ast_transforms in Julia 1.5+
In Julia 1.5 or later (https://github.com/JuliaLang/julia/pull/34626), you can just do
```jl
pushfirst!(Base.active_repl_backend.ast_transforms, my_transformation)
```
to tell the REPL to perform arbitrary AST transformations.  (`pushfirst!` will do `ast = my_transformation(ast)` on each parsed `ast` *before* the default `REPL.softscope` transformation in Julia 1.5.)

This should be simpler than your current `steal_repl_backend` machinery, and more importantly it is composable with other code that wants to hook into the REPL.

timholy:
Because the startup file loads before the REPL is available, we still need to wait for it to become available. Consequently I've kept the whole `wait_steal_repl_backend`, and made that do something different as appropriate. An advantage is that we don't need to tell folks to change their startup.jl files again.

Closes #423
CC @stevengj

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -1053,6 +1053,9 @@ function method_location(method::Method)
     return method.file, method.line
 end
 
+# On Julia 1.5.0-DEV.282 and higher, we can just use an AST transformation
+revise_first(ex) = Expr(:toplevel, :($revise()), ex)
+
 @noinline function run_backend(backend)
     while true
         tls = task_local_storage()
@@ -1081,11 +1084,16 @@ Replace the REPL's normal backend with one that calls [`revise`](@ref) before ex
 any REPL input.
 """
 function steal_repl_backend(backend = Base.active_repl_backend)
-    @async begin
-        # When we return back to the backend loop, tell it to start
-        # running our backend loop next, which differs only
-        # by processing the revision queue before evaluating each input.
-        put!(backend.repl_channel, (:($run_backend($backend)), 1))
+    if VERSION >= v"1.5.0-DEV.282"
+        # @warn "See Revise documentation for recommended configuration changes"
+        pushfirst!(backend.ast_transforms, revise_first)
+    else
+        @async begin
+            # When we return back to the backend loop, tell it to start
+            # running our backend loop next, which differs only
+            # by processing the revision queue before evaluating each input.
+            put!(backend.repl_channel, (:($run_backend($backend)), 1))
+        end
     end
     nothing
 end
@@ -1158,9 +1166,15 @@ function __init__()
     mode = get(ENV, "JULIA_REVISE", "auto")
     if mode == "auto"
         if isdefined(Base, :active_repl_backend)
-            steal_repl_backend(Base.active_repl_backend::REPL.REPLBackend)
+            if VERSION >= v"1.5.0-DEV.282"
+                pushfirst!(Base.active_repl_backend.ast_transforms, revise_first)
+            else
+                steal_repl_backend(Base.active_repl_backend::REPL.REPLBackend)
+            end
         elseif isdefined(Main, :IJulia)
             Main.IJulia.push_preexecute_hook(revise)
+        elseif VERSION >= v"1.5.0-DEV.282"
+            pushfirst!(REPL.repl_ast_transforms, revise_first)
         end
         if isdefined(Main, :Atom)
             setup_atom(getfield(Main, :Atom)::Module)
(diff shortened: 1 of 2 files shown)


src/Revise.jl
```diff
@@ -1040,6 +1040,9 @@ function method_location(method::Method)
     return method.file, method.line
 end
 
+# On Julia 1.5.0-DEV.260 and higher, we can just use an AST transformation
+revise_first(ex) = (revise(); return ex)
```
stevengj: Do you want this, or do you want `:($revise(); $ex)`?   i.e. don't you want to call `revise()` when the REPL actually executes the expression?   Or does it not matter?

stevengj: Or maybe something like `Expr(:toplevel, :($revise()), $ex)`, since e.g. `ex == :(import Foo)` can't work in a block.

stevengj:
Actually, you don't need to wait until the REPL starts up.   Just do:
```jl
pushfirst!(REPL.repl_ast_transforms, revise_first) # modify default transforms used for all new REPL instances
if isdefined(Base, :active_repl_backend)
    pushfirst!(Base.active_repl_backend.ast_transforms, revise_first) # modify running REPL, if any
end
```

src/Revise.jl
timholy: I think the only place where it matters is if the user edits a package providing another AST transform; with the version I have implemented, it will make updates in those packages immediately, including before they execute their transforms (for those packages that come later in the sequence than Revise). I *think* that's the behavior most people would want, so I'm inclined to leave it the way it is.

stevengj: I'm not sure if that use case is worth it for breaking the implicit interface that the AST transformations work by transforming the AST, rather than by side effects.    Suppose callers want to examine the transformed AST for other reasons than to execute it, for example.

timholy:
xref https://github.com/JuliaLang/julia/pull/34778. I'll put this on ice until that merges and we can get a version number.

src/Revise.jl
timholy: I wonder, though, if it would be better to move this to a `hooks` interface. But certainly I can live with the current interface, and I now agree with you that it's better to follow the implicit API.

timholy: @KristofferC, will OhMyREPL be able to work within the AST-transform interface or do you need side-effects?

KristofferC: OhMyREPal syntax highlights the text, it cannkt really do that with only AST transforms

timholy: The idea being that OhMyREPL will steal the REPL but if nothing else needs it, all is good? Should I also push to `REPL.repl_ast_transforms` or will you grab from `active_repl_backend.ast_transforms`?

KristofferC: > The idea being that OhMyREPL will steal the REPL but if nothing else needs it, all is good?

Yeah, pretty much. Not very nice of OhMyREPL but not sure how else I can do it.

Do I even have to do anything manually? The only thing OhMyREPL does is that it injects its own keybindings into the REPL but then it just calls the standard `on_done` methods. So hopefully that should just work?

timholy:
This seems like it's working. The attraction here is that the `startup.jl` file could be as simple as
```julia
using Revise
```
In the current version of the PR, I've taken an aggressive approach to getting people to update their `startup.jl` files. I'm not sure this is necessary; a gentler approach is to just advertise the new approach but let the old one continue to work without issuing a warning. In the fullness of time, this should result in the same outcome, albeit on a slower timescale. And perhaps it would make sense to be aggressive only after Julia 1.5 is widespread.

Thoughts? I'm happy to let this sit for a bit and collect opinions.

src/Revise.jl
```diff
@@ -1158,9 +1166,15 @@ function __init__()
...
     if mode == "auto"
         if isdefined(Base, :active_repl_backend)
-            steal_repl_backend(Base.active_repl_backend::REPL.REPLBackend)
+            if VERSION >= v"1.5.0-DEV.282"
+                pushfirst!(Base.active_repl_backend.ast_transforms, revise_first)
+            else
+                steal_repl_backend(Base.active_repl_backend::REPL.REPLBackend)
+            end
         elseif isdefined(Main, :IJulia)
             Main.IJulia.push_preexecute_hook(revise)
```
stevengj: (Eventually IJulia will just call the `REPL.repl_ast_transforms`, so you won't need this special case.)

timholy:
I decided on a less intrusive approach and will allow the migration to `using Revise` happen naturally, when more people use Julia 1.5+.

PR #426, opened 2020-02-15
Retry revisions on InterruptException

### LINKED ISSUE #418: try again automatically after interrupt
Sometimes a <kbd>Ctrl-C</kbd> after a command interrupts Revise, not the actual calculation. After this, the user repeatedly gets
```julia
^C┌ Error: Failed to revise /home/tamas/research/redacted/src/utilities.jl
│   exception = InterruptException:
└ @ Revise ~/.julia/packages/Revise/S7mrl/src/Revise.jl:590
┌ Warning: Due to a previously reported error, the running code does not match saved version for the following files:
│ 
│   /home/tamas/research/redacted/src/utilities.jl
│ 
│ Use Revise.errors() to report errors again.
└ @ Revise ~/.julia/packages/Revise/S7mrl/src/Revise.jl:639
```

Suggestion: special case `InterruptException` and try again.

timholy:
Fixes #418. Can you test @tpapp?

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -604,11 +604,13 @@ function revise()
     revision_errors = []
     finished = eltype(revision_queue)[]
     mexsnews = ModuleExprsSigs[]
+    interrupt = false
     for (pkgdata, file) in revision_queue
         try
             push!(mexsnews, handle_deletions(pkgdata, file)[1])
             push!(finished, (pkgdata, file))
         catch err
+            interrupt |= isa(err, InterruptException)
             push!(revision_errors, (pkgdata, file))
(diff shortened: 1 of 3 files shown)


tpapp:
Thanks for making a fix so quickly. I have been testing this today, and so far it looks like it resolves the issue.

That said, the issue itself is very hard to reproduce (one has to press Ctrl-C at the "right" time) so I will be working with this branch and report back in a few days.

tpapp:
Using this branch, I have not been able to trigger the issue in the past 4 days, using the same data and scripts that originally triggered it.

I think the issue is fixed by this PR.

PR #427, opened 2020-02-16
Fix revising baremodules

tkf:
It looks like Revise.jl can't handle baremodules as it is assuming `include` to exist in all modules.  This PR adds a simple fix and a test.


timholy:
Thanks for the fix!

PR #443, opened 2020-03-13
Support user-specified callbacks

tkluck:
This feature is one half of what `Revise.entr` already does. The function `entr` allows the user to install callbacks for when a file or module changes, and also enters a blocking loop that catches any errors that might arise from the callbacks.

This new `add_callback` function allows installing callbacks and keeping them around in the background. This commit also reimplements `entr` in terms of `add_callback` and `remove_callback`.

(Basically, `add_callback` now does what I thought `entr` would do after seeing its signature.)

I'm hoping to use this in [`HAML.jl`](https://github.com/tkluck/HAML.jl) for updating generated code when its source file changes. It may be similarly useful for hot code reloading in web servers, etc.

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -157,7 +157,15 @@ function remove_callback(key)
         delete!(cbs, key)
     end
     delete!(user_callbacks_by_key, key)
-    # TODO: stop watching these files
+
+    # possible future work: we may stop watching (some of) these files
+    # now. But we don't really keep track of what background tasks are running
+    # and Julia doesn't have an ergonomic way of task cancellation yet (see
+    # e.g.
+    #     https://github.com/JuliaLang/Juleps/blob/master/StructuredConcurrency.md
+    # so we'll omit this for now. The downside is that in pathological cases,
+    # this may exhaust inotify resources.
+
     nothing
 end
 
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -564,7 +564,7 @@ This is generally called via a [`Revise.Rescheduler`](@ref).
     latestfiles, stillwatching = watch_files_via_dir(dirname)  # will block here until file(s) change
     for (file, id) in latestfiles
         key = joinpath(dirname, file)
-        if key in keys(user_callbacks_by_file) # TODO: also do this for per-file watching
+        if key in keys(user_callbacks_by_file)
             union!(user_callbacks_queue, user_callbacks_by_file[key])
             notify(revision_event)
         end
@@ -603,6 +603,12 @@ function revise_file_queued(pkgdata::PkgData, file)
     end
 
     wait_changed(file)  # will block here until the file changes
+
+    if file in keys(user_callbacks_by_file)
+        union!(user_callbacks_queue, user_callbacks_by_file[file])
+        notify(revision_event)
+    end
+
     # Check to see if we're still watching this file
     dirfull, basename = splitdir(file)
     if haskey(watched_files, dirfull)
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -107,6 +107,18 @@ const user_callbacks_queue = Set{Any}()
 const user_callbacks_by_file = Dict{String, Set{Any}}()
 const user_callbacks_by_key = Dict{Any, Any}()
 
+"""
+    key = Revise.add_callback(f, files, modules=nothing; key=gensym())
+
+Add a user-specified callback, to be executed during the first run of
+`revise()` after a file in `files` or a module in `modules` is changed on the
+file system. In an interactive session like the REPL, Juno or Jupyter, this
+means the callback executes immediately before executing a new command / cell.
+
+You can use the return value `key` to remove the callback later
+(`Revise.remove_callback`) or to update it using another call
+to `Revise.add_callback` with `key=key`.
+"""
 function add_callback(f, files, modules=nothing; key=gensym())
     remove_callback(key)
 
@@ -134,6 +146,12 @@ function add_callback(f, files, modules=nothing; key=gensym())
     return key
 end
 
+"""
+    Revise.remove_callback(key)
+
+Remove a callback previously installed by a call to `Revise.add_callback(...)`.
+See its docstring for details.
+"""
 function remove_callback(key)
     for cbs in values(user_callbacks_by_file)
         delete!(cbs, key)
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -913,7 +913,6 @@ function entr(f::Function, files, modules=nothing; postpone=false, pause=0.02)
         sleep(pause)
         f()
     end
-    mycallbacks = [key]
     try
         while true
             wait(revision_event)
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -94,6 +94,74 @@ This list gets populated by callbacks that watch directories for updates.
 """
 const revision_queue = Set{Tuple{PkgData,String}}()
 
+const revision_event = Condition()
+
+"""
+    Revise.user_callbacks_queue
+
+Global variable, `user_callbacks_queue` holds `key` values for which the
+file has changed but the user hooks have not yet been called.
+"""
+const user_callbacks_queue = Set{Any}()
+
+const user_callbacks_by_file = Dict{String, Set{Any}}()
+const user_callbacks_by_key = Dict{Any, Any}()
+
+function add_callback(f, files, modules=nothing; key=gensym())
+    remove_callback(key)
+
+    files = map(abspath, files)
+    init_watching(files)
+
+    if modules !== nothing
+        for mod in modules
+            id = PkgId(mod)
+            pkgdata = pkgdatas[id]
+            for file in srcfiles(pkgdata)
+                absname = joinpath(basedir(pkgdata), file)
+                push!(files, absname)
+                track(mod, absname)
+            end
+        end
+    end
+
+    for file in files
+        cb = get!(Set, user_callbacks_by_file, file)
+        push!(cb, key)
+        user_callbacks_by_key[key] = f
+    end
+
+    return key
+end
+
+function remove_callback(key)
+    for cbs in values(user_callbacks_by_file)
+        delete!(cbs, key)
+    end
+    delete!(user_callbacks_by_key, key)
+    # TODO: stop watching these files
+    nothing
+end
+
+function process_user_callbacks!(keys = user_callbacks_queue; throw=false)
+    try
+        # use (a)sync so any exceptions get nicely collected into CompositeException
+        @sync for key in keys
+            f = user_callbacks_by_key[key]
+            @async Base.invokelatest(f)
+        end
+    catch err
+        if throw
+            rethrow(err)
+        else
+            @warn "[Revise] Ignoring callback errors" err
+        end
+    finally
+        empty!(keys)
+    end
+end
+
+
 """
     Revise.queue_errors
 
@@ -101,6 +169,8 @@ Global variable, maps `(pkgdata, filename)` pairs that errored upon last revisio
 `(exception, backtrace)`.
 """
 const queue_errors = Dict{Tuple{PkgData,String},Tuple{Exception, Any}}()
+
+const NOPACKAGE = PkgId(nothing, "")
 
 """
     Revise.pkgdatas
@@ -110,7 +180,7 @@ and julia objects, and allows re-evaluation of code in the proper module scope.
 It is a dictionary indexed by PkgId:
 `pkgdatas[id]` returns a value of type [`Revise.PkgData`](@ref).
 """
-const pkgdatas = Dict{PkgId,PkgData}()
+const pkgdatas = Dict{PkgId,PkgData}(NOPACKAGE => PkgData(NOPACKAGE))
 
 const moduledeps = Dict{Module,DepDict}()
 function get_depdict(mod::Module)
@@ -471,7 +541,7 @@ function init_watching(pkgdata::PkgData, files)
     end
     return nothing
 end
-init_watching(files) = init_watching(PkgId(Main), files)
+init_watching(files) = init_watching(pkgdatas[NOPACKAGE], files)
 
 """
     revise_dir_queued(dirname)
@@ -494,9 +564,16 @@ This is generally called via a [`Revise.Rescheduler`](@ref).
     latestfiles, stillwatching = watch_files_via_dir(dirname)  # will block here until file(s) change
     for (file, id) in latestfiles
         key = joinpath(dirname, file)
-        pkgdata = pkgdatas[id]
-        if hasfile(pkgdata, key)  # issue #228
-            push!(revision_queue, (pkgdata, relpath(key, pkgdata)))
+        if key in keys(user_callbacks_by_file) # TODO: also do this for per-file watching
+            union!(user_callbacks_queue, user_callbacks_by_file[key])
+            notify(revision_event)
+        end
+        if id != NOPACKAGE
+            pkgdata = pkgdatas[id]
+            if hasfile(pkgdata, key)  # issue #228
+                push!(revision_queue, (pkgdata, relpath(key, pkgdata)))
+                notify(revision_event)
+            end
         end
     end
     return stillwatching
@@ -520,6 +597,7 @@ function revise_file_queued(pkgdata::PkgData, file)
         sleep(0.1)  # in case git has done a delete/replace cycle
         if !file_exists(file)
             push!(revision_queue, (pkgdata, file0))  # process file deletions
+            notify(revision_event)
             return false
         end
     end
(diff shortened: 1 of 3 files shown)


timholy:
Just to see if I understand, the goal here is to allow more granular control: the ability to control behavior when errors are thrown, and to remove callbacks from the queue? If so, I'm all in favor.

However, this failed tests on every platform but Linux. One possible fix is another `sleep(mtimedelay)` before writing `"def"`.

tkluck:
> Just to see if I understand, the goal here is to allow more granular control: the ability to control behavior when errors are thrown, and to remove callbacks from the queue? If so, I'm all in favor.

Exactly. Happy that you're in favour, I'll look into the test failures then!

timholy:
I'm going to cycle CI on this and see how it does now that #450 is merged. There's a failure on 1.0 that I don't understand.

tkluck:
Still some CI failures on non-linux platforms. It's obviously important to support those as well.

Do you have any best practices that might help me tackle this? Myself, I'd be tempted to replace all reliance on `sleep` by adding a small API that creates a `Base.Event` that fires exactly when a certain change has been noticed by revise. But maybe/hopefully that's overkill and there's a better way?

timholy:
The `entr` tests have been a persistent low-grade problem, and no one has yet come up with a fix. If you've made it reliable, that's actually quite encouraging. While I agree that event-based handling would make it more reliable, to me that seems like cheating as no one will use events in their normal workflow. Since test failures occur reliably for `entr` and nowhere else, it suggests that a localized fix should be all that is required.

In the past it has been observed that magic dust occasionally helps: https://github.com/timholy/Revise.jl/blob/836d26134161d59dc41430f1d42694777655cfda/src/Revise.jl#L904.

Would someone with a mac or windows machine check out this branch and see if they can figure out the problem? I've been thinking @mlhetland should review this anyway.

mlhetland:
Hm. Just trying out the test more manually, my first observation is that the callback is not triggered unless the file is closed. However, if you (e.g.) replace `flush(io)` with `close(io)`, it seems to work just fine. Maybe that's acceptable? Or perhaps unavoidable? After all, that's how a text editor would behave, I guess? (I.e., a save would involve opening, writing and then closing the file?)

I'm not sure if the callback is supposed to be triggered automatically; it's only triggered once I call `revise`, but I suppose that's normal outside `entr`?

tkluck:
> Hm. Just trying out the test more manually, my first observation is that the callback is not triggered unless the file is closed. However, if you (e.g.) replace flush(io) with close(io), it seems to work just fine. Maybe that's acceptable? Or perhaps unavoidable? After all, that's how a text editor would behave, I guess? (I.e., a save would involve opening, writing and then closing the file?)

That's a great point. I pushed a commit that hopefully fixes it in the way you describe.

> I'm not sure if the callback is supposed to be triggered automatically; it's only triggered once I call revise, but I suppose that's normal outside entr?

Yes, that's the idea of these callbacks. If someone wants them to be called "asap" then they don't need `Revise` but just `FileWatching` + a background task. But using `Revise` ensures that the callbacks executes when the interactive user expects it.

My personal use case is loading custom template code in HAML.jl, but one might also imagine using this as (untested):
```julia
const COUPLING_CONSTANTS = Ref(read_json("coupling_constants.json"))
Revise.add_callback(["coupling_constants.json"]) do
    COUPLING_CONSTANTS[] = read_json("coupling_constants.json")
end
```
Then the user can be sure that the coupling constants don't change in mid-computation.

timholy:
For some reason Travis doesn't seem to be reporting back (not sure whether that's a Travis problem or a GitHub problem), but you can monitor it by navigating to https://travis-ci.org/github/timholy/Revise.jl, clicking on "More options," choosing "Requests," and then clicking on the build number for the most recent request. Looking very promising! I expect failure on non-Linux nightly due to https://github.com/JuliaLang/julia/issues/35297, so that's nothing to worry about.

src/Revise.jl
```diff
@@ -94,6 +94,74 @@ This list gets populated by callbacks that watch directories for updates.
...
+
+Global variable, `user_callbacks_queue` holds `key` values for which the
+file has changed but the user hooks have not yet been called.
+"""
+const user_callbacks_queue = Set{Any}()
+
+const user_callbacks_by_file = Dict{String, Set{Any}}()
+const user_callbacks_by_key = Dict{Any, Any}()
+
+function add_callback(f, files, modules=nothing; key=gensym())
```
timholy: This could use a docstring

src/Revise.jl
```diff
@@ -94,6 +94,74 @@ This list gets populated by callbacks that watch directories for updates.
...
+    remove_callback(key)
```
timholy: This is one valid option; another would be to throw if `key` has been previously used. I don't have strong feelings, but what was your thinking here?

src/Revise.jl
```diff
@@ -94,6 +94,74 @@ This list gets populated by callbacks that watch directories for updates.
...
+
+    files = map(abspath, files)
+    init_watching(files)
+
+    if modules !== nothing
```
timholy: This aspect of the functionality is not covered by tests.

src/Revise.jl
```diff
@@ -94,6 +94,74 @@ This list gets populated by callbacks that watch directories for updates.
...
+        for mod in modules
+            id = PkgId(mod)
+            pkgdata = pkgdatas[id]
+            for file in srcfiles(pkgdata)
+                absname = joinpath(basedir(pkgdata), file)
+                push!(files, absname)
+                track(mod, absname)
```
timholy: It only recently became safe to call `track` more than once on the same file, so this seems OK. But, under what circumstances would these not already be tracked?

src/Revise.jl
```diff
@@ -94,6 +94,74 @@ This list gets populated by callbacks that watch directories for updates.
...
+
+    return key
+end
+
+function remove_callback(key)
+    for cbs in values(user_callbacks_by_file)
+        delete!(cbs, key)
+    end
+    delete!(user_callbacks_by_key, key)
+    # TODO: stop watching these files
```
timholy: Worth contemplating what happens if two callbacks get registered on an overlapping set of files.

src/Revise.jl
tkluck: Pushing that now.

src/Revise.jl
tkluck: My thinking was to make the operation idempotent. I'm thinking this will be used in situations immediately after loading some data / code from a file in `files`. The operation of loading that data is presumably idempotent so it would be nice if adding `add_callback` right after the idempotent operation is idempotent as well. What do you think?

src/Revise.jl
tkluck: Good catch. Added.

src/Revise.jl
tkluck: I don't actually see any reason why they would already be tracked? What if someone does `Revise.add_callback([], Base) do ....` ? I'm probably missing something here.

src/Revise.jl
tkluck: I just pushed a commit that just leaves the watches hanging around (even if they won't trigger callbacks any more). I added my thinking as a comment to the code:
```
# possible future work: we may stop watching (some of) these files
# now. But we don't really keep track of what background tasks are running
# and Julia doesn't have an ergonomic way of task cancellation yet (see
# e.g.
#     https://github.com/JuliaLang/Juleps/blob/master/StructuredConcurrency.md
# so we'll omit this for now. The downside is that in pathological cases,
# this may exhaust inotify resources.
```
What do you think?

tkluck:
Looks like tests are passing on all platforms now :grinning: There's only an unrelated line-number issue on Julia nightly. Anything I can still improve?

mlhetland:
I haven't done a thorough review, but I've looked at the code, and it looks good to me! I also think this make the whole idea of callbacks a more natural part of Revise; `entr` on its own might seem like more of an outlier (as witnessed by @timholy's initial suggestion of placing it in a separate package).

I guess we might consider renaming `entr` as well (through a deprecation cycle, or just keeping the old name as a shortcut), somehow reflecting how it's now more of a callback + file monitoring, rather than an atomic piece of functionality. I originally took the name from the [entr project](http://eradman.com/entrproject/) (where it's short for *Event Notify Test Runner*… 🤷‍♂). Of course, using that name – while behaving similarly to that utility – makes the functionality potentially more predictable to those who know of it, but I'm not sure that's a weighty argument.

src/Revise.jl
tkluck: To make it more explicit, and to improve the example snippet from before, here's how I'd use it:
```julia
const COUPLING_CONSTANTS = Ref(read_json("coupling_constants.json"))
Revise.add_callback(["coupling_constants.json"], key=:COUPLING_CONSTANTS) do
    COUPLING_CONSTANTS[] = read_json("coupling_constants.json")
end
```
so it doesn't matter how often this code executes (in particular, maybe it executes in response to one of these callbacks!). It would be even better to have the key depend on `@__MODULE__` in some way, or whatever else makes it a true global primary key in the user's situation.

tkluck:
Thanks for adding your comments @mlhetland ! I also understand your point about the naming: I was one of the users that didn't know about the entr project so I was a bit confused about it :sweat_smile: 

I have no real opinion on whether we should also rename or add an alias to `entr` but I do agree with your reasoning.

src/Revise.jl
tkluck: Ah, now I see what you mean! Yes, this is faulty code then -- I'm trying to guard for the fact that the module may not be under tracking at all yet. Under that logic, we should _first_ call `track(mod)` and only then read the data from `pkgdata`, right?

If that's the case, I should also add a doctest that tests for this correct behaviour.

src/Revise.jl
```diff
@@ -792,39 +908,22 @@ This will print "update" every time `"/tmp/watched.txt"` or any of the code defi
...
-                push!(files, joinpath(basedir(pkgdata), file))
-            end
-        end
+    postpone || f()
+    key = add_callback(files, modules) do
+        sleep(pause)
+        f()
     end
-    active = true
+    mycallbacks = [key]
```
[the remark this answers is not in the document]
tkluck: Ah, that looks like an artefact of earlier iterations. Will remove.

src/Revise.jl
timholy: Any time you've said `using MyPkg` it should be tracked. Though perhaps the `watch_package` callback (see how it gets registered in `__init__`) has to fire first. Is that the issue?

I *think* I've make `track` safe for re-entry (see #433), but I tend to be jittery (probably, needlessly) about things getting done twice and messing up the internal accounting. Maybe poke at this a bit first and try to define more clearly what circumstances require this?

src/Revise.jl
timholy: For any task that needs rescheduling it could of course just fail to reschedule, but we would have to contend with one more execution. I'm fine with leaving it this way, at least until people complain (this is not functionality I tend to use a lot myself).

src/Revise.jl
tkluck: > Any time you've said `using MyPkg` it should be tracked. 
The two things that I have in mind are:
 - packages loaded before `Revise` was loaded, e.g.
```
using MyPackage
using Revise
Revise.add_callback([], [MyPackage]) do ....
```
 - packages in the system image (e.g. `Base` or anything the user added in `PackageCompiler`).

The first one could potentially be tested by spawning another Julia process. The second one seems almost impossible to test. But in both cases, calling `Revise.track(<module>)` _before_ reading `pkgdata` seems like it's the right thing to do -- what do you think?

I wasn't thinking about a race condition related to when `watch_package` is being called; do you think we should?

timholy: > packages loaded before Revise was loaded

If package precompilation were required for all modules, I could support that, but the existence (even in principle) of non-precompiled packages makes robust support for such a feature impossible. The problematic use case is loading the package, making some changes to the source, and then tracking with Revise: in this case Revise has no way of knowing that what's in the files does not match the code in your running session. Conversely, if you load Revise first, it "defensively" parses the entire package when it's loaded and then it has a baseline to compare any file changes. This is not necessary for precompied packages because I added a source-cache to `*.ji` files in https://github.com/JuliaLang/julia/pull/23898, and did the same for Julia itself in https://github.com/JuliaLang/julia/pull/24120. This means that you can always recover the state of the source at the time at which the `*.ji` file was built. But there is no such fix available for non-precompiled packages.

So I plan to support this feature by saying "load Revise first, otherwise you're screwed" :smile:. More seriously, I could contemplate supporting this specifically for precompiled packages, but we'd need to enforce it.

> packages in the system image (e.g. Base or anything the user added in PackageCompiler).

That's a much more important use-case. Do such files record their source text? If not, we're screwed for now but should fix that ASAP.

> But in both cases, calling Revise.track(<module>) before reading pkgdata seems like it's the right thing to do -- what do you think?

Yeah, `track` populates the structure and then you can do stuff with it.

ffevotte:
FYI, it looks like this PR also fixes #481, for which I submitted a (much less ambitious) fix in #482 

Looking at what's proposed here, I do think that my PR should be ignored in favor of this proposal. Please do not hesitate to ask me if I can help in any way. And thanks for everything!

ffevotte:
And for the record, it looks like this PR also fixes #470 by not relying on `FileWatching.watch_folder` anymore.

timholy:
I'd love to see this merged. @ffevotte, perhaps @tkluck would like some help pushing it over the finish line?

tkluck:
Yes, I'd appreciate the help! Sorry for not finishing + pushing this sooner -- I just started at a new employer and need to jump through a few legal hoops before I can continue contributing to open source again. Shouldn't be a problem but it takes some time. So feel free to beat me to it and finish this one :)

ffevotte:
OK, I'll try to re-read this discussion, paying more attention to details and trying to determine what's missing for this to get merged.

But if the next steps are clear in your mind and you can spend a few minutes listing them here, I'd greatly appreciate it!

src/Revise.jl
tkluck: @ffevotte this is the thing that needs resolving: in `add_callback` we need to move the call to `Revise.track(mod)` to the place _before_ reading `pkgdatas[id]`. We don't have to call `Revise.track(mod, absname)` after that.

After that change, hopefully this is done.

tkluck:
@ffevotte thank you! Just tagged you in the conversation that has an actionable.

ffevotte:
OK, I started working on this, but am not sure how to proceed to add commits to this PR. Should I make a PR from my working branch (https://github.com/ffevotte/Revise.jl/tree/tkluck/support-callbacks) to this one?

In any case, here is what I have so far:
1. ffevotte@9233aa986a356a8894eb8c2ba3be50f077415dce: the last required change, as I understand it
1. ffevotte@2fadde7315b8959604a97a12e507e81c62cf4285: some additional docstrings, in the hope that they can help future contributors (or my future self) better understand the implementation of this feature 
1. ffevotte@50a05c3b5bdf54ed8afc906b2be38d5dcd02c389: merge from the current master and conflict resolution. There have been significant changes in master since this branch has been created, which caused several conflicts. In particular, I'm not sure I correctly handled the new version of [`revise_file_queued`](https://github.com/ffevotte/Revise.jl/blob/50a05c3b5bdf54ed8afc906b2be38d5dcd02c389/src/Revise.jl#L646), especially because of conflicts related to 9c580e498df7100593022d26d45d441bd32d5ba9:
    - [l.662](https://github.com/ffevotte/Revise.jl/blob/50a05c3b5bdf54ed8afc906b2be38d5dcd02c389/src/Revise.jl#L662): file deletions seem to not be explicitly processed anymore by Revise, but I still think we should notify `entr` here.
    - [l.679](https://github.com/ffevotte/Revise.jl/blob/50a05c3b5bdf54ed8afc906b2be38d5dcd02c389/src/Revise.jl#L679): I'm not sure I fully understand the logic here, but I think no notification is needed here, since sending notifications to files watched by `entr` should have already been taken care of a [few lines above](https://github.com/ffevotte/Revise.jl/blob/50a05c3b5bdf54ed8afc906b2be38d5dcd02c389/src/Revise.jl#L674)

@timholy @tkluck I think it would be best if you could carefully review this: I fear I'm too new to the Revise codebase to be 100% sure of what I did; I might very well have made a few mistakes (especially regarding that last merge...)

(Tests seem to pass on my system (Linux), but `revise_file_queued` does not seem to be called on this platform)

timholy:
Thanks @ffevotte! Submitting a fresh PR that preserve authorship would be good. I suspect we're not going to want the full version history, so perhaps squashing all of @tkluck's into a single commit, and then squashing your own changes into a different commit right before we merge your PR makes the most sense.

I like your changes. I *think* the file deletion is actually OK, with the new watching logic the only part of it that it has to cover (?) is whether it should keep watching the file/directory. But since there's no CI for BSD, `revise_file_queued` is not well tested, perhaps we should do a separate run on Travis with `watching_files[] = true`. No BSD-users have complained, though...

ffevotte:
Thanks for the review! Is something like #488 what you had in mind?

tkluck:
Closing this one now #488 has been merged. Thank you so much @ffevotte !

PR #444, opened 2020-03-19
Document the different `.julia/config/startup.jl` formulation for VS Code

DilumAluthge:
In order to load Revise.jl by default in VS Code, you need to use a slightly different approach than the default one. Full details here: https://github.com/julia-vscode/julia-vscode/wiki/Known-issues-and-workarounds

This pull request adds this to the Revise.jl documentation.

cc: @timholy 
cc: @davidanthoff

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -26,6 +26,20 @@ catch e
 end
 ```
 
+If you are using Visual Studio Code, you should instead add the following to your `.julia/config/startup.jl` file: (full details [here](https://github.com/julia-vscode/julia-vscode/wiki/Known-issues-and-workarounds))
+
+```julia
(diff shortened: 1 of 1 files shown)


davidanthoff:
I think that documentation is probably outdated. We now have a configuration option in VS Code to load Revise, so ideally one wouldn't do anything in a startup file... But I'm not sure.

timholy:
Sorry to be slow to respond, COVID-19 shutdown of normal lab/graduate program/department operations has taken quite a lot of time.

I'll wait for greater clarity about what's required. FWIW the default seems to work for me on VSCode, but I'm only recently back to using it now that the Language Server has gotten *so* much more stable. (Thanks @davidanthoff!)

PR #445, opened 2020-03-20
intercept REPL without replacing the backend Task

JeffBezanson:
This allows us to use the root task as the REPL backend.

Fixes https://github.com/JuliaLang/julia/issues/35185


timholy:
Thanks Jeff. Test failure is unrelated.

timholy:
Should also xref #425

PR #449, opened 2020-03-25
Rescheduler: don't silently stop when errors are encountered

tkluck:
This was helpful during work on #443 to make sure my typos didn't pass silently. Maybe it's helpful for others too?

These thunks should be quite reliable (e.g., they don't contain user-specified code) but during development it's nice to have visibility on errors and keep the background workers functioning.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -242,7 +242,12 @@ struct Rescheduler{F,A}
 end
 
 function (thunk::Rescheduler{F,A})() where {F,A}
-    if thunk.f(thunk.args...)::Bool
-        schedule(Task(thunk))
+    reschedule = true
+    try
+        reschedule = thunk.f(thunk.args...)::Bool
+    catch err
+        @warn "[Revise] Error during a background task" err
+    finally
+        reschedule && schedule(Task(thunk))
     end
 end


timholy:
Probably a good idea, I'm interested to see what issues this might catch. You'll have to handle some circumstances, like file deletion, specially.

tkluck:
The CI failures are only on julia-nightly and are caused by https://github.com/JuliaLang/julia/pull/34660 . I think this PR is probably good to merge.

tkluck:
> You'll have to handle some circumstances, like file deletion, specially.

Wait, maybe I missed what you were trying to say here. You mean that some of the `Rescheduler`s actually _should_ silently exit if an error occurs? I completely missed that. Could you give me a pointer?

timholy:
> The failures are only on nightly

Check travis, you'll see *lots* of

```
┌ Warning: [Revise] Error during a background task
│   err = IOError: FileMonitor (start): no such file or directory (ENOENT)
└ @ Revise ~/build/timholy/Revise.jl/src/types.jl:249
```

The tests actually pass but that seems to cause Travis to fail to realize that, so basically all tests are marked as failing. I can't really merge this under those circumstances.

tkluck:
Ah thanks for pointing that out! That was my confirmation bias at work. Let me see what I can do.

tkluck:
I'm sorry, I'm going to close this pull request because I don't think I'll realistically have time to finish it. But I'm very excited that #488 got merged :-)

PR #460, opened 2020-04-20
dont assume we can find source code for loaded packages

KristofferC:
This can happen when Revise is put into a sysimage and it's dependencies are no longer in the Project/Manifest (as happens in https://github.com/JuliaLang/PackageCompiler.jl/issues/391)

### DIFF
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -94,7 +94,7 @@ function pkg_fileinfo(id::PkgId)
     # Try to find the matching cache file
     paths = Base.find_all_in_cache_path(id)
     sourcepath = Base.locate_package(id)
-    if length(paths) > 1
+    if sourcepath !== nothing && length(paths) > 1
         fpaths = filter_valid_cachefiles(sourcepath, paths)
         paths = isempty(fpaths) ? paths : fpaths
         # Work-around for #371 (broken dependency prevents tracking):
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -81,6 +81,7 @@ end
 # that's easier to precompile. (This is a hotspot in loading Revise.)
 function filter_valid_cachefiles(sourcepath, paths)
     fpaths = String[]
+    sorcepath === nothing && return fpaths
     for path in paths
         if Base.stale_cachefile(sourcepath, path) !== true
             push!(fpaths, path)
@@ -94,7 +95,7 @@ function pkg_fileinfo(id::PkgId)
     # Try to find the matching cache file
     paths = Base.find_all_in_cache_path(id)
     sourcepath = Base.locate_package(id)
-    if sourcepath !== nothing && length(paths) > 1
+    if length(paths) > 1
         fpaths = filter_valid_cachefiles(sourcepath, paths)
         paths = isempty(fpaths) ? paths : fpaths
         # Work-around for #371 (broken dependency prevents tracking):


src/pkgs.jl
```diff
@@ -94,7 +94,7 @@ function pkg_fileinfo(id::PkgId)
     # Try to find the matching cache file
     paths = Base.find_all_in_cache_path(id)
     sourcepath = Base.locate_package(id)
-    if length(paths) > 1
+    if sourcepath !== nothing && length(paths) > 1
```
timholy, line 97: If `sourcepath === nothing` is there ever the possibility that `length(paths) > 1`? And if so, is there any risk we could grab the wrong `path`?

KristofferC: I tweaked it slightly to make it behave more similar to the normal case. I am not 100% sure what the code here does, I just know that it doesn't handle the case where `locate_package` returns `nothing` which is a possibility as the issue above shows.

KristofferC:
For testing, you can force it with something like (assuming you have multiple precompile files stored for Example):

```

julia> using Revise

julia> using Example

julia> empty!(LOAD_PATH)
0-element Array{String,1}

julia> Revise.pkg_fileinfo(Base.PkgId(Example))
ERROR: MethodError: no method matching stale_cachefile(::Nothing, ::String)
Closest candidates are:
  stale_cachefile(::String, ::String) at loading.jl:1396
Stacktrace:
 [1] filter_valid_cachefiles(::Nothing, ::Array{String,1}) at /Users/kristoffercarlsson/.julia/packages/Revise/C272c/src/pkgs.jl:85
 [2] pkg_fileinfo(::Base.PkgId) at /Users/kristoffercarlsson/.julia/packages/Revise/C272c/src/pkgs.jl:98
 [3] top-level scope at REPL[5]:1
 [4] eval(::Module, ::Any) at ./boot.jl:331
 [5] eval_user_input(::Any, ::REPL.REPLBackend) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [6] run_backend(::REPL.REPLBackend) at
```

but I am not sure how to cause it to happen in normal usage

timholy:
I guess we should just go ahead with this, since that test is pretty artificial and would presumably fail at a slightly later line.

christopher-dG:
Could we make a release with this commit? I am also dealing with the PackageCompiler issue 🙂

timholy:
I was hoping to add something else, but let's get this out as-is.

christopher-dG:
Ah i see. I just wanted to make sure it wasn't going to get forgotten!

timholy:
Ugh, I should have read this more carefully. How on earth did the tests pass, I am really confused...

timholy:
https://travis-ci.org/github/timholy/Revise.jl/builds/677646144 claims it's the right commit, 15a561a, and shows all tests passing. Yet when I try it locally I get
```
Test Summary:                             | Pass  Error  Broken  Total
Revise                                    |  416      8       1    425
  PkgData                                 |    1                     1
  LineSkipping                            |    8                     8
  Equality and hashing                    |    5                     5
  Parse errors                            |    4                     4
  Signature extraction                    |    4                     4
  Comparison and line numbering           |   75              1     76
  Display                                 |    4                     4
  File paths                              |    6      1              7
  Base & stdlib file paths                |    5                     5
  Recursive types (issue #417)            |    1                     1
  Cross-module extension                  |    6                     6
  @__FILE__                               |    2                     2
  Module docstring                        |    6                     6
  Undef in docstrings                     |   72                    72
  Macro docstrings (issue #309)           |    4                     4
  Changing @inline annotations            |   16                    16
  Revising macros                         |    7                     7
  More arg-modifying macros               |    4                     4
  Line numbers                            |    5                     5
  Line numbers in backtraces and warnings |    6                     6
  New submodules                          |    3                     3
  Timing (issue #341)                     |    3                     3
  Method deletion                         |   58                    58
  Evaled toplevel                         |    4                     4
  Revision errors                         |   37                    37
  Retry on InterruptException             |   17                    17
  get_def                                 |    3      1              4
  Pkg exclusion                           |    3                     3
  Manual track                            |   17      1             18
  Auto-track user scripts                 |    4                     4
  Distributed                             |   15                    15
  Git                                     |    6                     6
  Recipes                                 |           1              1
  CodeTracking #48                        |    1                     1
  Methods at REPL                         |    2      4              6
  baremodule                              |    2                     2
ERROR: LoadError: Some tests did not pass: 416 passed, 0 failed, 8 errored, 1 broken.
in expression starting at /home/tim/.julia/dev/Revise/test/runtests.jl:71
```
as well as an error upon startup.

Explaining it as a weird Travis error also doesn't entirely make sense, because AppVeyor did the same thing: https://ci.appveyor.com/project/timholy/revise-jl/builds/32330643. Can anyone make sense of this?

PR #462, opened 2020-04-22
Fix up typo in filter_valid_cachefiles

angusmoore:
This line threw an UndefVarError when I hit it, so I'm guessing this is just a simple typo fix.


angusmoore:
I don't understand the test failure (sorry, Revise is not a codebase I'm familiar with). But it's only one architecture & version failing, so maybe spurious?

KristofferC:
😭 My bad...

timholy:
Thanks for the quick fix. @KristofferC, no worries, I'm much more concerned about the inexplicable failure of the CI services to catch this (https://github.com/timholy/Revise.jl/pull/460#issuecomment-618290966).

PR #464, opened 2020-04-23
try figure out why Travis passed a buggy PR [DO NOT MERGE]

KristofferC:


timholy:
Good idea, I should have tried that myself...

KristofferC:
Hmm, seems to pass here as well.

KristofferC:
Okay, it isn't stranger than that this branch is not taken:

https://github.com/timholy/Revise.jl/blob/66380df2499700a8af6719f84b4562e74aa3219a/src/pkgs.jl#L98

timholy:
Ah, very good! Thanks for figuring that out.

timholy:
xref #465

PR #471, opened 2020-04-30
Fix #470

### LINKED ISSUE #470: Bug in `entr` call to `watch_folder`

```julia
```
As the documentation says:

bogiebro:

```julia
```
As the documentation says:

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -781,16 +781,16 @@ end
 includet(file::AbstractString) = includet(Main, file)
 
 """
-    entr(f, files; postpone=false, pause=0.02)
-    entr(f, files, modules; postpone=false, pause=0.02)
+    entr(f, paths; postpone=false, pause=0.02)
+    entr(f, paths, modules; postpone=false, pause=0.02)
 
-Execute `f()` whenever files listed in `files`, or code in `modules`, updates.
-`entr` will process updates (and block your command line) until you press Ctrl-C.
-Unless `postpone` is `true`, `f()` will be executed also when calling `entr`,
-regardless of file changes. The `pause` is the period (in seconds) that `entr`
-will wait between being triggered and actually calling `f()`, to handle
-clusters of modifications, such as those produced by saving files in certain
-text editors.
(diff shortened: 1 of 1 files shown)


src/Revise.jl
```diff
@@ -819,9 +819,13 @@ function entr(f::Function, files, modules=nothing; postpone=false, pause=0.02)
...
             postpone || f()
             for file in files
-                waitfor = isdir(file) ? watch_folder : watch_file
+                is_file_dir = isdir(file)
                 @async while active
-                    ret = waitfor(file, 1)
+                    if is_file_dir
+                        ret = watch_folder(file, 1)[2]
+                    else
+                        ret = watch_File(file, 1)
```
timholy, line 827:
```suggestion
                        ret = watch_file(file, 1)
```

mlhetland:
> CC @mlhetland, since he's kind of the de facto expert on `entr`.

Heh, more a fan than an expert, but thanks :-) Seems like a good idea to me. (The [original `entr`](http://eradman.com/entrproject/) also has the ability to watch directories. It requires a specific flag, `-d`, but I see no need for that in our case, really, as we'll be pretty explicit about supplying the directory name, anyway?)

bogiebro:
I updated the docstrings, and I'm happy to add a test too. Is there any way to run a subset of tests in the suite, rather than of all of them (as in Pkg.test)?

timholy:
Sorry for the delay, I have been very distracted lately. Yes, you can run a subset of tests:
- julia 1.3 or higher: `Pkg.test("Revise", test_args=["Method deletion"])` (you can add more than one testset)
- from the shell prompt, if you have all test dependencies in your environment: `julia runtests.jl "Method deletion"`
- from an interactive session: `push!(ARGS, "Method deletion"); include("runtests.jl")`

ffevotte:
Just to tidy things up and let everyone interested know:
- @bogiebro: #470 should be fixed, now that #500 has been merged. Thanks for getting this started!
- @timholy: I think this PR could be closed now

PR #473, opened 2020-05-01
Fix on latest nightly

timholy:


timholy:
The most important work (that went well beyond fixing nightly) was actually done in #474.

Issue #481, opened 2020-05-16
Error handling for `entr` when watching multiple files


ffevotte:
When watching multiple files with `entr`, and the user-provided function throws an error, `entr` does not immediately exit as I think it is intended to.

Steps to reproduce:
```julia
julia> import Revise
julia> import Dates

# Create two empty files
julia> open((f)->nothing, "/tmp/file1", "w")
julia> open((f)->nothing, "/tmp/file2", "w")

#... and watch them
julia> Revise.entr(["/tmp/file1", "/tmp/file2"]; postpone=true) do
           println("Something changed: ", Dates.now())
           error("KO")
       end

# Touch file1 from outside julia
# -> the user-provided function runs
Something changed: 2020-05-16T13:13:37.046
# -> the error has been thrown, but nothing seems to happen yet
# -> entr does not return

# Touch file1 again
# -> nothing happens; the user-provided function is not called

# Later touch file2
Something changed: 2020-05-16T13:14:00.926
# -> now the errors are shown and `entr` exits
ERROR: TaskFailedException:
KO
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] (::var"#9#10")() at ./REPL[5]:3
 [3] #invokelatest#1 at ./essentials.jl:712 [inlined]
 [4] invokelatest at ./essentials.jl:711 [inlined]
 [5] macro expansion at /home/francois/.julia/packages/Revise/AMRie/src/Revise.jl:828 [inlined]
 [6] (::Revise.var"#81#82"{Float64,var"#9#10",typeof(FileWatching.watch_file),String})() at ./task.jl:358

...and 1 more exception(s).
```

I think this happens because one `@async` task is spawned for each watched file, and the `try...catch` block handling exception surrounds an `@sync` block synchronizing all these tasks. So that when an error is thrown in one of the `@async` tasks, nothing happens until all other tasks terminate one way or another.

I'm working on a (tentative) fix, and should be able to propose a PR in the coming hours...

ffevotte:
Closed by #488

PR #482, opened 2020-05-16
Handle errors at a finer level in `Revise.entr`

ffevotte:
This PR proposes a tentative fix for gh-481.

A `try..catch` block in introduced in a more inner position, where errors thrown by individual file-watching tasks can be handled. When one of the tasks throws an error, all other tasks are stopped (by setting `active=false`) before the error is rethrown. This allows the rethrown error to immediately propagate to the outer level, where it can be handled as usual.

(Since it wasn't completely clear to me at first glance, I should probably mention that the outer level of error handling is still necessary, because that's where interruptions will be caught.)

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -820,20 +820,24 @@ function entr(f::Function, files, modules=nothing; postpone=false, pause=0.02)
             postpone || f()
             for file in files
                 waitfor = isdir(file) ? watch_folder : watch_file
-                @async while active
-                    ret = waitfor(file, 1)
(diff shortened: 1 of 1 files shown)


ffevotte:
It looks like #443 also fixes #481.

So this PR might be useless if #443 is going to be merged. Or it might be used as a stopgap while waiting for #443.

ffevotte:
Closing this in favor of #443, especially since I'm not even sure anymore that it fixes #481 in all cases.

PR #483, opened 2020-05-18
More invalidation reduction

timholy:
The two sources fixed here are lack of type information for `lnn` and julia#15276.

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -44,7 +44,7 @@ function minimal_evaluation!(methodinfo, frame)
         stmt = src.code[id]
         me = false
         if isa(stmt, Expr)
-            if stmt.head == :call
+            if stmt.head === :call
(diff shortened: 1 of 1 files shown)


timholy:
I should clarify that this PR and related ones (https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/391, https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/29) are not to reduce invalidations *caused* by these packages, but to insulate them against invalidation by *other packages*. 

In other words, the next level of the pyramid is being climbed :smile: . For packages that run in the background for a lot of people, it seems to make sense to make sure they're well-protected from getting trashed. I wonder how much of the *perceived* slowness of the debuggers might be due to invalidation? (This aside from the *genuine* slowness of JuliaInterpreter.)

PR #487, opened 2020-05-24
Prevent `add_require` from beating `watch_package` with locking

### LINKED ISSUE #431: "Revise is not watching" pops up again
Very similar old issue is fine https://github.com/timholy/Revise.jl/issues/403 but if a line `using MAT` is included the warning `Revise is not watching` appears.
```julia
module debug_plotsrevise2

using Plots
using MAT
function __init__()
	pyplot()
end

end # module
```
```
julia> using debug_plotsrevise2
[ Info: Precompiling debug_plotsrevise2 [top-level]

julia> ┌ Warning: C:\Users\jaakkor2\.julia\packages\Plots\qZHsp\src\Plots.jl\ is not an existing directory, Revise is not watching
└ @ Revise C:\Users\jruohio\.julia\packages\Revise\ZOWOa\src\Revise.jl:489
```
Julia v1.3.1, Revise v2.5.1, MAT v0.7.0

Similar problem as in https://discourse.julialang.org/t/how-to-solve-warning-while-using-custom-package-with-revise/34743 .

timholy:
Fixes #431

### DIFF
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -280,20 +280,21 @@ end
 function maybe_add_includes_to_pkgdata!(pkgdata::PkgData, file, includes)
     for (mod, inc) in includes
         inc = joinpath(splitdir(file)[1], inc)
+        incrp = relpath(inc, pkgdata)
         hasfile = false
         for srcfile in srcfiles(pkgdata)
-            if srcfile == inc
+            if srcfile == incrp
                 hasfile = true
                 break
             end
         end
         if !hasfile
             # Add the file to pkgdata
-            push!(pkgdata.info.files, inc)
+            push!(pkgdata.info.files, incrp)
             fi = FileInfo(mod)
             push!(pkgdata.fileinfos, fi)
             # Parse the source of the new file
-            fullfile = joinpath(basedir(pkgdata), inc)
+            fullfile = joinpath(basedir(pkgdata), incrp)
             if isfile(fullfile)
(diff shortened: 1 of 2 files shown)


timholy:
Canceled one run to save resources for #488.

src/pkgs.jl
```diff
@@ -313,41 +314,46 @@ function add_require(sourcefile, modcaller, idmod, modname, expr)
...
+            exsnew = ExprsSigs()
+            exsnew[expr] = nothing
+            mexsnew = ModuleExprsSigs(modcaller=>exsnew)
+            # Before executing the expression we need to set the load path appropriately
+            prev = Base.source_path(nothing)
+            tls = task_local_storage()
+            tls[:SOURCE_PATH] = sourcefile
+            # Now execute the expression
+            mexsnew, includes = try
+                eval_new!(mexsnew, fi.modexsigs)
```
vchuravy, line 343: The lock is fairly coarse. What happens here if this includes some top-level code (unsure if that can happen) and by some unfortunate series of events you need to aquire the lock again (but maybe on a different thread?)  That would have the potential for a deadlock.

timholy: You're right that it encompasses more code than necessary. Thanks. I think there might be another way to do it, let me play with an alternative.

timholy: Might be better in #491? Would appreciate your insights if you have time.

PR #488, opened 2020-05-24
Support user-specified callbacks (v2)

ffevotte:
This PR extends and replaces #443, which it puts into a (hopefully) mergeable state:
- commits are squashed to simplify history, while preserving authorship
- it is merged with current master (as of 953b638)

CC: @timholy @tkluck


tkluck:
This looks very good to me. Thanks for taking this home @ffevotte !

timholy:
Thanks so much for doing this! Really excited to see this well fleshed-out.

I notice that all the OSX tests passed, something that's been a little rare lately. I wonder if this may have fixed the annoying flakiness on OSX. I'll test in an additional PR.

PR #489, opened 2020-05-24
Try running tests with GitHub actions

timholy:
CI is very slow for Revise, because we test a lot of versions (for good reason). Let's see if it's faster with GitHub actions.


timholy:
Disappointing, I was hoping this would be faster given that GitHub allows up to 5 macOS workers, but it really wasn't any better.

timholy:
I left some debugging code, but given how long it takes to run the tests I'll fix this in the next PR.

PR #490, opened 2020-05-25
Fix tests when watching files, run entr tests on OSX

timholy:
Thanks to #488 the `entr` tests may finally be reliable on OSX. Let's run them and see what happens.

This also adds CI testing for `Revise.watching_files[] == true`, and of course fixes a couple of bugs (one that was lurking for a while, another that was a missing check in #488). CC @ffevotte, @tkluck.

Because OSX workers are in short supply and also quite slow, this also tests fewer Julia versions and skips the `Revise.watching_files[] == true` tests on OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,24 +16,20 @@ notifications:
 git:
   depth: 99999999
 
-## uncomment/modify the following lines to allow failures
-## (tests will run but not make your overall status red)
-# matrix:
-#   allow_failures:
-#     - os: osx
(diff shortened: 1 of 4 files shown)


timholy:
Yay, the single OSX failure was not in the `entr` tests, and the single Windows failure was a download issue. This should make us considerably more reliable than we were before (it's unusual to have any not related to `entr`, let alone 2, but something to keep an out for).

PR #492, opened 2020-05-25
Allow `entr` to watch directories (fixes #470)

### LINKED ISSUE #470: Bug in `entr` call to `watch_folder`

timholy:
Fixes #470, CC @ffevotte.

Also updates the devdocs for all the latest changes.

### DIFF
--- a/docs/src/dev_reference.md
+++ b/docs/src/dev_reference.md
@@ -26,8 +26,19 @@ Revise.basebuilddir
 Revise.pkgdatas
 Revise.watched_files
 Revise.revision_queue
+Revise.NOPACKAGE
 Revise.queue_errors
 Revise.included_files
+```
+
+The following are specific to user callbacks (see [`Revise.add_callback`](@ref)) and
+the implementation of [`entr`](@ref):
+
+```@docs
+Revise.revision_event
+Revise.user_callbacks_queue
+Revise.user_callbacks_by_file
+Revise.user_callbacks_by_key
 ```
 
 ## Types
@@ -75,6 +86,14 @@ Revise.revise_dir_queued
 Revise.revise_file_queued
 ```
 
+The following functions support user callbacks, and are used in the implementation of `entr`
+but can be used more broadly:
+
+```@docs
+Revise.add_callback
+Revise.remove_callback
+```
+
 ### Evaluating changes (revising) and computing diffs
 
 [`revise`](@ref) is the primary entry point for implementing changes. Additionally,
@@ -96,10 +115,29 @@ Revise.parse_source
 Revise.parse_source!
 ```
 
+### Lowered source code
+
+Much of the "brains" of Revise comes from doing analysis on lowered code.
+This part of the package is not as well documented.
+
+```@docs
+Revise.hastrackedexpr
+```
+
 ### Modules and paths
 
 ```@docs
 Revise.modulefiles
+```
+
+### Handling errors
+
+In current releases of Julia, hitting Ctrl-C from the REPL can stop tasks running in the background.
+This risks stopping Revise's ability to watch for changes in files and directories.
+Revise has a work-around for this problem.
+
+```@docs
+Revise.throwto_repl
 ```
 
 ### Git integration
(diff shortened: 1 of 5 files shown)


timholy:
Hmm, trickier than expected. Putting this on hold.

ffevotte:
Hopefully this works, at least when `Revise.watching_files[] == false`:
- branch https://github.com/ffevotte/Revise.jl/tree/teh/fix_470
- commit b60939453561fb3747cd6a72896b0e3e72b47afa ([travis results](https://travis-ci.com/github/ffevotte/Revise.jl/builds/168195043))

I can think of at least 2 ways of handling the `Revise.watching_files[] == true` case:
1. it seems that this option is used on platforms where it is not possible to watch directories, so that it could be argued that we don't have to support watching directories in `entr` on such platforms. If we go this way, I think we should simply mention this in the documentation and deactivate these tests when `Revise.watching_files[] == true`.
2. preliminary tests seem to indicate that it is actually possible to make these tests work also in cases where `Revise.watching_files[] == true`, provided that we consider directories to be regular files. At the moment, [this condition](https://github.com/timholy/Revise.jl/blob/da6dc22590cd6ff6bcf5bbf6a5b3eb7a1dec2474/src/Revise.jl#L662) makes `Revise` forget about watched directories as soon as they are added, so that subsequent events have no chance being caught.

If we follow way (2), I'm not sure whether tests will be representative of what would happen on a real BSD system. In other words, I'm not sure whether we can claim that this feature is supported on all systems, because tests might very well pass only thanks to a Linux-specific behavior regarding directories-considered-as-regular-files.

ffevotte:
Here is an example of preliminary exploration of way (2) above:
- commit 26904430bf45c4b534e543f0bba26cdf2656905b ([travis results](https://travis-ci.com/github/ffevotte/Revise.jl/builds/168200517))

@timholy would you have an opinion on the best way to move forward?

timholy:
I'd say we should get `entr` fixes in ASAP, so perhaps split that commit out into a separate PR? We can release a 2.7.1 release once those are in.

For the more ambitous part,

> it could be argued that we don't have to support watching directories in entr on such platforms

I strongly agree with this viewpoint, though not opposed to making it work if it's pretty easy.

When I started to think this was more complicated I was starting to get concerned that the emptiness of [`latestfiles`](https://github.com/timholy/Revise.jl/blob/da6dc22590cd6ff6bcf5bbf6a5b3eb7a1dec2474/src/Revise.jl#L623) (if there are literally no files that we're individually watching in the directory) will be a problem. Not sure, I might have been going down a rabbit hole.

ffevotte:
> I'd say we should get entr fixes in ASAP, so perhaps split that commit out into a separate PR? We can release a 2.7.1 release once those are in.

Hopefully tests pass for #494, and that part will be done.

> I strongly agree with this viewpoint, though not opposed to making it work if it's pretty easy.

OK. I'll try to propose a PR that goes this way. Without having any easy way to test on BSD platforms, this seems to be the safest option.

> When I started to think this was more complicated I was starting to get concerned that the emptiness of latestfiles: if there are no individual files that we're individually watching in the directory, notify will never be called. Not sure, I might have been going down a rabbit hole.

I'll try and keep this (valid) concern in mind, but according to my first experiments around watching directories, this was not an issue when `watching_files[] == false`. I'll make sure to test for this case specifically.

timholy:
Sorry again to be slow to respond, I've been pretty consumed by #497. 

> Without having any easy way to test on BSD platforms, this seems to be the safest option.

As of be0b61ff1a4617d11b4a74c20724d58f53986d5e, CI runs its tests on Julia>=1.3, linux, once with file-watching `false` and once with `true`. The ability to do `Pkg.test(; test_args=["REVISE_TESTS_WATCH_FILES"])` was implemented in 94b029bf76a35438b9b8939442546305e144e498. So we should be sensitive to any bugs. If you're running Julia >=1.3, you can try that locally too.

ffevotte:
I think this can be safely closed, since everything mentioned here was either incorporated in #493 or in #500.

PR #496, opened 2020-05-27
Fix #494 & improve `entr` tests

### LINKED ISSUE #494: Two small bugs in `entr`
While trying to fix the tests for #470, I uncovered what I think are two small bugs in the new implementation of `entr`:

1. callbacks are not removed in all cases after `entr` ended, especially in cases where the user-provided callback throws an error
2. I *think* clusters of modifications tend to trigger user callback multiple times

A fix is provided in f6461b5cfe23062ca5214b0e00e2ebb99e178d32. Hopefully, it should land in a branch that provides a good and reliable test for `entr`. I'm only opening this issue to avoid forgetting to include this elsewhere in case I would not manage to fix tests for `entr`...

ffevotte:
Tests are modified so as to check that `entr`:
1. behaves well in cases of clustered events -> this is OK in the current master; it looks like incorrectly reported this as being a problem. Or perhaps it is only an issue when working with directories... At least we now have a way to check for this.
2. correctly removes user callbacks even when they throw -> this was indeed an issue ; a small fix is introduced in this PR.

Closes #494
(also related to #492)

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -973,8 +973,9 @@ function entr(f::Function, files, modules=nothing; postpone=false, pause=0.02)
         end
     catch err
         isa(err, InterruptException) || rethrow(err)
+    finally
+        remove_callback(key)
     end
-    remove_callback(key)
     nothing
 end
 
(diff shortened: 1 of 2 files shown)


timholy:
Thanks very much, and sorry for the delay!

PR #497, opened 2020-05-28
RFC: Revise 3.0, rewrite based on LoweredCodeUtils.CodeEdges

### LINKED ISSUE #249: Dependent blocks

### LINKED ISSUE #271: Sometimes error when Revising debug messages
I get this sporadically (when revising JuliaInterpreter I think):

```
┌ Warning: error evaluating in module JuliaInterpreter: #= C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\JuliaInterpreter.jl:26 =# @info "Generating builtins for this julia version..."
│   exception =
│    UndefVarError: #41#level not defined
│    Stacktrace:
│     [1] #collect_args#33(::Bool, ::Function, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:64
│     [2] collect_args at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:139 [inlined]
│     [3] lookup_callexpr at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\lowered.jl:13 [inlined]
│     [4] #methods_by_execution!#12(::Bool, ::Bool, ::Function, ::Any, ::Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::Frame) at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\lowered.jl:165
│     [5] (::getfield(Revise, Symbol("#kw##methods_by_execution!")))(::NamedTuple{(:define,),Tuple{Bool}}, ::typeof(Revise.methods_by_execution!), ::Function, ::Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::Frame) at .\none:0
│     [6] #methods_by_execution!#9(::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:define,),Tuple{Bool}}}, ::Function, ::Any, ::Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::Module, ::Expr) at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\lowered.jl:46
│     [7] #methods_by_execution! at .\none:0 [inlined]
│     [8] #eval_with_signatures#57(::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Module, ::Expr) at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:344
│     [9] #eval_with_signatures at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:0 [inlined]
│     [10] #instantiate_sigs!#58(::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::OrderedCollections.OrderedDict{Module,OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}}}) at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:352
│     [11] instantiate_sigs! at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:349 [inlined]
│     [12] maybe_parse_from_cache!(::Revise.PkgData, ::String) at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\pkgs.jl:224
│     [13] handle_deletions at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:468 [inlined]
│     [14] revise() at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:519
│     [15] revise at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:544 [inlined]
│     [16] run_backend(::REPL.REPLBackend) at C:\Users\Kristoffer\.julia\packages\Revise\17Klc\src\Revise.jl:846
│     [17] (::getfield(Revise, Symbol("##70#72")){REPL.REPLBackend})() at .\task.jl:259
```

### LINKED ISSUE #429: Type parameter causes UndefVarError
https://discourse.julialang.org/t/how-to-solve-error-while-compiling-custom-package-that-uses-turing-with-revise/34745/6?u=tim.holy

### LINKED ISSUE #438: suppress "revise is not tracking" warnings after first time
Sometimes when the code changes (e.g. when I update packages) there will be some problem that Revise doesn't know how to handle, so I get warnings of the form "Due to a previously reported error, the running code does not match saved version for the following files ..." with a list of files. Sometimes the files it's warning be about are for features I'm not using, so it would be great to have a way to disable those warnings after the first time they're displayed. As it is now they display every time I run code at the REPL.

### LINKED ISSUE #456: Slowdown in package load times
I'm seeing a strong slowdown in package load times that seems to be related to Revise being loaded.
If I do a single `@time using LsqFit`, the result of ~6 seconds actually prints in ~14 seconds. 
I'm illustrating this below using a `@time @time` approach, which might be flawed(?)

With Revise loaded (autoloaded)
```
julia> @time @time using LsqFit
  5.924649 seconds (14.27 M allocations: 722.932 MiB, 5.58% gc time)
 12.519490 seconds (29.66 M allocations: 1.445 GiB, 5.27% gc time)
```

Without Revise loaded
```
julia> @time @time using LsqFit
  4.529166 seconds (8.39 M allocations: 449.575 MiB, 3.31% gc time)
  4.550707 seconds (8.43 M allocations: 451.830 MiB, 3.30% gc time)
```

In this case, `LsqFit` isn't a dev-ed package, so I'm wondering whether a fix to this would be to skip tracking non-`dev`ed packages?

Julia 1.4.0
Revise v2.6.1

cc. @MasonProtter because we were talking about it on slack

### LINKED ISSUE #475: includet removes equal code in same module scope
This is kind of a strange error. I made an MWE:

File `revise_incl.jl`
```
a = 0.8
a = 0.7
a = 0.8
```

In the REPL:
```
julia> using Revise
julia> includet("revise_incl.jl")
julia> a
0.7 # I'm expecting 0.8 here
```

Change content of `revise_incl.jl` to
```
a = 0.81
a = 0.7
a = 0.8
```

In the REPL:
```
julia> using Revise
julia> includet("revise_incl.jl")
julia> a
0.8 # now it gives the expected result
```
It seems, that if a variable has the same value assigned, that second assignment is ignored.

### LINKED ISSUE #479: AstroBase.Bodies is a pathological case again
As the title states, I successfully managed to completely destroy Revise's performance when revising `AstroBase.Bodies` again 😅 . Most of the work is done in this [macro](https://github.com/JuliaAstro/AstroBase.jl/blob/master/src/Bodies/graph.jl#L32) after refactoring.
This is not an immediate problem because I consider this module "done" for now but I guess it might be an interesting benchmark.

### Steps to reproduce

1. `using Revise, AstroBase`
2. Make a change in e.g. [`planets.jl`](https://github.com/JuliaAstro/AstroBase.jl/blob/master/src/Bodies/planets.jl)
3. Wait a looong time... (EDIT: Cancelled after an hour)

### LINKED ISSUE #485: Startup time excessive
I've been monitoring with `@time using Revise` for startup:

```
$ julia
 10.132646 seconds (2.92 M allocations: 153.141 MiB, 0.36% gc time)

$ julia
  4.111674 seconds (2.35 M allocations: 125.874 MiB, 0.42% gc time)

$ julia
  4.120120 seconds (2.35 M allocations: 125.876 MiB, 0.38% gc time)
```

Do you know a reason for more allocations (and time)? I was messing with other packages, but not Revise lately. It's:

Version 1.5.0-DEV.685 (2020-04-25)
 _/ |\__'_|_|_|\__'_|  |  Commit ea669c3d3e (24 days old master)

and:

  [295af30f] Revise v2.6.6  (not 2.6.7, as not yet released, would it have mattered?).

It's worse, for 10-days old master:
```
$ time ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia
  6.312775 seconds (1.31 M allocations: 78.052 MiB, 0.18% gc time)


$ time ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia -O0
  3.175503 seconds (1.31 M allocations: 78.050 MiB, 0.36% gc time)

$ time julia-1.5 -O0
  2.330963 seconds (2.35 M allocations: 125.874 MiB, 0.68% gc time)
```

### LINKED ISSUE #508: convert an object of type `Tuple{LineNumberNode,Expr}`
I get this error in Julia 1.6.0-DEV, when I evaluate a function of a module, or I add some new deps to a module.
```julia
    end
end)MethodError: Cannot `convert` an object of type Tuple{LineNumberNode,Expr} to an object of type Array{Tuple{LineNumberNode,Expr},1}
Closest candidates are:
  convert(::Type{T}, ::AbstractArray) where T<:Array at array.jl:556
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\LinearAlgebra\src\factorization.jl:58
  ...
Stacktrace:
 [1] convert(::Type{Union{Missing, Array{Tuple{LineNumberNode,Expr},1}}}, ::Tuple{LineNumberNode,Expr}) at .\missing.jl:69
 [2] setindex!(::IdDict{Type,Union{Missing, Array{Tuple{LineNumberNode,Expr},1}}}, 
::Any, ::Any) at .\iddict.jl:72
 [3] add_signature!(::Revise.CodeTrackingMethodInfo, ::Any, ::LineNumberNode) at C:\Users\yahyaaba\.julia\packages\Revise\tV8FE\src\Revise.jl:489)Revise.LogRecord(Error, evaluation error starting at C:\Users\yahyaaba\.julia\packages\FilePathsBase\SZiBu\src\test.jl:609, lowered, Revise_e943ed8b, "C:\Users\yahyaaba\.julia\packages\Revise\tV8FE\src\lowered.jl", 106, (mod=FilePathsBase.TestPaths, ex=begin

```

### LINKED ISSUE #516: Revise causes lots of "Method definition ... overwritten on the same line." warnings on nightly
With revise loaded, on nightly julia I get lots (~100) of these printed. It doesn't happen without Revise

```
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.0-DEV.425 (2020-07-10)
 _/ |\__'_|_|_|\__'_|  |  Commit 8320fcc7d9 (1 day old master)
|__/                   |

julia> using Flux
WARNING: Method definition (::getfield(ChainRules, Symbol("#370#airyai_pullback#513")){x})(Any) in module ChainRules at /Users/ian/.julia/packages/ChainRulesCore/j0yny/src/rule_definition_tools.jl:194 overwritten on the same line.
WARNING: Method definition (::getfield(ChainRules, Symbol("#374#airyaiprime_pullback#514")){x})(Any) in module ChainRules at /Users/ian/.julia/packages/ChainRulesCore/j0yny/src/rule_definition_tools.jl:194 overwritten on the same line.
WARNING: Method definition (::getfield(ChainRules, Symbol("#378#airybi_pullback#515")){x})(Any) in module ChainRules at /Users/ian/.julia/packages/ChainRulesCore/j0yny/src/rule_definition_tools.jl:194 overwritten on the same line.
WARNING: Method definition (::getfield(ChainRules, Symbol("#382#airybiprime_pullback#516")){x})(Any) in module ChainRules at /Users/ian/.julia/packages/ChainRulesCore/j0yny/src/rule_definition_tools.jl:194 overwritten on the same line.
WARNING: Method definition (::getfield(ChainRules, Symbol("#386#besselj0_pullback#517")){x})(Any) in module ChainRules at /Users/ian/.julia/packages/ChainRulesCore/j0yny/src/rule_definition_tools.jl:194 overwritten on the same line.
WARNING: Method definition (::getfield(ChainRules, Symbol("#390#besselj1_pullback#518")){x})(Any) in module ChainRules at /Users/ian/.julia/packages/ChainRulesCore/j0yny/src/rule_definition_tools.jl:194 overwritten on the same line.
```

### LINKED ISSUE #519: Revise exits Julia
Hello,

I've just updated Julia to 1.5 and when I do `using Revise`, Julia exits for some reason. There is no error message at all. I'm using Juno.

versioninfo():

Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = "C:\Users\Administrator\AppData\Local\atom\app-1.49.0\atom.exe"  -a
  JULIA_NUM_THREADS = 12

### LINKED ISSUE #520: Strange error with PGFPlotsX and StatsPlots
I can't find a MWE, but do you know what might be causing this error? It doesn't seem to have any negative effects, as my code runs fine after.

```
loadingRevise
.1LogRecord
(Error, evaluation error starting at /home/peterj/.julia/packages/PGFPlotsX/rS8LC/src/requires.jl:60, lowered, Revise_e943ed8b, "/home/peterj/.julia/packages/Revise/BqeJF/src/lowered.jl", 106, (mod=PGFPlotsX, ex=begin
    begin
        #= /home/peterj/.julia/packages/PGFPlotsX/rS8LC/src/requires.jl:60 =#
        #= /home/peterj/.julia/packages/PGFPlotsX/rS8LC/src/requires.jl:60 =# Core.@doc "    $(SIGNATURES)\n\nConstruct table data from a `DataFrame`.\n" PGFPlotsX.TableData(df::DataFrames.DataFrame; rowsep = ROWSEP) = begin
                    #= /home/peterj/.julia/packages/PGFPlotsX/rS8LC/src/requires.jl:65 =#
                    TableData(hcat(DataFrames.eachcol(df, false)...), string.(names(df)), 0, rowsep)
                end
    end
    __pkguuid__ = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
end)UndefVarError: Base_Docs_4352b6d8 not defined
Stacktrace:
 [1] collect_args(::JuliaInterpreter.Frame, ::Expr; isfc::Bool) at /home/peterj/.julia/packages/JuliaInterpreter/RmxVj/src/interpret.jl:60)
```
versioninfo:
```
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = "/opt/visual-studio-code/code"
  JULIA_NUM_THREADS = 8
```

Thanks

timholy:
This is a major rewrite of the backedge component. Moving it to LoweredCodeUtils and giving it
a "real" API led to many improvements in selectivity of code-evaluation.

This rewrite aims to make Revise do just one thing, which is to update methods. "Data" are not touched. In this world, #249 is not a bug. The benefit is that `includet` should work much more
reliably to distinguish "code" from "method definitions"; by and large, the days of needing to separate code from method definitions in order to avoid confusing Revise should be over. However, a key question is whether people will find it confusing or annoying that changing `a = 0.2` to `a = 0.3` in package source code doesn't do anything anymore. If so, perhaps certain exemptions will need to be made (e.g., for literal constants). I am hoping to leave this open for a beta-period to see how i and others like it. But this change in behavior is the reason for considering this a 3.0 release rather than a 2.8 release.

Importantly, this introduces multiple "modes" of operation. `:sigs` is for signature extraction, `:eval` is  for full evaluation (as you might do during the initial `includet` on a file), and `:evalmeth` is used  during revision. This is essentially a generalization of `define::Bool` to allow more nuanced behavior.

Fixes #479
Closes #249
Fixes #520
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/423
Fixes #519
Fixes #516
Fixes #508
Fixes #485
Fixes #475
Fixes #438
Closes #429
Closes #271
Fixes #456


timholy:
CC @helgee, this should fix your AstroBodies issue. Please do give this a spin (`add Revise#revise3`) and tell me what you think!

helgee:
That does the trick 🎉  Thanks for providing and maintaining this essential piece of infrastructure.

I am not sure I understand what the trade-off is. This change would drop support for revising general top-level code and only support method (re-)definition and struct definition and would stop excessive (and currently slow) reinterpretation of tons of code? Is that it?

The simplicity of this approach appeals to me but there is always somebody who gets angry if a feature/bug is taken away: 
grafik

timholy:
That's why this will be Revise 3 rather than Revise 2.8; it's definitely a breaking change. Both AstroBodies and issue #249 demonstrate how fraught it is to promise to revise *everything*.

helgee:
I will take limited yet reliable over unlimited and unreliable every day 👍

tk3369:
> However, a key question is whether people will find it confusing or annoying that changing `a = 0.2` to `a = 0.3` in package source code doesn't do anything anymore. If so, perhaps certain exemptions will need to be made (e.g., for literal constants). I am hoping to leave this open for a beta-period to see how i and others like it. But this change in behavior is the reason for considering this a 3.0 release rather than a 2.8 release.

I think I would be annoyed by this.  Any changes in the source are deemed revisions to me. What would I do to get back the old behavior?

KristofferC:
FWIW, I quite often rely on the code re-evaluation . For example adding an entry to an enum type.

timholy:
> Any changes in the source are deemed revisions to me...What would I do to get back the old behavior?

Use Revise 2.x? More seriously, I am worried about this. The question is, how do you get to something that always works (Revise 2.x doesn't) but also gives you the tools you need? #249 illustrates some of the problems that crop up if Revise tries to "do work" and not just the more narrow task of revising methods. AstroBase, which partly prompted all these difficult thoughts, mixes a lot of initialization of internal data together with method definitions, and initialization uses the methods that get defined. If we change a single method, do we need to clear *all* of AstroBase's internal data stores and regenerate them from scratch? What if that takes a long time? What if it would call a C library that segfaults Julia when called a second time? What if redoing the "work" in AstroBase would delete files on the user's hard drive?

> FWIW, I quite often rely on the code re-evaluation. For example adding an entry to an enum type.

Interesting example. If we made an exemption for assignment statements (not just literal ones), adding an enum entry should work (barring bugs in the CodeEdges framework, which seem likely in the early days but we'll stomp them out eventually).

Assignment statements do seem like a category that we might need to support, and indeed I've anticipated that this might be necessary; check this PR (specifically lowered.jl, docstring for `methods_by_execution!`) for the `:evalassign` mode.

But what about a package that contains this as part of its source code?
```julia
mydata = Any[1]
push!(mydata, time())
```
Suppose you edit the first line to `mydata = Any[2]` and Revise performs the assignment. `mydata` is now a vector of length 1 rather than a vector of length 2, and running methods that use it will likely throw an error. Or should Revise also do the `push!`, even though you didn't edit that line? Where does all this stop?

timholy:
Understand I'm not complaining about your comments; these are precisely the kinds of questions we need to be considering. I'm just trying to make sure you see it from my end; as I try to make Revise more and more reliable, I get more and more worried about it performing certain kinds of operations. Striking the right balance is something we may all need to think through carefully, and I am grateful for the discussion.

tk3369:
I understand the concerns.  

The assignments that I have seen are pretty much global constants, regardless of using the `const` keyword or not.  As related to this particular example,

```julia
mydata = Any[1]
push!(mydata, time())
```

Does that even survive precompilation with consistent behavior?   Even though it may work, it seems to be an anti-pattern.  The mutation code should really go inside the `__init__` function, right?   So the next step is to make sure that `__init__` is re-run when any global is revised.

I feel the same empathy about #249.  I have seen that kind of usage with `@static` macro so that the evaluated code is different by Julia version, architecture, or some other external factor.  In that case, it tends to be less annoying because I would probably restart the REPL anyways.  On the other hand, if the switch is based on some real business logic then perhaps the program should be designed properly and do the switch within regular functions.  

Please correct me if I'm wrong.

timholy:
> Does that even survive precompilation with consistent behavior? Even though it may work, it seems to be an anti-pattern. The mutation code should really go inside the `__init__` function, right? 

With a few important exceptions (e.g., a Dict using types-as-keys which will require rehasing, https://docs.julialang.org/en/latest/manual/modules/#Module-initialization-and-precompilation-1), it does survive precompilation. I have [argued before](https://github.com/JuliaAstro/AstroBase.jl/pull/31) that such code belongs in an `__init__`, but doing that throughout the AstroBase code in detail one can't help but think that putting it in the `__init__` is uglier. Nevertheless, building the data structures from scratch in *some* function is the only way to consistently regenerate them after you've made changes to a method that they depend on.

> So the next step is to make sure that `__init__` is re-run when any global is revised.

That's an insanely demanding goal: how the heck do you determine all the dependencies of all bits of data on your system? There's no tracking done in who sets the value of an `Int`, and random number generators, etc, will make a hash of this. But believe it or not, in a few cases this is conceivable, but it also opens you up to all the dark side of #249.

tk3369:
> So the next step is to make sure that __init__ is re-run when any global is revised.
>> That's an insanely demanding goal: how the heck do you determine all the dependencies of all bits of data on your system? There's no tracking done in who sets the value of an Int, and random number generators, etc, will make a hash of this. But believe it or not, in a few cases this is conceivable, but it also opens you up to all the dark side of #249.

Sorry, I don't quite follow about this part.  The idea is to invoke `__init__` on behalf of the user when a global variable/const in the module is changed.  I didn't mean all globals... just globals inside the module.  I don't know the specifics of the dark side of #249 either.

On the tangent, I just realized a potential downside of this approach.  There is no guarantee that the `__init__` method is idempotent and so invoking it many times may introduce memory leak and all sorts of other issues.  

OK, let's say we go with your suggestion that data changes are going to be ignored in 3.0.  In that case:
1. Is the user going to be warned, or just need learn the new behavior?
2. What's the next thing the user should do in order to incorporate the change?  Restart REPL?  Is there an easier way e.g. destroying and reloading the module being monitored?

timholy:
Oh, I see, not a specific set of recomputations for everything, just run `__init__` whenever a non-method global in a package gets modified. Yes, that's much more doable. But...

> On the tangent, I just realized a potential downside of this approach. There is no guarantee that the `__init__` method is idempotent and so invoking it many times may introduce memory leak and all sorts of other issues.

That is exactly the dark side of #249 (sorry I was indirect).

> Is the user going to be warned, or just need learn the new behavior?

Certainly there would be `NEWS` in the package and an announcement on discourse. I am not sure how to notify people who don't check either of those.

> What's the next thing the user should do in order to incorporate the change?

That's a great question. Perhaps Revise could aggregate the changes and present them in some way that makes them easy to eval? Since it's REPL-based, I'd be tempted even to use [TerminalMenus](https://docs.julialang.org/en/latest/stdlib/REPL/#TerminalMenus-1) and let users mark which entries they want to evaluate. IJulia would require special consideration, though.

I will be curious how often this is really a problem. I haven't yet missed it in the day's worth of editing I've done. But I could easily see it being a big deal.

timholy:
Out of curiosity, anyone tried this yet in practice? I just pushed a bunch more updates, that fix some longstanding issues. @KristofferC, I've verified that `@enum` redefinition works (with some fixes in ba46c59) *if* you set
```julia
using Revise
Revise.revise_mode = :evalassign
```
This causes Revise to reevaluate methods and assignment statements (and nothing more). So it sticks a toe in the water of modifying data, but it's still pretty carefully controlled. Revise's default mode is `:evalmeth`, which only reevaluates methods.

I'd be curious about people's experiences. I'm really liking the new version so far.

KristofferC:
> Out of curiosity, anyone tried this yet in practice?

Made it my Revise installation now so will try it out from now on.

pfitzseb:
Sorry for sidetracking this, but
>  Perhaps Revise could aggregate the changes and present them in some way that makes them easy to eval? Since it's REPL-based, I'd be tempted even to use TerminalMenus and let users mark which entries they want to evaluate. IJulia would require special consideration, though.

sounds super interesting to me even from a GUI persepective.

We could for example have Revise provide info about method deletions/additions/re-definitions etc. pp. and show that like a git diff. Throw in a button/keybinding to accept the change and you have a hybrid Juno/Revise code update model that has the benfits of both: You have to be explicit about what code to update, but still see everything that changed and can e.g. delete methods (I think that's the one thing Juno can't easily provide).

timholy:
The hardest part about the UI part of this is that Revise will decide on a statement-by-statement level. See https://juliadebug.github.io/LoweredCodeUtils.jl/stable/edges/ and look at the second `print_with_code` example; Revise could execute just the statements marked `t`.

I've had another thought: what about supporting annotations as comments at the top of files? Something like
```
# REVISE: MODE EVALALL
```
would allow the developer to say, "there's no dangerous code in this file/package, just evaluate all changes in top-level expressions."

Currently [runtests.jl](https://github.com/timholy/Revise.jl/blob/master/test/runtests.jl#L1) has one such (undocumented) annotation, but we could add others.

timholy:
This branch now needs https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/406

timholy:
This is now officially a beta for release.

The big change from above is that I've gone back to "evaluate everything by default" for *packages*; only `includet` uses selective evauation by default. However, you can configure the mode by defining
```julia
# Cause Revise to only evaluate method definitions; do not change "data"
const __revise_mode__ = :evalmeth
# other options are: `:eval` (eval everything), `:evalassign` (also eval assignments), `:sigs` (no eval) 
```
in *each* module you want to change the mode in. @helgee, I've checked that this works for AstroBase.

KristofferC:
I don't seem to be able to Revise e.g. CodeTracking when that is the active project. Here is what I did:

```jl
~/Documents/JuliaPkgs/CodeTracking.jl master* 1m 59s
❯ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean

~/Documents/JuliaPkgs/CodeTracking.jl master
❯ /usr/local/bin/julia --project -q
julia> using Revise, CodeTracking

julia> m = @which sin(2.0)
sin(x::T) where T<:Union{Float32, Float64} in Base.Math at special/trig.jl:29

julia> CodeTracking.definition(String, m);
[ Info: tracking Base

julia> print(read(`git diff`, String)) # updated
diff --git a/src/CodeTracking.jl b/src/CodeTracking.jl
index 57c3855..f21e102 100644
--- a/src/CodeTracking.jl
+++ b/src/CodeTracking.jl
@@ -203,6 +203,7 @@ see [`definition(Expr, method::Method)`](@ref) instead.
 See also [`code_string`](@ref).
 """
 function definition(::Type{String}, method::Method)
+    @info "hello"
     file, line = whereis(method)
     line == 0 && return nothing
     src = src_from_file_or_REPL(file)

julia> CodeTracking.definition(String, m); # but doesn't show
```

timholy:
Call `Revise.add_revise_deps()` right after loading Revise. Part of the latency reduction required that Revise not compile the code to cache itself. And it's always been flaky for Revise to do surgery on itself while running, for obvious reasons.

KristofferC:
Yep, works after that. Makes sense. I had a hunch it might have to do with it being a Revise dependency.

timholy:
I think this is ready to go. I've been avoiding a rebase so that beta-testers can just `git pull` to keep up with the latest. But barring objections, tomorrow I'll rebase this branch and merge.

tpapp:
I have been using this branch for a bit more than a week now and it seems to be working fine, no apparent issues.

fredrikekre:
:tada:

PR #501, opened 2020-06-03
Move code around for easier VS Code integration

davidanthoff:
These changes would allow us to ship Revise.jl out of the box with the VS Code extension and run it integrated with the debugger in the Julia REPL of the extension.

All this does is move all the code from Revise.jl into core.jl, except for any `using` statements that load non stdlib packages. It also moves `using` statements from a number of other locations into the central Revise.jl file, so that all loading of non-stdlib packages happens in one place.

See also https://github.com/julia-vscode/julia-vscode/pull/1283.

CC @timholy


davidanthoff:
@timholy so we don't need to merge this to continue with the integration in VS Code, but it would make things a lot easier if this branch here existed in origin repo and not just in my fork. Do you think you could just push this branch here into this origin github repo here, unmerged? We use git submodules in the VS Code extension, and changing the URL of these down the road is a pain, so it would be much easier if we could use the URL for the origin repo from the beginning.

timholy:
Branch is here: https://github.com/timholy/Revise.jl/tree/davidanthoff-vscode-integration

davidanthoff:
Perfect!

davidanthoff:
Actually, I think the branch `davidanthoff-vscode-integration` does not point to the commit from this PR here, could you double check? Thanks!

timholy:
Sorry about that! I totally blew that. Try the `vscode-integration` branch.

davidanthoff:
Perfect, now it all works :)

timholy:
Now that Revise3 is out it would be fine to renew this effort.

timholy:
Superseded by #540

PR #521, opened 2020-08-28
fix for future changes in code loading

KristofferC:
The regex based parsing will dissapear, instead use the
TOML functionality in Base. We could introduce a dependency
on the TOML stdlib but AFAIU that means we would need to
bump the compat of Revise to 1.6.

With this, most tests pass on https://github.com/JuliaLang/julia/pull/36018 and the ones that don't does not seem related to manifest tracking.

### DIFF
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -442,7 +442,11 @@ end
 
 function manifest_file(project_file)
     if project_file isa String
-        mfile = Base.project_file_manifest_path(project_file)
+        mfile = @static if isdefined(Base, :TOMLCache)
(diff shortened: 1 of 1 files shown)


timholy:
Oh, thank you so much! I was working my way towards this, but from the look of it you knew a lot more about how to implement this than I would have. I am grateful!

timholy:
I'm wondering about refactoring `Base.explicit_manifest_uuid_path`, basically giving the loop over `entries` its own API. Thoughts? That makes Revise both more and less dependent on Base.

KristofferC:
I can see that making sense. Happy to help if there are any issues :)

PR #522, opened 2020-08-31
Add TOML to list of stdlibs for Julia v1.6

jmert:
Without this change, I encountered the following while trying to test-revise part of Julia:
```
$ make test-revise-Mmap
    JULIA test/revise-Mmap
ERROR: LoadError: no Revise.track recipe for module TOML
Stacktrace:
  [1] error(::String, ::Symbol)
    @ Base ./error.jl:42
...
```
I'm not sure if having `:TOML` added to the set conditional on the Julia version is what you want or not, so I can easily switch that to just a simple insert into the list if preferred.


timholy:
I don't know what would happen if it were there with no TOML module, but this looks fine. Thanks for the contribution!

PR #527, opened 2020-09-11
Pluto support 🎈

fonsp:
Hi!

Today I worked on a [highly requested Pluto feature](https://github.com/fonsp/Pluto.jl/issues/238): Revise support! Here is the [PR at Pluto](https://github.com/fonsp/Pluto.jl/pull/416/files#diff-446048a8133a25e0a07ba1f827fa93a2R37).

(It was surprisingly simple!) For this to work, I believe that we need a small change to Revise. 

_To explain why, some background on Pluto:_

Pluto uses `Distributed` to create notebook processes. The (HTTP) server has PID 1, notebooks are worker processes. These workers run independently from each other - they don't share code, loaded modules, package environments, etc., and if one notebook calls `using Revise`, then it is only loaded on that notebook's process. 

This is different from how Revise was designed to work inside Distributed, which is to load Revise [on the main process](https://github.com/timholy/Revise.jl/blob/95354947560efbc9eea1ea7d57e9c2f4566a4f60/docs/src/limitations.md#distributed-computing-multiple-workers-and-anonymous-functions). This PR adds support for a different configuration: Revise can be loaded on (some of) the worker processes, and the revised code is only loaded on one process. This feature is hidden behind a new ENV variable: `JULIA_REVISE_ON_WORKER`. (Feel free to suggest a different way to switch this feature.)

Let me know what you think!

How should this new "feature" be documented? Is it useful outside of Pluto support?

### DIFF
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -440,11 +440,13 @@ function eval_new!(exs_sigs_new::ExprsSigs, exs_sigs_old, mod::Module)
                     if VERSION < v"1.3.0" || !isexpr(thunk, :thunk)
                         thunk = ex
                     end
-                    for p in workers()
-                        p == myid() && continue
-                        try   # don't error if `mod` isn't defined on the worker
-                            remotecall(Core.eval, p, mod, thunk)
-                        catch
+                    if myid() == 1
+                        for p in workers()
+                            p == myid() && continue
+                            try   # don't error if `mod` isn't defined on the worker
+                                remotecall(Core.eval, p, mod, thunk)
+                            catch
+                            end
                         end
                     end
                     storedeps(deps, rex, mod)
@@ -1264,7 +1266,10 @@ function init_worker(p)
 end
 
 function __init__()
-    myid() == 1 || return nothing
+    run_on_worker = get(ENV, "JULIA_REVISE_ON_WORKER", "0")
+    if !(myid() == 1 || run_on_worker == "1")
+        return nothing
+    end
     if isfile(silencefile[])
         pkgs = readlines(silencefile[])
         for pkg in pkgs
--- a/src/Revise.jl
+++ b/src/Revise.jl
@@ -1266,7 +1266,7 @@ function init_worker(p)
 end
 
 function __init__()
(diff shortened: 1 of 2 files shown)


src/Revise.jl
```diff
@@ -440,11 +440,13 @@ function eval_new!(exs_sigs_new::ExprsSigs, exs_sigs_old, mod::Module)
                     if VERSION < v"1.3.0" || !isexpr(thunk, :thunk)
                         thunk = ex
                     end
-                    for p in workers()
-                        p == myid() && continue
-                        try   # don't error if `mod` isn't defined on the worker
-                            remotecall(Core.eval, p, mod, thunk)
-                        catch
+                    if myid() == 1
```
fonsp, line 443: `myid() == 1` used to be implied by the check `myid() == 1 && return nothing` in `__init__()` - other processes would never run this code

fonsp:
(It looks like CI failed on Julia nightly for the same reason as the Revise master branch. 🌛)

timholy:
Nice, thanks for contributing this!

It makes lots of sense to have this. The only thing I really wonder about is the name `JULIA_REVISE_ON_WORKER`: to me it almost suggests that you need to specify *which* worker runs Revise, whereas I think you mean this to be effectively true/false. If that's true, is there a clearer name? Maybe `JULIA_REVISE_WORKER_ONLY`?

timholy:
Also note we'll want to port this to the `revise3` branch (#497), which from the relative lack of bug reports so far during the beta period (https://discourse.julialang.org/t/ann-revise-3-0-in-beta-faster-and-better/46287) seems likely to be merged sometime next week. I could rebase just before merging, so I don't think there's anything you need to do @fonsp unless you want to test it there.

fonsp:
Awesome, thanks Tim! It would be great to release this together with Revise 3!

About the ENV: I've renamed it like you suggested 👍

About testing: the Pluto use case is covered by the test case that I added here, so just make sure that those are also ported and I'm happy! 👍👍

timholy:
Sorry for the delay. Let's merge this, release another 2.x so that anyone trapped on older versions can use Pluto, and then I'll forward-port it to revise3.

PR #530, opened 2020-09-16
Update the docs for Revise 3

timholy:


timholy:
Hmm. When this passed tests previously, it was before JuliaIntepreter 0.8.1 was out with https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/431. Oops.

timholy:
Test failures are all https://github.com/JuliaLang/julia/pull/37586#issuecomment-693434063 from https://github.com/timholy/Revise.jl/blob/d7f801fba803e9018c31c310ae32e9b53669df18/test/runtests.jl#L160

I don't think that needs to hold this up.

PR #534, opened 2020-09-17
Add Downloads stdlibs and dependents

timholy:
Stdlibs are breeding like rabbits these days!

Required for `make test-revise-*` to work.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -192,6 +192,12 @@ end
 if VERSION >= v"1.6.0-DEV.890"   # https://github.com/JuliaLang/julia/pull/37320
     push!(stdlib_names, :Artifacts)
 end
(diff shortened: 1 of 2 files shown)


src/recipes.jl
```diff
@@ -192,6 +192,12 @@ end
 if VERSION >= v"1.6.0-DEV.890"   # https://github.com/JuliaLang/julia/pull/37320
     push!(stdlib_names, :Artifacts)
 end
+if VERSION >= v"1.6.0-DEV.938"   # https://github.com/JuliaLang/julia/pull/37340
+    push!(stdlib_names, :LibCURL_jll)
+    push!(stdlib_names, :LibCURL)
+    push!(stdlib_names, :MozillaCACerts_jll)
```
fredrikekre, line 198: I think these three are just temporary so might have to have an upper Julia version bound for these later on.

fredrikekre (approved):

PR #536, opened 2020-09-18
debug pkgdir

KristofferC:


KristofferC:
Trying locally it seems to have to do with if the module is getting precompiled when loading it.

KristofferC:
Think I have a fix, now I have to write some tests heh.

KristofferC:
Seems like the all get paths now so should be indeed fixed.

PR #540, opened 2020-09-22
Rearrange code for VS Code integration

davidanthoff:
Replaces https://github.com/timholy/Revise.jl/pull/501, and is based on Revise 3.

@timholy once we have this on `master` here, I'll try to revise the branch over in the VS Code repo.


timholy:
Thanks!

Just checking, you don't need an actual release just for this, right?

davidanthoff:
No release needed, I think this is all that we need to move forward.

timholy:
Hmm. Did you `git mv` this or do it some other way? I was expecting all the line-by-line history from Revise.jl would have migrated over to `packagedef.jl`, and was trying to figure out if it was me who added a comment or someone else. But the history of that file begins with this PR.

davidanthoff:
I just copied the code over to the new file... I always thought git would be able to track that? If I understand the storage structure of git properly, it shouldn't make any difference how one does the modification...

timholy:
I've heard it said that `git mv` can sometimes help this kind of thing work better, but honestly I don't know. No worries though, just thought I'd ask in case this comes up for me.

davidanthoff:
Ok, I got curios about this and whether my understanding of things was right and read a bit more :) So git is fundamentally snapshot based, so whether one uses `git mv` or does things manually won't make a difference, git will record exactly the same information in the history. If git shows a "rename" in the history, it just means it was able to look at two snapshots and infer that this was a rename, not because some additional info was saved. One thing I found that apparently helps with this is to record a file name as one individual commit, and then make any additional changes as a separate commit, that will make it easier for git to follow things along. I didn't do that...

There is also a `--follow` argument for the `git log` command that is supposed to follow renames in the history, maybe that would recover the info?

PR #551, opened 2020-10-05
Add a fake Requires dependency and pin it at 1.0.x

timholy:
~~Fixes~~Workaround for the breakage caused by Requires 1.1.0. See https://github.com/JuliaPackaging/Requires.jl/issues/94


timholy:
I'm guessing the failure on nightly will be fixed in #553. Since this seems necessary anyway let's merge.

PR #553, opened 2020-10-05
add Tar and ArgTools stdlibs

simeonschaub:
This is really getting out of hand... :grin:


timholy:
I suspect you may also have to exclude them from the `signatures` test (similar to other recently-added stdlibs), but it looks like I have other problems to fix first. Maybe wait for #551 and rebase?

timholy:
Does ArgRead come from one of these? I haven't followed. If so, I can merge #551 now.

timholy:
OK, I went ahead and merged it, so worth another try.

timholy:
Thanks so much!

PR #556, opened 2020-10-07
WIP: fix #552: use invoke_in_world

### LINKED ISSUE #552: Segfault due to MethodError when tracking Base
Steps to reproduce:

Apply this patch to `base/range.jl` on the latest Julia master:
```diff
diff --git a/base/range.jl b/base/range.jl
index 5f30a31a0b..e004161551 100644
--- a/base/range.jl
+++ b/base/range.jl
@@ -318,7 +318,7 @@ be 1.
 """
 struct OneTo{T<:Integer} <: AbstractUnitRange{T}
     stop::T
-    OneTo{T}(stop) where {T<:Integer} = new(max(zero(T), stop))
+    OneTo{T}(stop) where {T<:Integer} = (@_inline_meta; new(max(zero(T), stop)))
     function OneTo{T}(r::AbstractRange) where {T<:Integer}
         throwstart(r) = (@_noinline_meta; throw(ArgumentError("first element must be 1, got $(first(r))")))
         throwstep(r)  = (@_noinline_meta; throw(ArgumentError("step must be 1, got $(step(r))")))
@@ -327,7 +327,7 @@ struct OneTo{T<:Integer} <: AbstractUnitRange{T}
         return new(max(zero(T), last(r)))
     end
 end
-OneTo(stop::T) where {T<:Integer} = OneTo{T}(stop)
+OneTo(stop::T) where {T<:Integer} = (@_inline_meta; OneTo{T}(stop))
 OneTo(r::AbstractRange{T}) where {T<:Integer} = OneTo{T}(r)
 
 ## Step ranges parameterized by length
```
Run Julia and do:
```
julia> using Revise

julia> Revise.track(Base)

julia> mat = zeros(10,10);
error in running finalizer: MethodError(f=Base.OneTo{T} where T<:Integer, args=(16,), world=0x000000000000722e)
error in running finalizer: MethodError(f=Base.OneTo{T} where T<:Integer, args=(16,), world=0x000000000000722e)
error in running finalizer: MethodError(f=Base.OneTo{T} where T<:Integer, args=(16,), world=0x000000000000722e)
error in running finalizer: MethodError(f=Base.OneTo{T} where T<:Integer, args=(16,), world=0x000000000000722e)
error in running finalizer: MethodError(f=Base.OneTo{T} where T<:Integer, args=(16,), world=0x000000000000722e)
┌ Error: Failed to revise /home/simeon/Documents/Julia/julia/base/range.jl
│   exception =
│    LoadError: MethodError: no method matching Base.OneTo(::Int64)
│    Closest candidates are:
│      Base.OneTo(::AbstractRange{T}) where T<:Integer at range.jl:331
│    Stacktrace:
│     [1] map(f::Type{Base.OneTo}, t::Tuple{Int64})
│       @ Base ./tuple.jl:179
│     [2] axes
│       @ ./abstractarray.jl:75 [inlined]
│     [3] _array_for(#unused#::Type{Symbol}, itr::UnitRange{Int64}, #unused#::Base.HasShape{1})
│       @ Base ./array.jl:678
│     [4] vect(::Symbol, ::Vararg{Symbol, N} where N)
│       @ Base ./array.jl:130
│     [5] docm(source::LineNumberNode, mod::Module, meta::Any, ex::Any, define::Bool)
│       @ Base.Docs ./docs/Docs.jl:545
│     [6] @doc(__source__::LineNumberNode, __module__::Module, x::Vararg{Any, N} where N)
│       @ Core ./boot.jl:508
│     [7] lower
│       @ ./meta.jl:124 [inlined]
│     [8] methods_by_execution!(recurse::Any, methodinfo::Revise.CodeTrackingMethodInfo, docexprs::Dict{Module, Vector{Expr}}, mod::Module, ex::Expr; mode::Symbol, disablebp::Bool, always_rethrow::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
│       @ Revise ~/.julia/packages/Revise/fUuEE/src/lowered.jl:178
│    in expression starting at /home/simeon/Documents/Julia/julia/base/range.jl:312
└ @ Revise ~/.julia/packages/Revise/fUuEE/src/Revise.jl:705
┌ Warning: The running code does not match the saved version for the following files:
│ 
│   /home/simeon/Documents/Julia/julia/base/range.jl
│ 
│ If the error was due to evaluation order, it can sometimes be resolved by calling `Revise.retry()`.
│ Use Revise.errors() to report errors again. Only the first error in each file is shown.
│ Your prompt color may be yellow until the errors are resolved.
└ @ Revise ~/.julia/packages/Revise/fUuEE/src/Revise.jl:799

SYSTEM (REPL): showing an error caused an error

SYSTEM (REPL): caught exception of type MethodError while trying to handle a nested exception; giving up

julia>

###
### I press `a`
###

fatal: error thrown and no exception handler available.
MethodError(f=Base.OneTo{T} where T<:Integer, args=(32,), world=0x000000000000722e)
jl_method_error_bare at /home/simeon/Documents/Julia/julia/src/gf.c:1767
jl_method_error at /home/simeon/Documents/Julia/julia/src/gf.c:1785
jl_lookup_generic_ at /home/simeon/Documents/Julia/julia/src/gf.c:2355 [inlined]
jl_apply_generic at /home/simeon/Documents/Julia/julia/src/gf.c:2370
map at ./tuple.jl:179
axes at ./abstractarray.jl:75 [inlined]
has_offset_axes at ./abstractarray.jl:85
require_one_based_indexing at ./abstractarray.jl:89
GenericIOBuffer at ./iobuffer.jl:19
GenericIOBuffer at ./iobuffer.jl:27
#IOBuffer#348 at ./iobuffer.jl:98
Type##kw at ./iobuffer.jl:91
#IOBuffer#349 at ./iobuffer.jl:114
GenericIOBuffer at ./iobuffer.jl:112
#with_output_color#795 at ./util.jl:71
with_output_color##kw at ./util.jl:71
unknown function (ip: 0x7fb64c1734c4)
#printstyled#796 at ./util.jl:105
printstyled##kw at ./util.jl:105
unknown function (ip: 0x7fb64c1733eb)
display_error at ./client.jl:102
display_error at ./client.jl:107
jl_apply at /home/simeon/Documents/Julia/julia/src/julia.h:1687 [inlined]
do_apply at /home/simeon/Documents/Julia/julia/src/builtins.c:672
jl_f__apply_latest at /home/simeon/Documents/Julia/julia/src/builtins.c:722
#invokelatest#2 at ./essentials.jl:709 [inlined]
invokelatest at ./essentials.jl:708 [inlined]
_start at ./client.jl:487
jfptr__start_31012 at /home/simeon/Documents/Julia/julia/usr/lib/julia/sys.so (unknown line)
jl_apply at /home/simeon/Documents/Julia/julia/src/julia.h:1687 [inlined]
true_main at /home/simeon/Documents/Julia/julia/src/jlapi.c:553
repl_entrypoint at /home/simeon/Documents/Julia/julia/src/jlapi.c:672
main at /home/simeon/Documents/Julia/julia/cli/loader_exe.c:46
__libc_start_main at /usr/lib/libc.so.6 (unknown line)
_start at ./julia (unknown line)

atexit hook threw an error: MethodError(f=Base.OneTo{T} where T<:Integer, args=(1,), world=0x000000000000722e)%
```

That should work, right? My best guess is that the compiler is trying to inline a method that Revise just deleted.

simeonschaub:
Depends on https://github.com/JuliaLang/julia/pull/37902 and still needs to be made backwards-compatible.

fixes #552 (at least on future Julia versions)

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -273,7 +273,7 @@ function delete_missing!(exs_sigs_old::ExprsSigs, exs_sigs_new)
                             #try  # guard against serialization errors if the type isn't defined on the worker
                                 future = remotecall(Core.eval, p, Main, :(delete_method_by_sig($sig)))
                                 finalizer(future) do f
-                                    Base.invoke_revisefunc(worldage[], Distributed.finalize_ref, f)
+                                    invoke_revisefunc(Distributed.finalize_ref, f)
                                 end
                             #catch
                             #end
@@ -1204,7 +1204,7 @@ end
 # This uses invokelatest not for reasons of world age but to ensure that the call is made at runtime.
 # This allows `revise_first` to be compiled without compiling `revise` itself, and greatly
 # reduces the overhead of using Revise.
-revise_first(ex) = Expr(:toplevel, :(isempty($revision_queue) || Base.invoke_revisefunc($revise)), ex)
+revise_first(ex) = Expr(:toplevel, :(isempty($revision_queue) || (worldage[] = Base.get_world_counter(); invoke_revisefunc($revise))), ex)
 
 @noinline function run_backend(backend)
     while true
@@ -1355,8 +1355,8 @@ function __init__()
     id = PkgId(nothing, "@REPL")
     pkgdatas[id] = pkgdata = PkgData(id, nothing)
     # Set the lookup callbacks
-    CodeTracking.method_lookup_callback[] = get_def
-    CodeTracking.expressions_callback[] = get_expressions
+    CodeTracking.method_lookup_callback[] = x -> (worldage[] = Base.get_world_counter(); invoke_revisefunc(get_def, x))
+    CodeTracking.expressions_callback[] = x -> (worldage[] = Base.get_world_counter(); invoke_revisefunc(get_expressions, x))
 
     # Watch the manifest file for changes
     mfile = manifest_file()
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -198,13 +198,14 @@ const silence_pkgs = Set{Symbol}()
 const depsdir = joinpath(dirname(@__DIR__), "deps")
 const silencefile = Ref(joinpath(depsdir, "silence.txt"))  # Ref so that tests don't clobber
 
-"""
-    Revise.worldage
-
-The world age Revise was started in. Needed so that Revise doesn't delete methods
-from under itself.
-"""
-const worldage = Ref{Union{Nothing,UInt}}(nothing)
+#"""
+#    Revise.worldage
+#
+#The world age Revise was started in. Needed so that Revise doesn't delete methods
+#from under itself.
+#"""
+#const worldage = Ref{Union{Nothing,UInt}}(nothing)
+using CodeTracking: worldage
 
 ##
 ## The inputs are sets of expressions found in each file.
@@ -1190,12 +1191,16 @@ if VERSION < v"1.6.0-DEV.1162"
     const lower_in_reviseworld = Meta.lower
 else
     function invoke_revisefunc(f, args...; kwargs...)
+        @show worldage[]
+        Base.show_backtrace(backtrace[1:2])
         return Base.invoke_in_world(worldage[], f, args...; kwargs...)
     end
     function lower_in_reviseworld(m::Module, @nospecialize(ex))
+        @show worldage[]
+        Base.show_backtrace(backtrace[1:2])
         return ccall(:jl_expand_in_world, Any,
             (Any, Ref{Module}, Cstring, Cint, Csize_t),
-            ex, m, "none", 0, world,
+            ex, m, "none", 0, worldage[],
         )
     end
 end
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -198,6 +198,11 @@ const silence_pkgs = Set{Symbol}()
 const depsdir = joinpath(dirname(@__DIR__), "deps")
 const silencefile = Ref(joinpath(depsdir, "silence.txt"))  # Ref so that tests don't clobber
 
+"""
+    world age
+"""
+const worldage = Ref{Union{Nothing,UInt}}(nothing)
+
 ##
 ## The inputs are sets of expressions found in each file.
 ## Some of those expressions will generate methods which are identified via their signatures.
@@ -1178,7 +1183,7 @@ end
 # This uses invokelatest not for reasons of world age but to ensure that the call is made at runtime.
 # This allows `revise_first` to be compiled without compiling `revise` itself, and greatly
 # reduces the overhead of using Revise.
-revise_first(ex) = Expr(:toplevel, :(isempty($revision_queue) || Base.invokelatest($revise)), ex)
+revise_first(ex) = Expr(:toplevel, :(isempty($revision_queue) || Base.invoke_in_world($(worldage[]), $revise)), ex)
 
 @noinline function run_backend(backend)
     while true
@@ -1277,6 +1282,7 @@ function init_worker(p)
 end
 
 function __init__()
+    worldage[] = Base.get_world_counter()
     run_on_worker = get(ENV, "JULIA_REVISE_WORKER_ONLY", "0")
     if !(myid() == 1 || run_on_worker == "1")
         return nothing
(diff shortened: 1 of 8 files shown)


src/packagedef.jl
```diff
@@ -262,10 +267,13 @@ function delete_missing!(exs_sigs_old::ExprsSigs, exs_sigs_new)
...
-                            try  # guard against serialization errors if the type isn't defined on the worker
-                                remotecall(Core.eval, p, Main, :(delete_method_by_sig($sig)))
-                            catch
-                            end
+                            #try  # guard against serialization errors if the type isn't defined on the worker
+                                future = remotecall(Core.eval, p, Main, :(delete_method_by_sig($sig)))
+                                finalizer(future) do f
+                                    Base.invoke_in_world(worldage[], Distributed.finalize_ref, f)
+                                end
+                            #catch
```
simeonschaub, line 275: Do we need to catch all errors here, or can we only catch certain ones? This bit me during debugging, when I tried to throw an error in the finalizer to get a stacktrace for debugging.

timholy: Certainly it's fine to comment out while debugging, or to `catch e; isa(e, SomeKindOfError) && throw(e)` if `SomeKindOfError` deserves special handling. What kind of errors did you have in mind?

Off-topic, but since we talked about `backtrace()` in our other discussion, just checking, do you know about `catch_backtrace()`? It's sometimes quite useful.

simeonschaub: I was more wondering, what kind of errors this was initially supposed to catch, because catching all possible errors seems seems a bit dangerous to me, since we possibly ignore errors the user would like to know about. So I was more thinking of something along the lines of `e isa ErrorThatsDistributedsFault || rethrow()`, i.e. that the default would be to rethrow errors. Perhaps that's just not possible in this case though.

> Off-topic, but since we talked about backtrace() in our other discussion, just checking, do you know about catch_backtrace()?

I do now! :laughing:

simeonschaub: Or can we at least log those errors with something like `@warn`?

timholy: I see your point. This probably is the "lazy" approach, but doing better requires that we anticipate two things:

- it seems likely that many errors might have already been caught by the main process, and so a report from a worker would be redundant. However, I suspect this was more of a concern in previous versions of Revise where there was more `try/catch` stuff; now, it seems likely that it wouldn't even get here if it fails in the main process.
- workers may not all have the same code---for example, in some cases I've deliberately loaded Gtk only in the main process and had the workers be just computational engines. In such cases, some revise attempts will fail, but it would be wrong to send a warning to the user who would mentally think "I know, I did it on purpose."

As long as we can avoid pestering users incorrectly, I'm all in favor of pestering them when they deserve it :smile:.

simeonschaub: Is there a way to check whether an error was thrown during regular "revising" or something else? If not, I would propose we just keep it as it is currently, but use either `@debug` or Revise's logging facilities to log the error, so people who want to know about these kinds of errors can still get them.

timholy: The pkgdata might be in `revision_errors` but the needed info doesn't propagate down here. I think logging it makes sense.

timholy:
Doesn't it make more sense to only update the world age at the beginning of a revision? I worry that updating it during each package's deletion phase will just trigger the parent issue all over again. You might not see that in your tests, but add a *second* deletion and then the update happens before Revise gets a chance to define the replacement, then I believe you are in trouble.

simeonschaub:
The problem is that macro expansion at least needs to be aware of macros that Revise itself creates when revising code. (Perhaps there should be a separate `worldage` for that?) This is where the error from `@testmac` [here](https://travis-ci.org/github/timholy/Revise.jl/jobs/735539537#L517) comes from. I honestly don't quite know how to fix that, because as far as I understand it, for every file `Revise` revises, it first does all the relevant method deletions and only then evaluates the new definition. For macros this means, that we would need to update our worldage between those evaluations, because otherwise later method definitions using those macros wouldn't get macro expanded correctly, or error because the macro wasn't defined in the current worldage yet. If we did update our worldage between each of these definitions though, we wouldn't be able to solve #552 anymore. Do you have any ideas on how this could still work?

timholy:
Oh, interesting. Perhaps the easiest would be to change the order: handle each package (both deletion and evaluation in one package before moving to the next).

In the old days, doing the deletions first was basically a hack to allow moving a method from one file to another (within the same package) without accidentally deleting it after evaluating in the new location. But Revise is now much smarter about that: it sorts packages and files in dependency order, and it supports having multiple locations transiently. So it seems like it should be safe to switch the strategy now.

PR #558, opened 2020-10-12
fixes for changes in code loading

KristofferC:


timholy:
See companion in https://github.com/timholy/CodeTracking.jl/pull/71

PR #560, opened 2020-10-17
handle `GlobalRef` case in `matches_eval`:

aviatesk:
https://github.com/JuliaLang/julia/pull/37573 seems to have changed lowering:
```julia
julia> lwr = Meta.@lower begin
           for T in (Int, Float64, String)
               @eval mytypeof(x::$T) = $T
           end
       end
:($(Expr(:thunk, CodeInfo(
    @ none:2 within `top-level scope'
1 ─ %1  = Core.tuple(Int, Float64, String)
│         #s192 = Base.iterate(%1)
│   %3  = #s192 === nothing
│   %4  = Base.not_int(%3)
└──       goto #4 if not %4
2 ┄ %6  = #s192
│         T = Core.getfield(%6, 1)
│   %8  = Core.getfield(%6, 2)
│  └
│   @ none:3 within `top-level scope'
│   %9  = Core._expr(:(::), :x, T)
│   %10 = Core._expr(:call, :mytypeof, %9)
│   %11 = Core._expr(:block, $(QuoteNode(:(#= none:3 =#))), T)
│   %12 = Core._expr(:(=), %10, %11)
│         Core.eval(Main, %12)
│         #s192 = Base.iterate(%1, %8)
│   %15 = #s192 === nothing
│   %16 = Base.not_int(%15)
└──       goto #4 if not %16
3 ─       goto #2
4 ┄       return nothing
))))

julia> first(lwr.args).code[13]
:(Core.eval(Main, %12)) # <= `GlobalRef`
```

this will fix this test case that is broken on the current Julia master: 
https://github.com/timholy/Revise.jl/blob/397db6d63b0f1c845ae51fa879c4c2645e35072b/test/runtests.jl#L1731-L1734
maybe there're other places where changes are needed  ¯\_(ツ)_/¯

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -52,7 +52,7 @@ function matches_eval(stmt::Expr)
     f = stmt.args[1]
     return f === :eval ||
            (callee_matches(f, Base, :getproperty) && is_quotenode_egal(stmt.args[end], :eval)) ||
-           (isa(f, GlobalRef) && f.name === :eval)
+           (isa(f, GlobalRef) && f.name === :eval) || is_quotenode_egal(f, Core.eval)
 end
 
 function categorize_stmt(stmt)
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -50,7 +50,9 @@ end
 function matches_eval(stmt::Expr)
     stmt.head === :call || return false
     f = stmt.args[1]
-    return f === :eval || (callee_matches(f, Base, :getproperty) && is_quotenode_egal(stmt.args[end], :eval))
+    return f === :eval ||
+           (callee_matches(f, Base, :getproperty) && is_quotenode_egal(stmt.args[end], :eval)) ||
+           (isa(f, GlobalRef) && f.name === :eval)
 end
 
 function categorize_stmt(stmt)


timholy:
Thanks, really grateful for this!

I won't merge immediately because there are other tests broken, and I want to at least understand them first. It looks like the failures start in JuliaInterpreter, so I'll merge this once I've worked my way up the stack.

aviatesk:
yeah I think I was a bit too hasty; it would be better to understand why this change is needed first. 
e.g. we may want to have something like a transformation pass to "normalize" a lowered code in JuliaInterpreter, then it would get rid of changes in downstreams.

timholy:
You probably know this, but just to document it: on Julia 1.5, `@eval feval1(::Float32) = 1` gets lowered to access `eval` through a two-step process:
```julia
%9  = Base.getproperty(Base.Core, :eval)
...
(%9)(Main.Toplevel, %13)
```
On Julia-master (as of today), it's more direct:
```julia
Core.eval(Main.Toplevel, %12)
```
where the first argument of the `:call` expression is `GlobalRef(Core, :eval)`.

JuliaInterpreter's [`optimize!`](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/2f5f80034bc287a60fe77c4e3b5a49a087e38f8b/src/optimize.jl#L132-L240) further transforms this by looking up the GlobalRef when the framecode is constructed. That means that as far as Revise is concerned, there will also be some that are `QuoteNode(Core.eval)`. So I think if you change this to
```diff
diff --git a/src/lowered.jl b/src/lowered.jl
index 08fc877..e7dca51 100644
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -52,7 +52,7 @@ function matches_eval(stmt::Expr)
     f = stmt.args[1]
     return f === :eval ||
            (callee_matches(f, Base, :getproperty) && is_quotenode_egal(stmt.args[end], :eval)) ||
-           (isa(f, GlobalRef) && f.name === :eval)
+           (isa(f, GlobalRef) && f.name === :eval) || is_quotenode_egal(f, Core.eval)
 end
 
 function categorize_stmt(stmt)
```
then Revise seems to pass tests (at least with a couple of other fixes I've made locally).

I'm not sure why we need *both* the GlobalRef & the `quotenode_egal` check, except that there might be times where someone might want to call `matches_eval` on the result of `Meta.lower` rather than a framecode.

src/lowered.jl
```diff
@@ -50,7 +50,9 @@ end
 function matches_eval(stmt::Expr)
     stmt.head === :call || return false
     f = stmt.args[1]
-    return f === :eval || (callee_matches(f, Base, :getproperty) && is_quotenode_egal(stmt.args[end], :eval))
+    return f === :eval ||
+           (callee_matches(f, Base, :getproperty) && is_quotenode_egal(stmt.args[end], :eval)) ||
+           (isa(f, GlobalRef) && f.name === :eval)
```
timholy, line 55:
```suggestion
           (isa(f, GlobalRef) && f.name === :eval) || is_quotenode_egal(f, Core.eval)
```

aviatesk:
> I'm not sure why we need both the GlobalRef & the quotenode_egal check, except that there might be times where someone might want to call matches_eval on the result of Meta.lower rather than a framecode.

yeah, I made this patch just because I found https://github.com/JuliaLang/julia/pull/37573 broke my own code that is equivalent to `matches_eval` but works on a `lower`ed code that has not been `JuliaInterpreter.optimize!`d:
https://github.com/aviatesk/JET.jl/blob/9c12bf1138dff390bc7d46695a21ab5faab3d1e3/src/virtualprocess.jl#L350-L363

So I would like to say let's remove the `GlobalRef` part, which is really not necessary for `Revise.matches_eval`.

timholy:
I don't think it hurts anything, though?

aviatesk:
> I don't think it hurts anything, though?

yeah

PR #575, opened 2020-11-18
Exclude only Requires version 1.1.0

c42f:
I *think* this is the correct syntax for excluding only Requires-1.1.0. Testing locally seemed to work despite the lack of registration triggering at https://github.com/JuliaPackaging/Requires.jl/commit/4a23093e13208a4eec7e268802443d1712189cdd.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -21,7 +21,8 @@ CodeTracking = "1"
 JuliaInterpreter = "0.8"
 LoweredCodeUtils = "1.2"
 OrderedCollections = "1"
-Requires = "~1.0"
+# Exclude Requires-1.1.0 - see https://github.com/JuliaPackaging/Requires.jl/issues/94
+Requires = "~1.0, ^1.1.1"
 julia = "1"
 
 [extras]


timholy:
Thanks! I hadn't noticed the failure to register Requires, I just submitted https://github.com/JuliaRegistries/General/pull/24854. I'll re-run CI here just to be sure.

Project.toml
```diff
@@ -21,7 +21,8 @@ CodeTracking = "1"
 JuliaInterpreter = "0.8"
 LoweredCodeUtils = "1.2"
 OrderedCollections = "1"
-Requires = "~1.0"
+# Exclude Requires-1.1.0 - see https://github.com/JuliaPackaging/Requires.jl/issues/94
```
KristofferC, line 24: Unfortunately, I think Pkg will delete this comment next time it writes the project file.

timholy: Yes, I've noticed this. I think I understand why this is nontrivial to deal with, although establishing a convention that comments go with their next item seems like it might allow the problem to be resolved in most cases.

c42f: Yes it will, but I still thought adding a comment was an improvement over not adding one ;-) This can either be maintained manually, or if it's deleted at least it'll appear in the git history.

PR #576, opened 2020-11-18
Fix revise() of non-package files with entr / add_callback

### LINKED ISSUE #574: Individual tracked files not revised with `entr`
Yesterday I was trying to use `Revise.entr` to force live code reloading for web server development and became greatly confused because `entr` just seems to work inconsistently with the REPL-triggered Revise event loop.

Here's a MWE in which there's a file `foo.jl` containing a trivial function `foo()` which I'd like to revise. I expected that calling `includet("foo.jl")` at the top would automatically keep the function `foo()` up to date in the `entr` callback, like it does in the REPL. However

```julia
using Revise

open("foo.jl", write=true) do io
    write(io, "foo() = 0")
end

includet("foo.jl")

@sync begin
    # Simulate the user asynchronously modifying file foo 5 times
    @async for i=1:5
        sleep(1)
        open("foo.jl", write=true) do io
            write(io, "foo() = $i")
        end
    end

    entr(["foo.jl"]) do
        # I'd expect foo() to be already up to date once execution reaches this
        # point.
        @show Base.invokelatest(foo)
        # Press Ctrl-C to exit ...
    end
end
```

The output:

```text
julia> include("foo_test.jl")
Base.invokelatest(foo) = 0  # should be 1
Base.invokelatest(foo) = 0  # should be 2
Base.invokelatest(foo) = 0  # etc
Base.invokelatest(foo) = 0
Base.invokelatest(foo) = 0
Base.invokelatest(foo) = 0
^C
```

Another thing I noticed - getting back into the REPL after running the `entr()` loop, the revision of `foo()` is now broken in the REPL as well.

@ffevotte I think this change was introduced by the callback rewrite https://github.com/timholy/Revise.jl/pull/488 as pinning Revise to the version v2.6.7 before that was merged makes this example behave as I expect, but v2.7.0 does not.

@clarkevans this may affect you, as I was taking inspiration from your example code in https://github.com/JuliaWeb/HTTP.jl/issues/587 when I hit this issue.

c42f:
The problem here was that `add_callback()` calls `init_watching` with no
`pkgdata`, so the id of the files which are now watched by the callback
gets set unconditionally to NOPACKAGE. In turn, this breaks revision of
any of these files which were already being revised, because the id
check in `revise_dir_queued` fails and it no longer adds them to the
revision queue.

Instead, change init_watching to not update an existing id when pkgdata
== NOPACKAGE is supplied, unless we're not already watching the file.

This fixes #574.  I'm not sure it's the *best* way to fix it, but without providing a module to `add_callback` where the files will be `revise()`d within, it seems somewhat reasonable. I suppose the other option would be to default to `Main` like `Revise.track` does, but that might just lead to confusion as well.

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -510,14 +510,20 @@ function init_watching(pkgdata::PkgData, files=srcfiles(pkgdata))
     for file in files
         dir, basename = splitdir(file)
         dirfull = joinpath(basedir(pkgdata), dir)
-        already_watching = haskey(watched_files, dirfull)
-        already_watching || (watched_files[dirfull] = WatchList())
-        push!(watched_files[dirfull], basename=>pkgdata)
-        if watching_files[]
-            fwatcher = TaskThunk(revise_file_queued, (pkgdata, file))
-            schedule(Task(fwatcher))
-        else
-            already_watching || push!(udirs, dir)
+        already_watching_dir = haskey(watched_files, dirfull)
+        already_watching_dir || (watched_files[dirfull] = WatchList())
+        watchlist = watched_files[dirfull]
+        current_id = get(watchlist.trackedfiles, basename, nothing)
+        new_id = pkgdata.info.id
+        if new_id != NOPACKAGE || current_id === nothing
+            # Allow the package id to be updated
+            push!(watchlist, basename=>pkgdata)
+            if watching_files[]
+                fwatcher = TaskThunk(revise_file_queued, (pkgdata, file))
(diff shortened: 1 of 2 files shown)


c42f:
Aha, tests passed. When in doubt, add more `sleep(mtimedelay)`? :-/

timholy:
Thank you so much for digging into this and fixing it! These are tricky bugs to figure out.

c42f:
You're welcome Tim, thanks for making Revise! And maintaining it with such determination :-)

Issue #568, opened 2020-11-11
typemax error


cstjean:
On Revise 1.3.6 + Julia 1.5, I have some macro invocation like

```julia
@def_model Foo begin
    f(x) = 5
end
```

which macroexpands into a lot of function definitions, but nothing particularly exotic. In particular, it defines a function like `g(x) = 5`. When I change `5` to `6`, I would expect Revise to pick up the change, but it doesn't: `g(2)` still returns 5. Calling `Revise.revise(that_module)` works. No error is shown. But if I turn on the logger via `rlogger = Revise.debug_logger()`, then revising (i.e. changing the 5 to 7) now fails with an error

```julia
┌ Error: Failed to revise /Users/cedric/blah.jl
│   exception =
│    AssertionError: ld[idx] < typemax(eltype(ld))
│    Stacktrace:
│     [1] (::Revise.var"#79#81"{OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}},OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}}})() at /Users/cedric/.julia/packages/Revise/nWJXk/src/packagedef.jl:255
└ @ Revise ~/.julia/packages/Revise/nWJXk/src/packagedef.jl:693
┌ Warning: The running code does not match the saved version for the following files:
```

Does that ring a bell? If not, I'll produce an MWE, but there's a lot of trimming involved.

timholy:
It's from https://github.com/timholy/Revise.jl/blob/023b6c254272601a51d11b2517eb9238eb395ee1/src/packagedef.jl#L250-L258. That code handles multiple definitions, e.g., if you copy a method from one file to another but forget to delete the first one before making a revision. It seems to suggest line number problems, but beyond that I cannot say.

cstjean:
MWE:

```julia
module MyModule

using MacroTools: @q

function my_fun end

macro some_macro(value)
    return esc(@q $MyModule.my_fun() = $value)   # with normal `quote` it works fine
end

@some_macro 20   # changing this value isn't revised
end

> using Revise, MyModule
> MyModule.my_fun()
```

> It seems to suggest line number problems, but beyond that I cannot say.

Good guess. It's caused by `@q`, which strips away line numbers. I admit that I was never quite sure where it was appropriate to use. If you think it's just not right to strip away line numbers at the top-level, I'll repent and replace it with regular `quote`.

[...]

PR #578, opened 2020-11-19
Add test for issue 568

timholy:
We should test here that #568 stays fixed. The fix in https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/54 is very focused, which is good, but we want to make sure the overall behavior works.

CC @cstjean


cstjean (approved):
LGTM. I'm surprised you don't make a function for the repeated `open(...) do io` block where only the `@some_macro` line changes.

timholy:
> I'm surprised you don't make a function for the repeated `open(...) do io` block where only the `@some_macro` line changes.

The test framework has changed a few times, and indeed it might be time for a new round of cleanups. Happy to accept PRs!

PR #585, opened 2020-12-10
Fix adding docstring to module with none

### LINKED ISSUE #583: Revising the addition of a docstring
Ok, this is a little silly, but I was just demoing Revise to someone and blindly tried adding a docstring, expecting it to become available (akin to other changes).

Not a crucial thing, but could be nice to support.

timholy:
Workaround for julia#38819
Fixes #583


timholy:
Travis passed on one run but didn't signal back.

Issue #599, opened 2021-01-26
Functions defined in type definitions break Revise


thofma:
```julia
shell> cat test.jl
mutable struct A
  x::Int

  A(x) = f(x)

  function f(x)
    x^2
  end
end
```
This works fine with `include`:
```
julia> include("test.jl")

julia> A(2)
4
```
But Revise is choking on this:
```
julia> Revise.includet("test.jl")
ERROR: UndefVarError: f not defined
top-level scope at /home/bla/test.jl:1
in expression starting at /home/bla/test.jl:1
```
This is julia 1.5 and master branch of Revise.

timholy:
It's a LoweredCodeUtils issue (different org so I can't transfer this issue). In case I don't get to it soon, here's progress so far:
```julia
 1 f 1 ─       global A
 2 f │         const A
 3 f │   %3  = Core.svec()
 4 f │   %4  = Core.svec(:x)
 5 f │         _1 = Core._structtype(Main, :A, %3, %4, true, 1)
 6 f │         Core._setsuper!(_1, $(QuoteNode(Any)))
 7 f │   %7  = $(Expr(:isdefined, :A))
 8 f └──       goto #6 if not %7
 9 f 2 ─ %9  = A
10 f │   %10 = Core._equiv_typedef(%9, _1)
11 f └──       goto #4 if not %10
12 f 3 ─       _1 = %9
13 f └──       goto #5
14 f 4 ─       A = _1
15 f 5 ┄       goto #7
16 f 6 ─       A = _1
17 f 7 ┄ %17 = _1
18 f │   %18 = Core.svec(Int)
19 f │         Core._typebody!(%17, %18)
20 f │         _2 = ($(QuoteNode(Core.Box)))()
21 f │         global A
22 t │         $(Expr(:method, :A))
23 t │   %23 = Core.Typeof(A)
24 t │   %24 = Core.svec(%23, $(QuoteNode(Any)))
25 t │   %25 = Core.svec()
26 t │   %26 = Core.svec(%24, %25, $(QuoteNode(:(#= /tmp/revise/test.jl:4 =#))))
27 t │   %27 = $(Expr(:copyast, :($(QuoteNode(Any[Symbol("#self#"), :x])))))
[...]

Issue #557, opened 2020-10-07
exception = function type in method definition is not a type


hdavid16:
I am having an issue that is difficult to reproduce and I haven't been able to fix. When I make any change to my test.jl file. I get the error below (even when I change a comment line). I should note that line 98 just has an `end` to close a `for` loop. Also, there is no `Function` type declared in the arguments for any of the functions in this file. Any thoughts?

```julia
┌ Error: Failed to revise c:\..\test.jl
│   exception =
│    function type in method definition is not a type
│    top-level scope at c:\...\test.jl:98
│    Revise evaluation error at c:\...\test.jl:98
│ 
└ @ Revise C:\Users\HD\.julia\packages\Revise\qYvgD\src\packagedef.jl:691

timholy:
Without more information, basically no idea. There's a small chance you'll get more hints if you `dev` Revise and follow the instructions to uncomment a line in `trim_toplevel!`. (Just moments ago I pushed a commit documenting this trick, so make sure you're on the latest.)

hdavid16:
Ok. I'll just wait for a future update. I am actually getting Revise to complain about line 98 when I edit other files in the package I'm working on. The only common thing I can find between the files that error upon edit is that they use `ResumableFunctions`. Thanks for your time though...

timholy:
> I'll just wait for a future update.

You mean, you're hoping this will magically be fixed in the future? It might happen, but it will depend on someone finding a reproducer. If you can share your code but it's big, you can make a [gist](https://gist.github.com/) if necessary.

hdavid16:
Unfortunately, I can't share the code at the moment. I'll do some code maintenance on my package and get back to this in the future. You may close the issue if you wish and I'll reopen if I can make the error reproducible.

A-ali123:
https://github.com/timholy/Revise.jl/issues/557#issuecomment-705739257

hdavid16:
@timholy, I have now put together a very simple example to reproduce this error. I put it in the following repo: https://github.com/hdavid16/ReviseResumable.jl

If I run `./examples/ex1.jl`, it will run. However, once I make any change to the resumable function in `./src/resumable.jl` and then try to rerun `ex1.jl`, I will get the error:

```julia
┌ Error: Failed to revise c:\Users\HD\.julia\dev\ReviseResumable\src\resumable.jl
│   exception =
│    function type in method definition is not a type
[...]

PR #611, opened 2021-02-27
Test whether windows fail correlates with LCU1.2.8

timholy:
LCU1.2.8 fixed two issues (#599 and #557), so it seems like a good
change. This tests whether it's causing the CI failures in #608.


timholy:
Hmm, OK, LCU 1.2.8 indeed causing a new windows-specific problem.

PR #622, opened 2021-04-14
Pass changed expression to callback

rikhuijzer:
This is a draft pull request. I'm very open to discuss the changes and will definitely add some proper documentation and tests once I have some feedback on the changes.

In my package [Books.jl](https://github.com/rikhuijzer/books.jl), I am generating documents containing Julia output; similar to [Weave.jl](https://github.com/JunoLab/Weave.jl), [Franklin.jl](https://github.com/tlienart/Franklin.jl), [Publish.jl](https://github.com/MichaelHatherly/Publish.jl) or [bookdown](https://github.com/rstudio/bookdown). However, as far as I know, these packages all evaluate user code from strings. In my package, I encourage users to define their output as functions inside a module. My package, then, evaluates these functions, captures their outputs and puts it into the document. This way, my package can use all the benefits that packages have including things as tests, Revise.jl support and compat entries.

So far, this all works absolutely great and fine. For example, I can put `Revise.entr` on a method which generates a table and have an updated web page within a second. However, the user needs to manually decide what methods to run again when the code is changed or needs to run all methods again upon each Revise trigger. It would be super awesome if Revise.jl can pass me the changed methods so that I only have to re-run the changed methods.

Revise did not pass the changed expressions to the callback, so that is what I implemented in this PR. With this, I was able to obtain `methods::Vector{Method}` of the changed methods with
```
using Revise
using CodeTracking
using Foo

changed = []

function f(; changed_exs=[])
    # println("changed:")
    # print(changed_exs)
    global changed = changed_exs

    for ex in changed_exs
        @show methodinfo = Revise.CodeTrackingMethodInfo(ex)
        node = ex.args[1]
        file = string(node.file)
        line = node.line
        sigs = signatures_at(file, line)
        # This thing can contain noise, but at least does contain what I want.
        sigs = unique(sigs)
        methods = which.(sigs)
        @show methods
        return methods
    end
end

Revise.add_callback(f, [], [Foo])
```
Of course, I still need to figure out what other methods changed based on these, but the functionality proposed in this PR would already be useful for my package. And, hopefully, also for others in the future. Maybe, for example, [Pluto.jl#508](https://github.com/fonsp/Pluto.jl/issues/508).

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -109,12 +109,21 @@ function remove_callback(key)
     nothing
 end
 
-function process_user_callbacks!(keys = user_callbacks_queue; throw=false)
+function process_user_callbacks!(keys = user_callbacks_queue; throw=false, changed_exs=[])
     try
         # use (a)sync so any exceptions get nicely collected into CompositeException
         @sync for key in keys
-            f = user_callbacks_by_key[key]
-            @async Base.invokelatest(f)
+            if VERSION >= v"1.2"
+                f = user_callbacks_by_key[key]
+                if hasmethod(f, Tuple{}, (:changed_exs,))
+                    @async Base.invokelatest(f; changed_exs=changed_exs)
+                else
+                    @async Base.invokelatest(f)
+                end
+            else
+                f = user_callbacks_by_key[key]
+                @async Base.invokelatest(f)
+            end
         end
     catch err
         if throw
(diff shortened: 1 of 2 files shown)


rikhuijzer:
Bump

rikhuijzer:
Closing PR due to inactivity. I'll figure out a workaround.

PR #644, opened 2021-08-18
Type-assert the result of `predicate(stmt)`

timholy:
With the `@nospecialize` added, the result is no longer inferrable.


timholy:
Passes locally on Julia 1.0 so I'll call this a sporadic failure.

PR #650, opened 2021-09-29
add current type mutation workaround doc

xgdgsc:
Didn' t find this in official doc. Might include this for better visibility of this workaround.

### DIFF
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -32,7 +32,9 @@ function processFoo(foo::FooStruct)  # no need to change this
     @info foo.bar
 end
 ```
-Then it could be used properly as long as the new type name don' t conflict with existing type name. So you can restart from version number 1 if you restart the REPL.
+This works as long as the new type name doesn' t conflict with an existing name; within a session you need to change the name each time you change the definition.
+
+Once your development has converged on a solution, it's best to switch to the "permanent" name: in the example above, `FooStruct` is a non-constant global variable, and if used internally in a function there will be consequent performance penalties. Switching to the permanent name will force you to restart your session.
 
 In addition, some situations may require special handling:
 
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -20,7 +20,7 @@ function processFoo(foo::FooStruct)
     @info foo.bar
 end
 ```
-and update type like
+and then the type can be updated like
 ```jl
 # 2nd version
 struct FooStruct2  # change version here
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -32,7 +32,7 @@ function processFoo(foo::FooStruct)  # no need to change this
     @info foo.bar
 end
 ```
-This works as long as the new type name doesn' t conflict with an existing name; within a session you need to change the name each time you change the definition.
+This works as long as the new type name doesn't conflict with an existing name; within a session you need to change the name each time you change the definition.
 
 Once your development has converged on a solution, it's best to switch to the "permanent" name: in the example above, `FooStruct` is a non-constant global variable, and if used internally in a function there will be consequent performance penalties. Switching to the permanent name will force you to restart your session.
 
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -8,7 +8,7 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
 
 These kinds of changes require that you restart your Julia session. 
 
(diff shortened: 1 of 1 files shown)


docs/src/limitations.md
```diff
@@ -6,7 +6,33 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+struct FooStruct2  # change version here
+    bar::Int
+    str::String
+end
+FooStruct=FooStruct2   # change version here
+function processFoo(foo::FooStruct)  # no need changing this
+    @info foo.bar
+end
+```
+Then it could be used properly as long as the new type name don' t conflict with existing type name. So you can restart from version number 1 if you restart the REPL.
```
timholy, line 35:
```suggestion
This works as long as the new type name doesn' t conflict with an existing name; within a session you need to change the name each time you change the definition.

Once your development has converged on a solution, it's best to switch to the "permanent" name: in the example above, `FooStruct` is a non-constant global variable, and if used internally in a function there will be consequent performance penalties. Switching to the permanent name will force you to restart your session.
```

timholy (commented):
Great idea to document this, thanks! I mostly have English-usage suggestions, otherwise it's solid.

docs/src/limitations.md
```diff
@@ -6,7 +6,35 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
 - conflicts between variables and functions sharing the same name
 - removal of `export`s
 
-These kinds of changes require that you restart your Julia session.
+These kinds of changes require that you restart your Julia session.
```
fredrikekre, line 9:
```suggestion
These kinds of changes require that you restart your Julia session.
```

docs/src/limitations.md
```diff
@@ -6,7 +6,35 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
 - conflicts between variables and functions sharing the same name
 - removal of `export`s
 
-These kinds of changes require that you restart your Julia session.
+These kinds of changes require that you restart your Julia session. 
+
+During early stages of development, it's quite common to want to change type definitions. You can work around Julia's/Revise's limitations by temporary renaming:
+
+```jl
```
fredrikekre, line 13:
````suggestion
```julia
````

File: docs/src/limitations.md [hunk 4]
```diff
@@ -6,7 +6,35 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+struct FooStruct1
+    bar::Int
+end
+FooStruct=FooStruct1
+function processFoo(foo::FooStruct)
+    @info foo.bar
+end
+```
+and then the type can be updated like
+```jl
```
fredrikekre, line 24:
````suggestion
```julia
````

File: docs/src/limitations.md [hunk 5]
```diff
@@ -6,7 +6,35 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+These kinds of changes require that you restart your Julia session. 
+
+During early stages of development, it's quite common to want to change type definitions. You can work around Julia's/Revise's limitations by temporary renaming:
+
+```jl
+# 1st version
+struct FooStruct1
+    bar::Int
+end
+FooStruct=FooStruct1
```
fredrikekre, line 18:
```suggestion
FooStruct = FooStruct1
```

docs/src/limitations.md
```diff
@@ -6,7 +6,35 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+end
+```
+and then the type can be updated like
+```jl
+# 2nd version
+struct FooStruct2  # change version here
+    bar::Int
+    str::String
+end
+FooStruct=FooStruct2   # change version here
```
fredrikekre, line 30:
```suggestion
FooStruct = FooStruct2   # change version here
```

docs/src/limitations.md
```diff
@@ -8,6 +8,34 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+struct FooStruct2  # change version here
+    bar::Int
+    str::String
+end
+FooStruct = FooStruct2   # change version here
+function processFoo(foo::FooStruct)  # no need to change this
+    @info foo.bar
+end
+```
+This works as long as the new type name doesn' t conflict with an existing name; within a session you need to change the name each time you change the definition.
```
timholy, line 35:
```suggestion
This works as long as the new type name doesn't conflict with an existing name; within a session you need to change the name each time you change the definition.
```

timholy (approved):

Issue #659, opened 2021-12-15
Revise should track `NetworkOptions`


t-bltg:
```julia
julia> Revise.track(NetworkOptions)
ERROR: no Revise.track recipe for module NetworkOptions
[...]
```

Reference PR: https://github.com/JuliaLang/julia/pull/38362.

I'd be happy to propose a PR, but I do not know how to determine the `v"1.6.0-DEV.????"` version as in: https://github.com/timholy/Revise.jl/blob/e4503d3381e4068fb5d11a065c28001c0e038d50/src/recipes.jl#L201

fredrikekre:
You figure out the relevant commit (from the PR you linked: [`e486c23829b1c4c3a7d73a2266e4ff77e060da2f`](https://github.com/JuliaLang/julia/commit/e486c23829b1c4c3a7d73a2266e4ff77e060da2f)) and then run
```
$ ./contrib/commit-name.sh e486c23829b1c4c3a7d73a2266e4ff77e060da2f
1.6.0-DEV.1468
```
from the root of the Julia repo.

t-bltg:
Good to know, thanks :+1: , will make the PR.

PR #660, opened 2021-12-15
Add stdlib NetworkOptions to recipes

t-bltg:
Fix https://github.com/timholy/Revise.jl/issues/659.


fredrikekre:
Do you want to update the version too? I believe a patch bump is appropriate.

fredrikekre (approved):

timholy:
Thanks to both!

PR #661, opened 2021-12-16
update to JuliaInterpreter@0.9, drop old Julia support

aviatesk:


pfitzseb:
Should change julia compat to 1.6 and disable CI for older Julia versions as well.

aviatesk:
Ah, right. Will try some cleanup.

aviatesk:
So now this seems to be good to go.

timholy:
Thanks so much for doing this!

PR #669, opened 2022-02-03
Simplify precompiles

timholy:
This switches to "do a little work" for precompilation,
rather than explicitly spelling out the specific signatures.
This should be more maintainable and potentially more
exhaustive (or could be made to be so) since it
precompiles across runtime dispatch boundaries.

This also uses `Base.inferencebarrier` in several places to
prevent an argument from being specialized.

I was using this as a test case for https://github.com/JuliaLang/julia/pull/43990 to see if I can eliminate all inference from Revise. It turns out there's one CodeInstance whose inferred code is either not cached or is deleted after compilation. Needs investigation. But the total inference time is <50ms, so in practice it's not a big deal.

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -255,11 +255,7 @@ function methods_by_execution!(@nospecialize(recurse), methodinfo, docexprs, fra
         if isa(stmt, Expr)
             head = stmt.head
             if head === :toplevel
-                local value
-                for ex in stmt.args
-                    ex isa Expr || continue
-                    value = methods_by_execution!(recurse, methodinfo, docexprs, mod, ex; mode=mode, disablebp=false, skip_include=skip_include)
-                end
+                value = handle_toplevel(recurse, methodinfo, docexprs, mod, stmt, mode, skip_include)
                 isassign(frame, pc) && assign_this!(frame, value)
                 pc = next_or_nothing!(frame)
             # elseif head === :thunk && isanonymous_typedef(stmt.args[1])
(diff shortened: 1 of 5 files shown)


timholy:
I've investigated with help from @aviatesk, and determined that the stubborn MethodInstance was having it's compiled code discarded because of a `LimitedAccuracy` annotation. This was pretty easy to work around (though documentation on this point will need to be added to, e.g., SnoopCompile). Now the total inference time is <5ms on https://github.com/JuliaLang/julia/pull/43990. :v:

src/lowered.jl
```diff
@@ -122,7 +122,7 @@ end
 function methods_by_execution(mod::Module, ex::Expr; kwargs...)
     methodinfo = MethodInfo()
     docexprs = DocExprs()
-    value, frame = methods_by_execution!(JuliaInterpreter.Compiled(), methodinfo, docexprs, mod, ex; kwargs...)
+    value, frame = methods_by_execution!(Base.inferencebarrier(JuliaInterpreter.Compiled()), methodinfo, docexprs, mod, ex; kwargs...)
```
aviatesk, line 125: This seems to be an actual usecase example of https://github.com/JuliaLang/julia/pull/41931. As far as I understand, does it turn out to be effective to turn off inference of `recurse(...)` within a callstack of `methods_by_execution!`? If `recurse(...)` is less-frequently called actually, then that sounds a reasonable hack to improve latency.

timholy: Yes, this is exactly a `@noinfer`. All the execution time specific to a particular `recurse` is in JuliaInterpreter anyway.

timholy:
Wow, the GC corruption on 1.7 is a bit scary---we might need to debug that and patch 1.7 before merging this.

But the test failure on nightly is https://github.com/JuliaLang/julia/pull/42822, which appears to be deliberate. (CC @IanButterworth.)

timholy:
This isn't necessary anymore

PR #671, opened 2022-02-15
tweak for upcoming `Core.Compiler` change

aviatesk:
https://github.com/JuliaLang/julia/pull/43800 will define new module `EscapeAnalysis`
inside `Core.Compiler`, and it requires this minor special handling.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -149,7 +149,15 @@ function track_subdir_from_git!(pkgdata::PkgData, subdir::AbstractString; commit
             rethrow(err)
         end
         fmod = get(juliaf2m, fullpath, Core.Compiler)  # Core.Compiler is not cached
-        fmod === Core.Compiler && endswith(fullpath, "compiler.jl") && continue  # defines the module, skip
+        if fmod === Core.Compiler
+            endswith(fullpath, "compiler.jl") && continue  # defines the module, skip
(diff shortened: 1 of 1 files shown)


aviatesk:
Please do not merge this until we merge #43800.

pfitzseb:
Should this just check for `isdefined(Core.Compiler, :EscapeAnalysis)`?

aviatesk:
Right, we should have that static check also.

timholy (approved):

ianatol:
@timholy Can this be merged? Revise is broken on master without it

timholy:
@aviatesk do you want to make the new release? Just in case you know stuff that needs changing before we do that. I'm not running on the very latest master as I continue to work on getting https://github.com/JuliaLang/julia/pull/43990 over the finish line.

aviatesk:
It seems like JuliaInterpreter needs a bit more update to work on 1.8. We should probably wait for it to be updated before making the new release of Revise.

aviatesk:
@timholy Now Revise test suite runs successfully both on 1.7 and the very latest master, I'd like to ask you to publish the new release (maybe the patch one)?

timholy:
I don't know why CI is hanging on nightly, but I did just check that it works locally, so I'll tag.

PR #672, opened 2022-02-19
Support module-style 2-argument includes

### LINKED ISSUE #670: Revise gives MethodError on "module" 2-argument includes
The `include(mapexpr, path)` case is discussed at #634, but similarly, Revise can't handle `include(m::Module, path)` either.

See https://github.com/JuliaLang/julia/pull/43800#discussion_r802173902 for an example.

For the module case, I tried the naive fix of modifying `methods_by_execution` as follows:

```julia
elseif skip_include && (f === modinclude || f === Base.include || f === Core.include)
                    # include calls need to be managed carefully from several standpoints, including
                    # path management and parsing new expressions
                    mod_or_path = @lookup(frame, stmt.args[2])
                    if isa(mod_or_path, Module) # include(mod, pathname)
                        add_includes!(methodinfo, mod_or_path, @lookup(frame, stmt.args[3]))
                    else
                        add_includes!(methodinfo, mod, mod_or_path)
                    end
                    assign_this!(frame, nothing)  # FIXME: the file might return something different from `nothing`
                    pc = next_or_nothing!(frame)
```

This passes Revise tests but doesn't seem to correctly give Revise access to the names from the `include`, as we hit a `UndefVarError` in the EscapeAnalysis example above. Will update if I can fix later today!

ianatol:
Fixes #670

CC: @aviatesk

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -388,10 +388,27 @@ function methods_by_execution!(@nospecialize(recurse), methodinfo, docexprs, fra
                     end
                     assign_this!(frame, value)
                     pc = next_or_nothing!(frame)
-                elseif skip_include && (f === modinclude || f === Base.include || f === Core.include)
+                elseif skip_include && (f === modinclude || f === Core.include)
                     # include calls need to be managed carefully from several standpoints, including
                     # path management and parsing new expressions
-                    add_includes!(methodinfo, mod, @lookup(frame, stmt.args[2]))
+                    if length(stmt.args) == 2
+                        add_includes!(methodinfo, mod, @lookup(frame, stmt.args[2]))
+                    else
+                        error("include(mapexpr, path) is not supported") # TODO (issue #634)
+                    end
+                    assign_this!(frame, nothing)  # FIXME: the file might return something different from `nothing`
+                    pc = next_or_nothing!(frame)
+                elseif skip_include && f === Base.include
+                    if length(stmt.args) == 2
+                        add_includes!(methodinfo, mod, @lookup(frame, stmt.args[2]))
+                    else # either include(module, path) or include(mapexpr, path)
+                        mod_or_mapexpr = @lookup(frame, stmt.args[2])
+                        if isa(mod_or_mapexpr, Module)
+                            add_includes!(methodinfo, mod_or_mapexpr, @lookup(frame, stmt.args[3]))
+                        else
+                            error("include(mapexpr, path) is not supported")
+                        end
+                    end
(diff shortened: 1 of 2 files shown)


src/lowered.jl
```diff
@@ -391,7 +391,12 @@ function methods_by_execution!(@nospecialize(recurse), methodinfo, docexprs, fra
                 elseif skip_include && (f === modinclude || f === Base.include || f === Core.include)
                     # include calls need to be managed carefully from several standpoints, including
                     # path management and parsing new expressions
-                    add_includes!(methodinfo, mod, @lookup(frame, stmt.args[2]))
+                    mod_or_path = @lookup(frame, stmt.args[2])
+                    if isa(mod_or_path, Module) # include(mod, pathname)
```
timholy, line 395: More directly, can you first check whether `stmt` has 2 or 3 args?

The other issue is that there is another 2-arg invocation of `include`:

```julia
help?> include(f, fname)
  include([mapexpr::Function,] path::AbstractString)

  Evaluate the contents of the input source file in the global scope of the containing module. Every module (except those defined with baremodule) has its own definition of include, which evaluates the
  file in that module. Returns the result of the last evaluated expression of the input file. During including, a task-local include path is set to the directory containing the file. Nested calls to
  include will search relative to that path. This function is typically used to load source interactively, or to combine files in packages that are broken into multiple source files. The argument path
  is normalized using normpath which will resolve relative path tokens such as .. and convert / to the appropriate path separator.

  The optional first argument mapexpr can be used to transform the included code before it is evaluated: for each parsed expression expr in path, the include function actually evaluates mapexpr(expr).
  If it is omitted, mapexpr defaults to identity.

  Use Base.include to evaluate a file into another module.

  │ Julia 1.5
  │
  │  Julia 1.5 is required for passing the mapexpr argument.
```

Perhaps we should distinguish `include` from `Base.include` since in fact these don't seem to be the same function.

timholy:
> WIP as the test gives UndefVarError: B670 not defined

You get the same error if you execute this without Revise, no? (Just from the julia prompt.) I don't think we support running additional commands after close-of-module when precompiling a package.

Perhaps try adding an internal module to B670 and then execute this as `Base.include(Internal, ...)` from within `B670`.

timholy:
From my reading, your test doesn't use the 2-argument form of `include` in code that Revise sees. Calling it from within the test code directly isn't visible to Revise.

timholy:
Thanks, @ianatol!

ianatol:
Thanks for the help along the way!

PR #673, opened 2022-02-19
don't run `Broken dependencies` test on 1.8

aviatesk:
This functionality won't be available on 1.8 since it seems like
a package can't be loaded at all when its precompilation failed
because of a broken dependency in v1.8 and higher.


aviatesk:
ok, then it would be reasonable to skip this test case on 1.8 and higher.
Thanks for your confirmation!

PR #679, opened 2022-04-19
Remove dependency on `Libc.TimeVal` by directly calling `time()` instead

Seelengrab:
As far as I can tell, this has the exact same functionality but allows Base.Libc to move to `timespec` internally (and get rid of deprecated unix functionality). Ref https://github.com/JuliaLang/julia/pull/45023. Shouldn't break anything, as this is internal, right?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -108,13 +108,8 @@ function pushex!(exsigs::ExprsSigs, ex::Expr)
 end
 
 ## WatchList utilities
-function systime()
-    # It's important to use the same clock used by the filesystem
-    tv = Libc.TimeVal()
-    tv.sec + tv.usec/10^6
-end
 function updatetime!(wl::WatchList)
-    wl.timestamp = systime()
+    wl.timestamp = time()
 end
 Base.push!(wl::WatchList, filenameid::Pair{<:AbstractString,PkgId}) =
     push!(wl.trackedfiles, filenameid)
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -108,21 +108,16 @@ function pushex!(exsigs::ExprsSigs, ex::Expr)
 end
 
 ## WatchList utilities
-function systime()
-    # It's important to use the same clock used by the filesystem
```
Seelengrab, line 112: 8Regarding this comment - even the old `TimeVal` wasn't using the same clock as the file system, at least not on windows. Regardless, I think moving to `time()` here (which uses `TimeVal` or in the future `TimeSpec` under the hood) is more portable anyway than relying on whatever `Libc` returns.

timholy:
Sorry for the delay here. I can't find any evidence that it's a bad idea so :shrug:

I'm not sure why the tests didn't run, let's see if a close/reopen triggers them (or at least an approval button).

Seelengrab:
Looks like the close/reopen worked - are those failures related/benign? They don't seem to occur with code running here..

Seelengrab:
Thank you! I'm not too familiar with the impact a higher precision could have here - naively I didn't think it would impact things, so maybe it "just" uncovers other bugs?

Seelengrab:
Any update on this?

t-bltg:
Maybe rebase this PR against master and see if CI passes ?

t-bltg:
`nightly` failures are expected (https://github.com/timholy/Revise.jl/pull/694).

timholy:
Let's try this

PR #680, opened 2022-05-02
Tracking non-.jl source files

oscardssmith:
This PR removes the requirement that `file` be a `::AbstractString` which makes it possible to teach Revise about non-julia source files by making your own file type (eg `PyFile`).


src/pkgs.jl
```diff
@@ -147,6 +147,7 @@ maybe_extract_sigs!(pkgdata::PkgData, file::AbstractString) = maybe_extract_sigs
 
 function maybe_add_includes_to_pkgdata!(pkgdata::PkgData, file::AbstractString, includes; eval_now::Bool=false)
     for (mod, inc) in includes
+        @show inc
```
KristofferC, line 150:
```suggestion
```

KristofferC:
It would be good to document what are the exact methods one need to overload in order to hook up a custom file format to Revise. And then to finish it off, add a test as well.

KristofferC:
(fatfingered the close button)

oscardssmith:
Should I add a section to the developer reference?

KristofferC:
I think that is a good place, yes.

oscardssmith:
Just so I don't forget, here is a relatively minimal example of using this feature
```
struct MyFile
    file::String
end
Base.abspath(file::MyFile) = MyFile(Base.abspath(file.file))
Base.isfile(file::MyFile) = Base.isfile(file.file)
Base.isabspath(file::MyFile) = Base.isabspath(file.file)
Base.findfirst(str::String, file::MyFile) = Base.findfirst(str, file.file)
Base.String(file::MyFile) = file.file
function make_module(file::MyFile)
    Expr(:toplevel, :(baremodule :test
        Expr(:toplevel,
            :(include($(file.file)))
            end)
        end), :(using .test))
end
function Base.include(mod::Module, file::MyFile)
    Core.eval(mod, make_module(file))
end
Base.include(file::MyFile) = Base.include(Core.Main, file)
function Revise.parse_source!(mod_exprs_sigs::Revise.ModuleExprsSigs, file::MyFile, mod::Module; kwargs...)
    ex = make_module(file)
    Revise.process_source!(mod_exprs_sigs, ex, file, mod; kwargs...)
end
Revise.is_same_file(a::MyFile, b::String) = a.file == b
```

oscardssmith:
@KristofferC can you re-run CI?

oscardssmith:
for reference, this depends on https://github.com/timholy/CodeTracking.jl/pull/95

test/fake_lang/non_jl_test.jl
```diff
@@ -0,0 +1,44 @@
...
+function Revise.parse_source!(mod_exprs_sigs::Revise.ModuleExprsSigs, file::MyFile, mod::Module; kwargs...)
+    ex = make_module(file)
+    @show mod_exprs_sigs, file, mod, kwargs
+    Revise.process_source!(mod_exprs_sigs, ex, file, mod; kwargs...)
+end
+m=MyFile("test.program")
+includet(m)
+@test test.y() == "2"
+@test test.x() == "1"
+
```
oscardssmith, line 43: @timholy how can I edit the file from the tests?

oscardssmith: @KristofferC any suggestions?

oscardssmith:
@timholy bump on this? What are your thoughts?

test/fake_lang/non_jl_test.jl
timholy: The way Revise tests work in general is to write a little mini-package, `using` it, sleep for a bit to allow the `mtime` to age a bit, and then overwrite the file with new contents. That doesn't seem to fit this particular test, but is there any reason you couldn't create a test that does depend in some way on the content of a file?

src/packagedef.jl
```diff
@@ -640,16 +641,19 @@ function handle_deletions(pkgdata, file)
...
     mexsold = fi.modexsigs
-    filep = normpath(joinpath(basedir(pkgdata), file))
+    idx = fileindex(pkgdata, file)
+    filep = pkgdata.info.files[idx]
+    if isa(filep, AbstractString)
+        filep = normpath(joinpath(basedir(pkgdata), file))
+    end
     topmod = first(keys(mexsold))
-    fileok = file_exists(filep)
+    fileok = file_exists(String(filep))
```
timholy, line 650: If you're doing it this way, any reason not to use `string` instead of `String`? Conversely, to insulate this from invalidation, `String(filep)::String` might be necessary.

timholy, line 650: If you're doing it this way, any reason not to use `string` instead of `String`? Conversely, to insulate this from invalidation, `String(filep)::String` might be necessary.

test/fake_lang/non_jl_test.jl
oscardssmith: My goal here was to add a test for `@test test.x() == "2"` after modifying the file, but I couldn't figure out a good way to modify the file.

test/fake_lang/non_jl_test.jl
```diff
@@ -0,0 +1,50 @@
...
+Pkg.activate(".")
+Pkg.instantiate()
+
+struct MyFile
+    file::String
+end
+Base.abspath(file::MyFile) = MyFile(Base.abspath(file.file))
+Base.isabspath(file::MyFile) = Base.isabspath(file.file)
+Base.joinpath(str::String, file::MyFile) = Myfile(Base.joinpath(str, file.file))
+Base.normpath(file::MyFile) = Myfile(Base.normpath(file.file))
```
timholy, line 11:
```suggestion
Base.joinpath(str::String, file::MyFile) = MyFile(Base.joinpath(str, file.file))
Base.normpath(file::MyFile) = MyFile(Base.normpath(file.file))
```

src/parsing.jl
```diff
@@ -44,6 +44,10 @@ function parse_source!(mod_exprs_sigs::ModuleExprsSigs, src::AbstractString, fil
         ln = count(isequal('\n'), SubString(src, 1, min(pos, length(src)))) + 1
         throw(LoadError(filename, ln, ex.args[1]))
     end
+    return process_source!(mod_exprs_sigs, ex, filename, mod; kwargs...)
+end
+
+function process_source!(mod_exprs_sigs::ModuleExprsSigs, @nospecialize(ex), filename, mod::Module; mode::Symbol=:sigs)
```
timholy, line 50: Why is this `@nospecialize(ex)`? `ExprSplitter` has to take an `Expr` as its second argument, and calling `ExprSplitter` is the very first thing you do with `ex`.

test/fake_lang/non_jl_test.jl
```diff
@@ -0,0 +1,50 @@
...
+    @show mod_exprs_sigs, file, mod, kwargs
+    Revise.process_source!(mod_exprs_sigs, ex, file, mod; kwargs...)
+end
+try
+    cp("test.program", "tmp.program")
+    m=MyFile("tmp.program")
+    includet(m)
+    @test test.y() == "2"
+    @test test.x() == "1"
+    mv("new_test.program.old", "tmp.program", force=true)
```
timholy, line 45:
```julia
ERROR: LoadError: IOError: open("new_test.program.old", 0, 0): no such file or directory (ENOENT)
Stacktrace:
  [1] uv_error
    @ ./libuv.jl:97 [inlined]
  [2] open(path::String, flags::UInt8, mode::Int64)
    @ Base.Filesystem ./filesystem.jl:105
  [3] open
    @ ./filesystem.jl:97 [inlined]
  [4] sendfile(src::String, dst::String)
    @ Base.Filesystem ./file.jl:992
  [5] cp(src::String, dst::String; force::Bool, follow_symlinks::Bool)
    @ Base.Filesystem ./file.jl:384
  [6] rename(src::String, dst::String; force::Bool)
    @ Base.Filesystem ./file.jl:981
  [7] rename
    @ ./file.jl:977 [inlined]
  [8] #mv#15
    @ ./file.jl:426 [inlined]
  [9] top-level scope
    @ ~/.julia/dev/Revise/test/fake_lang/non_jl_test.jl:45
 [10] include(fname::String)
    @ Base.MainInclude ./client.jl:476
 [11] top-level scope
    @ REPL[4]:1
in expression starting at /home/tim/.julia/dev/Revise/test/fake_lang/non_jl_test.jl:39
```

The source file, `"new_test.program.old"`, doesn't exist. Did you forget to commit a file?

src/parsing.jl
```diff
@@ -35,7 +35,7 @@ string. `pos` is the 1-based byte offset from which to begin parsing `src`.
 
 See also [`Revise.parse_source`](@ref).
 """
-function parse_source!(mod_exprs_sigs::ModuleExprsSigs, src::AbstractString, filename::AbstractString, mod::Module; mode::Symbol=:sigs)
+function parse_source!(mod_exprs_sigs::ModuleExprsSigs, src::AbstractString, filename::AbstractString, mod::Module; kwargs...)
```
timholy, line 38: Do you expect to call this? My understanding is you're intercepting the method on line 22. If we can change line 27 to
```diff
diff --git a/src/parsing.jl b/src/parsing.jl
index 32ef509..34b9019 100644
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -24,7 +24,7 @@ function parse_source!(mod_exprs_sigs::ModuleExprsSigs, filename::AbstractString
         @warn "$filename is not a file, omitting from revision tracking"
         return nothing
     end
-    parse_source!(mod_exprs_sigs, read(filename, String), filename, mod; kwargs...)
+    parse_source!(mod_exprs_sigs, read(filename, String)::String, String(filename)::String, mod; kwargs...)
 end
 
 """
```
(together with the `@nospecialize(ex)` fix later in this file) then that gets rid of the main new source of invalidation risk.

timholy (commented):
This mostly looks good. I'm proposing a couple of changes to decrease the risk of invalidation of Revise's own code when people load packages.

We do need the test working before we can merge this. I think you also need to run this test from `runtests.jl`?

src/parsing.jl
oscardssmith: Good catch. can you re-run CI?

test/fake_lang/non_jl_test.jl
oscardssmith: Good catch. Can you re-run CI?

oscardssmith:
One complication here is that these tests won't pass until https://github.com/timholy/CodeTracking.jl/pull/95 is merged and tagged.

src/packagedef.jl
```diff
@@ -910,13 +914,13 @@ function track(mod::Module, file::AbstractString; mode=:sigs, kwargs...)
             CodeTracking._pkgfiles[id] = pkgdata.info
         end
         push!(pkgdata, relpath(file, pkgdata)=>FileInfo(fm))
-        init_watching(pkgdata, (file,))
+        init_watching(pkgdata, (String(file),))
```
timholy, line 917: If `file` is only known to be `Any`, this should be `String(file)::String`. Sadly there is no guarantee that constructors return an object of the given type, https://github.com/JuliaLang/julia/issues/42372.

oscardssmith:
Invalidations fixed. Tests pass (with the codetracking PR+Julia1.7. On 1.9, the tests fail with or without this PR, but including my test file passes).

src/packagedef.jl
```diff
@@ -520,6 +520,7 @@ Use the `pkgdata` version if the files are supplied using relative paths.
 function init_watching(pkgdata::PkgData, files=srcfiles(pkgdata))
     udirs = Set{String}()
     for file in files
+        file = String(file)
```
timholy, line 523:
```suggestion
        file = String(file)::String
```

src/packagedef.jl
```diff
@@ -643,16 +644,19 @@ function handle_deletions(pkgdata, file)
...
     mexsold = fi.modexsigs
-    filep = normpath(joinpath(basedir(pkgdata), file))
+    idx = fileindex(pkgdata, file)
+    filep = pkgdata.info.files[idx]
+    if isa(filep, AbstractString)
+        filep = normpath(joinpath(basedir(pkgdata), file))
+    end
     topmod = first(keys(mexsold))
-    fileok = file_exists(filep)
+    fileok = file_exists(String(filep))
```
timholy, line 653:
```suggestion
    fileok = file_exists(String(filep)::String)
```

src/packagedef.jl
timholy: You marked this as resolved but didn't implement the suggested change. Please don't do that.

src/packagedef.jl
```diff
@@ -913,13 +917,13 @@ function track(mod::Module, file::AbstractString; mode=:sigs, kwargs...)
             CodeTracking._pkgfiles[id] = pkgdata.info
         end
         push!(pkgdata, relpath(file, pkgdata)=>FileInfo(fm))
-        init_watching(pkgdata, (file,))
+        init_watching(pkgdata, (String(file),))
```
timholy, line 920:
```suggestion
        init_watching(pkgdata, (String(file)::String,))
```

src/parsing.jl
timholy: Here too, please don't mark as resolved if you don't answer the query.

Your docs only describe overriding the call on line 22. If you are calling this, you might need to expand your docs: what `ExprSplitter` method do people need to create? Any other methods needed? Conversely, if they don't need to override `ExprSplitter`, then that suggests `ex` is an `Expr`, and changing this to `@nospecialize(ex)` hurts the quality of type inference and makes these methods more vulnerable to invalidation.

timholy (changes requested):
There still seems to be some unnecessary invalidation risk here. *Realized* invalidations depend on which package you load; you might try InlineStrings.jl as a package that defines a lot of new string types and methods, e.g.,
```julia
using SnoopCompileCore
using Revise    # Revise precompiles a lot, so you're probably OK, otherwise force a revision to make sure everything is compiled before the next step
invs = @snoopr using InlineStrings
using SnoopCompile
trees = invalidation_trees(invs)
```

oscardssmith:
apologies. the things I resolved were changes that I made that I then lost while rebasing (and didn't check)

src/parsing.jl
oscardssmith: They don't need to override `ExprSplitter`. Is putting `ex = # julia Expr returned from running transpiler` sufficient?

test/non_jl_test.jl
```diff
@@ -0,0 +1,48 @@
...
+
+using Revise
+function Revise.parse_source!(mod_exprs_sigs::Revise.ModuleExprsSigs, file::MyFile, mod::Module; kwargs...)
+    ex = make_module(file)
+    Revise.process_source!(mod_exprs_sigs, ex, file, mod; kwargs...)
+end
+
+path = "test.program"
+try
+    cp(joinpath("fake_lang", "test.program"), path, force=true)
```
timholy, line 36:
```suggestion
    cp(joinpath(@__DIR__, "fake_lang", "test.program"), path, force=true)
```
might fix the CI path error.

timholy:
That fixed the first path issue but there are more. Presumably it fails locally in the same way with `Pkg.test()`?

oscardssmith:
No, this is passing locally (even if I don't run the tests from the Revise folder).

test/non_jl_test.jl
```diff
@@ -0,0 +1,48 @@
...
+end
+Base.include(file::MyFile) = Base.include(Core.Main, file)
+
+using Revise
+function Revise.parse_source!(mod_exprs_sigs::Revise.ModuleExprsSigs, file::MyFile, mod::Module; kwargs...)
+    ex = make_module(file)
+    Revise.process_source!(mod_exprs_sigs, ex, file, mod; kwargs...)
+end
+
+path = "test.program"
```
pfitzseb, line 34:
```suggestion
path = joinpath(@__DIR__, "test.program")
```
might help?

timholy:
OK. Then try the same trick with all the rest of your paths?

timholy:
You have another one on line 42?

test/non_jl_test.jl
```diff
@@ -0,0 +1,48 @@
...
+
+path = joinpath(@__DIR__, "test.program")
+try
+    cp(joinpath(@__DIR__, "fake_lang", "test.program"), path, force=true)
+    m=MyFile(path)
+    includet(m)
+    Revise.revise()
+    @test fake_lang.y() == "2"
+    @test fake_lang.x() == "1"
+    cp(joinpath("fake_lang", "new_test.program"), path, force=true)
```
pfitzseb, line 42:
```suggestion
    cp(joinpath(@__DIR__, "fake_lang", "new_test.program"), path, force=true)
```

pfitzseb:
Non-nightly tests seem to pass now.

oscardssmith:
Thanks so much!

PR #688, opened 2022-06-26
Track packages compiled into sysimage

### LINKED ISSUE #685: Track packages that are compiled into sysimage with PackageCompiler.jl
Is there a way to automatically track changes in packages after they are in a sysimage? My current workaround is:

```
const SYSIMAGE_TIME = time()

function track_sysimage_package(p)
    id = Base.PkgId(p)
    Revise.watch_package(id)
    pkgdata = Revise.pkgdatas[id]
    for file in Revise.srcfiles(pkgdata)
        fullfile = joinpath(Revise.basedir(pkgdata), file)
        if stat(fullfile).mtime > SYSIMAGE_TIME
            push!(Revise.revision_queue, (pkgdata, file))
        end
    end
end
```

Is there a better way?

ktdq:


ktdq:
Fix #685

src/recipes.jl
```diff
@@ -94,7 +95,14 @@ function _track(id, modname; modified_files=revision_queue)
         track_subdir_from_git!(pkgdata, compilerdir; modified_files=modified_files)
         # insertion into pkgdatas is done by track_subdir_from_git!
     else
-        error("no Revise.track recipe for module ", modname)
+        pkgdata = watch_package(id)
+        modtime = mtime(Base.pkgorigins[id].cachepath)
```
timholy, line 99: What happens if you do this?

```julia
julia> module EmptyModule end
Main.EmptyModule

julia> Revise.track(EmptyModule)
ERROR: no Revise.track recipe for module EmptyModule
Stacktrace:
 [1] error(::String, ::Symbol)
   @ Base ./error.jl:44
 [2] _track(id::Base.PkgId, modname::Symbol; modified_files::Set{Tuple{Revise.PkgData, String}})
   @ Revise ~/.julia/packages/Revise/VskYC/src/recipes.jl:97
 [3] #track#44
   @ ~/.julia/packages/Revise/VskYC/src/recipes.jl:12 [inlined]
 [4] track(mod::Module)
   @ Revise ~/.julia/packages/Revise/VskYC/src/recipes.jl:9
 [5] top-level scope
   @ REPL[2]:1
```

timholy (commented):
I'm not a PackageCompiler user, but this seems great! Many thanks for contributing this.

I left one comment, but otherwise:
- this could use a test. You could test the error condition I proposed in my comment, and for a working case you could manually create a `Base.pkgorigins` entry. I know that writing these tests is a little annoying but it's the only way to keep Revise working across Julia versions (it breaks with literally every new Julia point release since Revise and its supporting infrastructure reach deep into Julia's guts).
- this seems [NEWS](https://github.com/timholy/Revise.jl/blob/master/NEWS.md)-worthy to me

timholy:
Incorporated into #1063. Thanks @ktdq !

Issue #689, opened 2022-06-26
Failing `includet`


t-bltg:
Consider the following reproducer:

**main.jl**
```julia
# do_include = "includet"  # or Revise.includet
do_include = "Base.include"  # uncomment => works

#######################################
common = joinpath(pwd(), "common.jl")
write(common, """
module Common
    const foo = 2
end
""")

routines = joinpath(pwd(), "routines.jl")
write(routines, """
module Routines
    using Revise
    $do_include(@__MODULE__, "$common")
    using .Common
end
""")

codes = joinpath(pwd(), "codes.jl")
write(codes, """
module Codes
    using Revise
    $do_include(@__MODULE__, "$common")
    using .Common
end
""")

driver = joinpath(pwd(), "driver.jl")
write(driver, """
module Driver
    using Revise
    $do_include(@__MODULE__, "$routines")
    using .Routines
    $do_include(@__MODULE__, "$codes")
    using .Codes
end
""")
#######################################

using Revise

Revise.includet(@__MODULE__, "driver.jl")
using .Driver
```

It fails (on `1.6` or `1.7`) with the following stacktrace, whereas using `Base.include` (see commented line in reproducer) works:
```julia
$ julia main.jl
ERROR: UndefVarError: Common not defined
Stacktrace:
 [1] top-level scope
   @ [...]/codes1d.jl:4
in expression starting at [...]/codes1d.jl:4
```

Issue #561, opened 2020-10-17
Issue with revise?


PallHaraldsson:
Hi,

I hit a rather weird error in VS Code, that I could reproduce again and again, scratching my head why, until I meant to show a colleague and then it was gone (I did something minor in-between)! So I can't show you either.

Anyway, I didn't really implicate Revise yesterday, but thought there might be some corruption going on with Julia until I see now trying in the same REPL (EDIT: see next comment, Revise probably just a symptom):

```
julia> 1/3
ERROR: UndefVarError: Base_Docs_4352b6d8 not defined
Stacktrace:
 [1] _invoked_shouldlog(logger::Revise.ReviseLogger, level::Base.CoreLogging.LogLevel, _module::Module, group::Symbol, id::Symbol)
   @ Base.CoreLogging logging.jl:78
 [2] doc!(__module__::Module, b::Base.Docs.Binding, str::Base.Docs.DocStr, sig::Type{Union{}})
   @ Base.Docs logging.jl:361
 [3] top-level scope
   @ ~/.julia/packages/JLLWrappers/m1BKg/src/toplevel_generators.jl:79Revise evaluation error at /home/pharaldsson_sym/.julia/packages/JLLWrappers/m1BKg/src/toplevel_generators.jl:79

Stacktrace:
  [1] methods_by_execution!(recurse::Any, methodinfo::Revise.CodeTrackingMethodInfo, docexprs::Dict{Module, Vector{Expr}}, mod::Module, ex::Expr; mode::Symbol, disablebp::Bool, always_rethrow::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Revise ~/.julia/dev/Revise/src/lowered.jl:230
  [2] eval_with_signatures(mod::Module, ex::Expr; mode::Symbol, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Revise ~/.julia/dev/Revise/src/packagedef.jl:455
  [3] instantiate_sigs!(modexsigs::OrderedDict{Module, OrderedDict{Revise.RelocatableExpr, Union{Nothing, Vector{Any}}}}; mode::Symbol, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Revise ~/.julia/dev/Revise/src/packagedef.jl:463
[...]

Issue #505, opened 2020-06-11
type #pynamespace has no field mod error


marius311:
I'm getting the following error on Julia 1.4.2 / Revise 2.7.2 which is 100% reproducible and doesn't involve changing any code, its just triggered when these packages are loaded. Is this error message enough for an expert to dechiper what's going on? If not let me know and I will work on coming up with a MWE. 

The crux of it is that I have something like

```julia
module Ultradeep100d
@init @require Py2Call begin
   using .Py2Call
   @eval py2"..."
end
end
```

where Py2Call is another of my packages that provides the `py2"..."` string macro (which calls Python 2 using PyCall and an auxilliary Julia worker), the macro itself expanding to include another call to `@eval` and `PyCall.@py_str`. 

To trigger the bug I do `using Revise, Ultradeep100d, Py2Call` and I get the error below. The reference to `pynamespace` in the actual error must be [this one](https://github.com/JuliaPy/PyCall.jl/blob/c442738dcef680412ac726ea5595b25d46dd301e/src/pyeval.jl#L228). 

Again, let me know if a MWE is needed.

<details>

```
julia> using Revise, Ultradeep100d, Py2Call
Revise.LogRecord(Error, evaluation error starting at /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:335, lowered, Revise_e943ed8b, "/global/homes/m/marius/.julia/packages/Revise/tV8FE/src/lowered.jl", 106, (mod=Ultradeep100d, ex=begin
    begin
        #= /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:333 =#
        using Py2Call
        #= /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:335 =#
        #= /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:335 =# @eval begin
                #= /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:337 =#
                py2"import sptpol_software.observation.sky as sky
from sptpol_software.observation.sky import Map
"
                #= /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:342 =#
                function readFromHDF5(filename; T = Float32, removeWeight = true, flattenPol = true)
                    #= /global/cscratch1/sd/marius/ultradeep100d/src/Ultradeep100d.jl:343 =#
[...]

PR #690, opened 2022-06-26
fix `includet` using `mod` argument

t-bltg:
Fix https://github.com/timholy/Revise.jl/issues/689.

Shouldn't all `PkgId(mod)` be mapped to `PkgId(mod, string(mod)` (else `PkgId(mod)` maps to `PkgId(mod, "Main")`) ?

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -875,7 +875,7 @@ If this produces many errors, check that you specified `mod` correctly.
 function track(mod::Module, file::AbstractString; mode=:sigs, kwargs...)
     isfile(file) || error(file, " is not a file")
     # Determine whether we're already tracking this file
-    id = PkgId(mod, string(mod))
+    id = Base.moduleroot(mod) == Main ? PkgId(mod, string(mod)) : PkgId(mod)  # see #689 for `Main`
     if haskey(pkgdatas, id)
         pkgdata = pkgdatas[id]
         relfile = relpath(abspath(file), pkgdata)
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -875,7 +875,7 @@ If this produces many errors, check that you specified `mod` correctly.
 function track(mod::Module, file::AbstractString; mode=:sigs, kwargs...)
     isfile(file) || error(file, " is not a file")
     # Determine whether we're already tracking this file
-    id = PkgId(mod)
+    id = PkgId(mod, string(mod))
     if haskey(pkgdatas, id)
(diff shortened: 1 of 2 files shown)


t-bltg:
Weird bugs on `1`: `  LoadError: UndefVarError: Base_Docs_4352b6d8 not defined`.

Xref:
https://github.com/timholy/Revise.jl/issues/505
https://github.com/timholy/Revise.jl/issues/520
https://github.com/timholy/Revise.jl/issues/561

The error on `1.6` & windows (of course :/) looks related to the test :
```
   in expression starting at C:\Users\RUNNER~1\AppData\Local\Temp\ML0Fd83oSC\driver.jl:3
caused by: LoadError: "invalid escape sequence"
```

timholy:
> Shouldn't all PkgId(mod) be mapped to PkgId(mod, string(mod) (else PkgId(mod) maps to PkgId(mod, "Main")) ?

[Base uses `moduleroot`](https://github.com/JuliaLang/julia/blob/f1991edd30bab79b82381e89a282975ece39d44b/base/pkgid.jl#L12) to get the top-level module. In this case the root really is `Main`, because you're not loading a package.

I think this change would break for a submodule of `Base` (or any package submodule):

```julia
julia> Revise.track(Base)

julia> id = Base.PkgId(Base.Filesystem)
Base [top-level]

julia> pkgdata = Revise.pkgdatas[id];   # works

julia> id = Base.PkgId(Base.Filesystem, string(Base.Filesystem))
Base.Filesystem [top-level]

julia> pkgdata = Revise.pkgdatas[id]
ERROR: KeyError: key Base.Filesystem [top-level] not found
Stacktrace:
 [1] getindex(h::Dict{Base.PkgId, Revise.PkgData}, key::Base.PkgId)
   @ Base ./dict.jl:498
 [2] top-level scope
   @ REPL[10]:1
```

 Would it be worth first trying the version that's already in the code, and then if `haskey(pkgdatas, id)` fails, maybe try your version?

Thanks very much for the test!

t-bltg:
> In this case the root really is Main, because you're not loading a package.

I see, good point.

> Would it be worth first trying the version that's already in the code, and then if haskey(pkgdatas, id) fails, maybe try your version

Thanks for the hint. This is a good idea, I'll try to work something out now.

t-bltg:
Tests on `1.6` are ✔, so I think that this PR is ready.

timholy:
Thanks!

PR #696, opened 2022-07-23
NFC simplify `runtests.jl` (rework `write`s)

t-bltg:
This is a non functional change (NFC), rewriting `runtests.jl` by replacing:

```julia
open(path, "w") do io
  print(io, ...)
end
```
with:

```julia
write(path, ...)
```

Also make indentation consistent (one indentation level) on these `write`s.

I know `NFC`s are controversial, but IMO in this case, removing `400+` lines in a `~4000` lines file is worth considering for readability.


timholy:
This is worth it, thanks!

Issue #697, opened 2022-07-23
Test suite fails `get_def` test


t-bltg:
Against master, a local julia build, where the source files are available in `Revise.builddir`.

```julia
] test Revise
[...]
Revise is currently tracking the following files in ReviseFileNow [top-level]: ["src/ReviseFileNow.jl"]
[ Info: tracking Base
get_def: Test Failed at ~/julia-1.7.3/.julia/packages/Revise/VskYC/test/runtests.jl:2495
  Expression: definition(m) isa Expr
   Evaluated: nothing isa Expr
Stacktrace:
 [1] macro expansion
   @ ~/share/julia/stdlib/v1.7/Test/src/Test.jl:445 [inlined]
 [2] macro expansion
   @ ~/.julia/packages/Revise/VskYC/test/runtests.jl:2495 [inlined]
 [3] macro expansion
   @ ~/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
 [4] macro expansion
   @ ~/.julia/packages/Revise/VskYC/test/runtests.jl:2457 [inlined]
 [5] macro expansion
   @ ~/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
 [6] top-level scope
   @ ~/.julia/packages/Revise/VskYC/test/runtests.jl:72
┌ Warning: skipping git tests because Revise is not under development
└ @ Main ~/.julia/packages/Revise/VskYC/test/runtests.jl:2990
extra
[ Info: tracking Pkg
┌ Warning: REPL tests skipped
└ @ Main ~/.julia/packages/Revise/VskYC/test/runtests.jl:3141
Test Summary:                                   | Pass  Fail  Broken  Total
Revise                                          |  705     1       2    708
[...]
  get_def                                       |    6     1              7
[...]
ERROR: LoadError: Some tests did not pass: 705 passed, 1 failed, 0 errored, 2 broken.
in expression starting at ~/julia-1.7.3/.julia/packages/Revise/VskYC/test/runtests.jl:71
ERROR: Package Revise errored during testing
```

PR #698, opened 2022-07-23
fix `parentfile`

t-bltg:
`parentfile` is inconsistent with `filedata`: on a local build of julia, where `juliadir` points to `basebuilddir`, `parentfile` returns a path starting with `Sys.BINDIR/Base.DATAROOTDIR` because of `Base.find_source_file`.

This is not consistent with `filedata` (`Base._included_files`) which point to files in `basebuilddir`.

- fix https://github.com/timholy/Revise.jl/issues/697;
- add a specific ci test using a forced `juliadir` path (`test/juliadir.jl`).
- update `ci.yaml`, use `@latest` if possible, rewrite some parts for clarity;
- use `julia-actions/cache@v1`;

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,22 +23,13 @@ jobs:
         arch:
           - x64
     steps:
-    - uses: actions/checkout@v2
-    - uses: julia-actions/setup-julia@v1
+    - uses: actions/checkout@v3
+    - uses: julia-actions/setup-julia@latest
       with:
         version: ${{ matrix.version }}
         arch: ${{ matrix.arch }}
         show-versioninfo: ${{ matrix.version == 'nightly' }}
-    - uses: actions/cache@v1
-      env:
-        cache-name: cache-artifacts
-      with:
-        path: ~/.julia/artifacts
-        key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-        restore-keys: |
-          ${{ runner.os }}-test-${{ env.cache-name }}-
-          ${{ runner.os }}-test-
-          ${{ runner.os }}-
+    - uses: julia-actions/cache@v1
     - uses: julia-actions/julia-buildpkg@latest
     # Revise's tests need significant customization
     # Populate the precompile cache with an extraneous file, to catch issues like in #460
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -54,21 +45,41 @@ jobs:
...
-        julia --project --code-coverage=user -e 'ENV["JULIA_REVISE_POLL"]="1"; using Pkg, Revise; include(joinpath(dirname(pathof(Revise)), "..", "test", "polling.jl"))'
+        julia --project --code-coverage=user -e '
+          ENV["JULIA_REVISE_POLL"]="1"
+          using Pkg, Revise
+          include(joinpath(dirname(pathof(Revise)), "..", "test", "polling.jl"))
+        '
         # The REPL wasn't initialized, so the "Methods at REPL" tests didn't run. Pick those up now.
-        TERM="xterm" julia --project --code-coverage=user -e 'using InteractiveUtils, REPL, Revise; @async(Base.run_main_repl(true, true, false, true, false)); while !isdefined(Base, :active_repl_backend) sleep(0.1) end; pushfirst!(Base.active_repl_backend.ast_transforms, Revise.revise_first); include(joinpath("test", "runtests.jl")); REPL.eval_user_input(:(exit()), Base.active_repl_backend)' "Methods at REPL"
+        TERM="xterm" julia --project --code-coverage=user -e '
+          using InteractiveUtils, REPL, Revise
```
t-bltg, line 55: In bash, single (strong) quotes can span multiple lines, this is clearer than a single line statement with semicolons.

test/juliadir.jl
```diff
@@ -0,0 +1,11 @@
+using Revise, InteractiveUtils, Test
```
timholy, line 1: Should this be executed from `runtests.jl`?

timholy:
Thanks!

test/juliadir.jl
t-bltg: Yes I thought it was relevant to add it to 'extra test' section in ci, because we download the julia source files for that version, and force `juliadir` to mimic `basebuilddir` for testing. Since it's a bit of a corner case, and hackery machinery is involved, I think it's relevant to split it from `runtests.jl`.

t-bltg:
Thanks for merging ;)

PR #705, opened 2022-08-19
Fix require nothing bug

lassepe:
I recently hit an [edge case in LazySets.jl](https://github.com/JuliaReach/LazySets.jl/issues/3040) where a `@require Package="<UUID>" nothing` statement would not hit the `add_require` dispatch in Revise because the `nothing` is parsed as a `Symbol`, not an `Expression`. I have thus widened the type constraint in `add_expression` to accept `Union{Expr,Symbol}` and adjusted the body to always convert to an Expr first.

The test that I added fails on master but works with this PR. However, there may be a more meaningful test to write here. I'm not too familiar with the test structure here.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -39,6 +39,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 RoundingIntegers = "d5f540fe-1c90-5db3-b776-2e2f362d9394"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UnsafeArrays = "c4a57d5a-5b31-53a6-b365-19f8c011fbd6"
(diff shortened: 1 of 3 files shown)


lassepe:
I have no idea why the CI fails on this. The tests for the changes that I made pass locally for me.

timholy:
Presumably you need to add it as an `[extras]` and `[targets]` dependency? https://github.com/timholy/Revise.jl/blob/ca3b5cffd26325e3a3fd69932f5dc827ff156a3b/Project.toml#L28-L44

But one issue is that your new test passes for me even on Revise master, so I don't think it's testing what you think it is.

timholy:
This is stale and not obviously needed. Closing for now, happen to reopen if we sharpen it.

PR #710, opened 2022-09-18
test: hanging `1.8` ci

t-bltg:
Test reverting https://github.com/timholy/Revise.jl/pull/679, in order to determine if it is the cause of `1.8.1` failures / hang in ci.
Tests on master are passing locally on `1.8.1`.
They should take around ~3min, but hang in CI (~6h).

1.8.1 ✘ (master):
https://github.com/timholy/Revise.jl/actions/runs/3055718432/jobs/4940473518

1.7.3 (previous commit, without pr) ✔:
https://github.com/timholy/Revise.jl/actions/runs/2767526837/jobs/4349142764

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -358,9 +358,10 @@ end
 
 # These are typically bypassed in favor of expression-by-expression evaluation to
 # allow handling of new `include` statements.
-function eval_new!(exs_sigs_new::ExprsSigs, exs_sigs_old, mod::Module; mode::Symbol=:eval)
+function eval_new!(exs_sigs_new::ExprsSigs, exs_sigs_old, mod::Module; mode::Symbol=:eval, debug=false)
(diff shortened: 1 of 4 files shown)


t-bltg:
It seems that the PR is not to blame, and that the hang is related to the julia `1`: `1.7.3` -> `1.8.1` in workflows.

t-bltg:
The test hangs at `@eval using RoundingIntegers` in https://github.com/timholy/Revise.jl/blob/3b2275c90106e34d63bc7f67b2edbc5724489bd9/test/runtests.jl#L3147.

t-bltg:
And further investigation shows that it hangs here https://github.com/timholy/Revise.jl/blob/3b2275c90106e34d63bc7f67b2edbc5724489bd9/test/runtests.jl#L3101

timholy:
Closed by #722

PR #712, opened 2022-10-26
docs: update dead link

tfiers:
An alternative link is http://web.archive.org/web/20200722031736/https://linuxacademy.com/blog/linux/introduction-using-diff-and-patch/

..but that loads a bit slower


t-bltg:
Thanks !

PR #720, opened 2022-12-28
Wait for repl async to allow for startup.jl loading sequence

### LINKED ISSUE #719: When loaded from startup.jl Revise always waits 1s (times out) waiting for the repl to exist
On julia master, when started from startup.jl `@time using Revise` reports ~1.2s

When started from the repl the time is ~0.2s

It seems that Revise is always waiting for 20 x 0.05s waits until it times out here
https://github.com/timholy/Revise.jl/blob/92d8b636715825f1679d6f896535796f661a6cc7/src/packagedef.jl#L1322-L1325

Is it possible to detect that load came from startup.jl and bypass the check, given the repl won't ever be loaded at that time, I believe?

IanButterworth:
Fixes #719 (speeds up loading Revise from 1.2 to 0.2 seconds when loading from startup.jl)

```
$ ./julia
using Revise: 0.225344 seconds (556.59 k allocations: 36.000 MiB)
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.10.0-DEV.215 (2022-12-28)
 _/ |\__'_|_|_|\__'_|  |  Commit 82fbf54131 (0 days old master)
|__/                   |

julia> includet("test.jl")

julia> foo(1)
1

#edit foo in file

julia> foo(1)
2
```

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1317,14 +1317,22 @@ function __init__()
         else
             pushfirst!(REPL.repl_ast_transforms, revise_first)
             # #664: once a REPL is started, it no longer interacts with REPL.repl_ast_transforms
-            iter = 0
-            # wait for active_repl_backend to exist
-            while !isdefined(Base, :active_repl_backend) && iter < 20
-                sleep(0.05)
-                iter += 1
-            end
             if isdefined(Base, :active_repl_backend)
                 push!(Base.active_repl_backend.ast_transforms, revise_first)
+            else
+                # wait for active_repl_backend to exist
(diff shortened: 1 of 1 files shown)


IanButterworth:
CI seems sad. Seeing what it's like on master here https://github.com/timholy/Revise.jl/pull/721

IanButterworth:
@timholy the CI problems are on master too https://github.com/timholy/Revise.jl/pull/721

timholy:
Lovely, thanks so much!

PR #721, opened 2022-12-29
Check CI on master

IanButterworth:


timholy:
xref https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/560

KristofferC:
Let's rerun after https://github.com/JuliaRegistries/General/pull/74784

timholy:
Note that we'll get some hangs, see #722. But those shouldn't affect nightly.

timholy:
This won't be necessary, more changes coming.

PR #722, opened 2022-12-29
Circumvent CI hang on 1.8

timholy:
#710 attempted to debug a Requires-based hang that seems
to only occur on CI with Julia 1.8. It may be due to lock contention,
but the precise cause was not obvious.

This just disables the problematic test.
Since it doesn't happen on nightly, and since Requires will be
supplanted by the new weakdeps mechanism, we can direct any
users who encounter this bug to the alternatives.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3144,18 +3144,20 @@ do_test("New files & Requires.jl") && @testset "New files & Requires.jl" begin
     @test isempty(read(warnfile, String))
     # Issue #431
     @test_throws UndefVarError TrackRequires.ftrack()
-    @eval using RoundingIntegers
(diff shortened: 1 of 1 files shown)


timholy:
CC @t-bltg

timholy:
These failures are different and will be fixed by #561. Looks like I pushed this just a few minutes before a new JuliaInterpreter was registered.

t-bltg:
Yes, it's probably best if we can move forward, and rework if we encounter a real case scenario bug related to this.

> Looks like I pushed this just a few minutes before a new JuliaInterpreter was registered.

Let's rerun ci ?

timholy:
This does not seem to be a regression, so let's merge.

PR #723, opened 2023-01-04
Rely on code-edge dependency analysis

timholy:
Formerly we excluded interpretation of type definitions, solely because they didn't seem to be needed (in most cases) and because omitting them avoided some problems of type re-definition. However, Julia has become more sophisticated about avoiding re-definition and LoweredCodeUtils is significantly better at determining code-dependencies. Currently `test/sigtest.jl` has failures that stem from referencing types only defined on Windows on a Linux system; with this PR, we rely on LoweredCodeUtils to discover that this code shouldn't be executed, thus fixing the bug.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Revise"
 uuid = "295af30f-e4ad-537b-8983-00126c2a3abe"
-version = "3.4.0"
+version = "3.5.0"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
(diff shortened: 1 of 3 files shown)


aviatesk (approved):

timholy:
@oscardssmith, can you look at why the `fake_lang` test always fails on mac? (e.g., https://github.com/timholy/Revise.jl/actions/runs/3836673432/jobs/6531099485) I don't have easy access to one.

timholy:
[it's not easy getting to green](https://www.youtube.com/watch?v=rRZ-IxZ46ng)

timholy:
All failures except the closed-stream issue have been fixed. I think we can merge.

PR #725, opened 2023-01-06
Add more logging for "stream is closed or unusable"

timholy:
Currently CI errors (only on Julia nightly) with `UNHANDLED TASK ERROR: IOError: stream is closed or unusable`. 

I can't replicate this error locally, so it's not easy to debug.


timholy:
This doesn't do anything other than clarify where the errors are coming from, but that's not a bad thing.

PR #731, opened 2023-02-03
Do not track files during pre-compilation

### LINKED ISSUE #728: discourage packages from depending on Revise?
There are packages embedding Revise as a dependency (https://juliahub.com/ui/Packages/Revise/M1Qoh/3.5.0?page=2). This seems like generally a user-hostile API design for these packages, since they are implicitly forcing the user to depend on Revise for a random subset of packages (whatever happens to get loaded after Revise happened to get loaded).

This also can show up as a bug during precompilation, since Revise starts tracking some files in the precompile process, which may lead to incorrect evaluation (attempting to load and evaluate files that should not be part of the package) and other similar issues.

I wonder if we should define `__init__() = precompile(false)` here, so that those packages are discouraged from depending on a package which makes that precompilation less reliable (this one)

LilithHafner:
This might be a very bad idea. Tries to fix #728

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1254,6 +1254,7 @@ function init_worker(p)
 end
 
 function __init__()
+    ccall(:jl_generating_output, Cint, ()) == 1 && return nothing
     run_on_worker = get(ENV, "JULIA_REVISE_WORKER_ONLY", "0")
     if !(myid() == 1 || run_on_worker == "1")
         return nothing


timholy:
Yep, this might be necessary eventually, but let's wait for resolution of https://github.com/JuliaLang/julia/issues/48506 first.

vtjnash:
This seems best to me. It would be exceptionally disconcerting to me for a precompile process to detect and react to file system changes with some random probability depending on the performance of the filesystem. It might then attempt to call `eval`, causing the task to error because it was not in the right module, or to arbitrarily attempt to change the current module (again, causing the task to error or miscompile, since replacing methods is undefined behavior).

timholy:
Alright, I'll merge this and release a new version. My prediction, though, is that there are going to be more independent packages failing because of https://github.com/JuliaLang/julia/issues/48506. Merging this will indeed help us discover things that are Revise-independent.

vtjnash:
My remaining question–that is not related to https://github.com/JuliaLang/julia/issues/48506 but inspired by it–is the intended interpretation of this statement in the docs:

> julia> using Revise # you must do this before loading any revisable packages

at https://github.com/timholy/Revise.jl/blob/master/docs/src/cookbook.md#pkgtemplates

Is intended as a behavior that occurs from the arbitrary moment when Revise gets `__init__` called, or is intended as a literal instruction for the user to type at the REPL or startup.jl file? The former interpretation means that we could not include this as a default stdlib in the sysimg, since it is always "active" and cannot be avoided or turned off. This similarly also means that when loaded transitively (e.g. by Genie), it will watch a random subset of Genie's dependencies as well as Genie itself–whatever got loaded after the `__init__` ran, which may be different from run to run (depending on the order in which we iterated some hash tables).

But the later interpretation means there needs to be a created an action that implies a stateful user intent to enable it broadly, and to be a trigger that is not usually serialized (so that it has to be explicitly enabled for each current session). For example, I think this could check whether `Main.Revise === Revise`, but there could be other detection triggers too, such as detecting the REPL being defined after `Revise.__init__` ran.

Finally, one other alternative option is that Revise should instead start to watch every package, outside of the sysimg, regardless of the order in which Revise was loaded.

Do those questions and thoughts make sense?

timholy:
> Finally, one other alternative option is that Revise should instead start to watch every package, outside of the sysimg, regardless of the order in which Revise was loaded.

I think this is best, and probably achievable once we started caching source text in `.ji` files.

vtjnash:
Do we not do that caching already?

timholy:
Yes. I was unclear: "*was* probably achievable..."

PR #732, opened 2023-02-06
Fix several sources of invalidation

timholy:
These changes fix invalidation from the JuliaData ecosystem,
presumably mostly due to new AbstractString subtypes.


timholy:
This is stalled because it causes failures on the "fake language" tests. I was concerned at the time that these would make it harder to avoid invalidations.

timholy:
This will be solved by running Revise at fixed world.

PR #733, opened 2023-02-15
Simplify RNG used for making temp dirs

jakobnissen:
Use the documented `default_rng` instead of the internal GLOBAL_RNG.

Julia's testsets will automatically restore the state of the global RNG to improve tests' reproducibility. This is a problem for Revise, that needs actual random directories.
Previous code would mutate the global RNG to keep it from being restored. This code instead uses a local RNG object.

### DIFF
--- a/test/common.jl
+++ b/test/common.jl
@@ -1,19 +1,13 @@
 using Random
 using Base.Meta: isexpr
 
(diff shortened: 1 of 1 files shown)


timholy:
Thanks!

PR #743, opened 2023-04-01
Fix nightly

timholy:
Nightly has changed the internals of method-lookup, this uses the
world age rather than `typemax(UInt)`.


timholy:
Hah, already fixed in #740 (I forget so quickly...)

PR #751, opened 2023-07-12
make REPL AST transform work when `active_module` is set to `Core.Compiler`

aviatesk:
Otherwise REPL will result in the following error since `isempty` is resolved within `Core.Compiler` context module:
```julia
Core.Compiler) julia> 42
ERROR: MethodError: no method matching iterate(::Base.Set{Tuple{Revise.PkgData, String}})
You may have intended to import Base.iterate

Closest candidates are:
  iterate(::MethodSpecializations)
   @ Core reflection.jl:1146
  iterate(::MethodSpecializations, ::Nothing)
   @ Core reflection.jl:1152
  iterate(::MethodSpecializations, ::Int64)
   @ Core reflection.jl:1153
  ...

Stacktrace:
 [1] isempty(itr::Base.Set{Tuple{Revise.PkgData, String}})
   @ Core.Compiler ./essentials.jl:952
 [2] top-level scope
   @ none:1
```

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1231,7 +1231,7 @@ function revise_first(ex)
         isa(exu, Expr) && exu.head === :call && length(exu.args) == 1 && exu.args[1] === :exit && return ex
     end
     # Check for queued revisions, and if so call `revise` first before executing the expression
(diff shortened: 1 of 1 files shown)


timholy:
Sorry, forgot about it! Thanks!

PR #757, opened 2023-08-23
Correct default use of `skip_include` in `includet` docstring

axsk:
The default behaviour is to skip recursively included files for tracking. This is now correctly mirrored in the doctring

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -985,7 +985,7 @@ This can be automated using [`entr`](@ref).
 
 `includet` is essentially shorthand for
 
-    Revise.track(Main, filename; mode=:eval, skip_include=true)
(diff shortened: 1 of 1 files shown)


src/packagedef.jl
```diff
@@ -985,7 +985,7 @@ This can be automated using [`entr`](@ref).
 
 `includet` is essentially shorthand for
 
-    Revise.track(Main, filename; mode=:eval, skip_include=false)
+    Revise.track(Main, filename; mode=:eval, skip_include=true)
```
timholy, line 988:
```suggestion
    Revise.track(Main, filename; mode=:includet, skip_include=true)
```

timholy (commented):
Nice catch!

PR #764, opened 2023-09-27
Update to Documenter v1

carstenbauer:
This gives many nice convenience features like, e.g., a better search, a GitHub link to the repo at the top of the docs, collapsable docstrings and many more. I only had to fix a few mistakes in the docs (which under v1 lead to errors instead of warnings), like a few dead links etc.

Tested locally, let's see what CI says.


carstenbauer:
Failing CI (Julia nightly) seems unrelated to me.

pfitzseb (approved):

PR #769, opened 2023-10-25
allow CI nightly to fail so badge is green

Moelf:
There are two ways of doing this, another possibility is to have a separate `ci_nightly.yml`, let me know if maintainers prefer that route.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,13 +15,17 @@ jobs:
         version:
           - '1.6'     # LTS
           - '1'       # current stable
-          - 'nightly'
         os:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
         arch:
           - x64
+        include:
+          - version: 'nightly'
+            os: ubuntu-latest
+            arch: x64
+            allow_failure: true
     steps:
     - uses: actions/checkout@v4
     - uses: julia-actions/setup-julia@latest


timholy:
The issue is that this failure is going to be on all Julia versions starting with 1.10, so making an exception for nightly soon won't help anything. We just have to fix the test failures.

Moelf:
but this would be useful for future future? I think it's a general idea to not show red CI badge for Nightly

timholy:
Fair enough!

timholy:
Hmm, the badge is still red even though only nightly is failing.

Moelf:
I wonder if this is correct https://github.com/timholy/Revise.jl/blob/9f7bd4ea3d7b2c3bd37e64130abb718aba2a8072/README.md?plain=1#L3C49-L3C49

timholy:
If I try updating it with
```
[![CI](https://github.com/timholy/Revise.jl/actions/workflows/ci.yml/badge.svg)](https://github.com/timholy/Revise.jl/actions/workflows/ci.yml)
```
(the text supplied by the "Create Status Badge" option, it still shows red.

Moelf:
ah, we need this line https://github.com/JuliaHEP/UnROOT.jl/blob/77b75d8f8a7d5a6a2b8c408efbcac2c00817e798/.github/workflows/ci.yml#L14C1-L14C51

PR #771, opened 2023-10-30
Update cookbook.md

nathanrboyer:
Need to `dev MyPkg` before using it or it will error.

There was confusion in [this Discourse post](https://discourse.julialang.org/t/when-should-i-activate/105581) following the "Cookbook". I tried to add a simple fix.

### DIFF
--- a/docs/src/cookbook.md
+++ b/docs/src/cookbook.md
@@ -70,7 +70,16 @@ Generating project MyPkg:
 In the first few lines you can see the location of your new package, here
 the directory `/home/tim/.julia/dev/MyPkg`.
 
-Before doing anything else, let's try it out:
+Press `]` to enter the [Pkg REPL](https://pkgdocs.julialang.org/v1/getting-started/#Basic-Usage).
+Then add the new package to your current environment with the `dev` command.
+
+```julia
+(<environment>) pkg> dev /home/tim/.julia/dev/MyPkg
+```
+
+Press the backspace key to return to the Julia REPL.
+
+Now let's try it out:
 
 ```julia
 julia> using Revise   # you must do this before loading any revisable packages


docs/src/cookbook.md
```diff
@@ -70,7 +70,16 @@ Generating project MyPkg:
 In the first few lines you can see the location of your new package, here
 the directory `/home/tim/.julia/dev/MyPkg`.
 
-Before doing anything else, let's try it out:
+Press `]` to enter the [Pkg REPL](https://pkgdocs.julialang.org/v1/getting-started/#Basic-Usage).
+Then add the new package to your current environment with the `dev` command.
+
+```julia
+(<environment>) pkg> dev /home/tim/.julia/dev/MyPkg
```
timholy, line 77: Does it work with just `pkg> dev MyPkg`? If so, it would be nice to keep `tim` out of the commands people actually type, so it's copy/pastable.

timholy (commented):
Thanks! One small suggestion.

docs/src/cookbook.md
nathanrboyer: I think that would work in this context since the REPL will be in the correct directory from having just made the package. However, I think it is important to show that this is a path, which behaves differently from naming a registered package. That path with "tim" is already shown a few places above, but I could change it to `./MyPkg`.

nathanrboyer: Actually, I just tested it and `dev MyPkg` does work because somehow Julia knows to look in the `~/.julia/dev/` directory that PkgTemplates uses. (It does not create the package in the current directory like I thought.)

nathanrboyer:
I just tested the PkgTemplates and Pkg sections. Both work now. The PkgTemplates MyPkg.jl file does not come with the `greet` function, but the `Pkg.generate` MyPkg.jl file does.

nathanrboyer:
Bump

pfitzseb (approved):

Issue #735, opened 2023-02-23
Revising `"docstring" f(::A{T}) where T` fails with semi-helpful error message


mgkurtz:
I can write code like

```julia
module Foo
struct A{T} end
"""Magic function"""
f(::A{T}) where T
end
```

This is valid Julia code, but Revise cannot revise it and gives a not-so-helpful error message.

Steps to reproduce: write the above code in `/path/to/file/Foo.jl`. Start the Julia REPL and type

```julia
`push!(LOAD_PATH, "/path/to/file")`
using Foo
```

Then add a line like `x=42` to Foo.jl and try to access it in the REPL. You will get

    ┌ Error: Failed to revise /path/to/file/Foo.jl
    │   exception = lowering returned an error, :($(Expr(:error, "invalid \"::\" syntax")))
    └ @ Revise ~/.julia/packages/Revise/P8ITN/src/packagedef.jl:717

This is semi-helpfull. Working on a large file written by others, I had some trouble pin-pointing this error. Changing `methods_by_execution!` to also print the statement producing the error, I got

    exception = lowering returned an error, (:($(Expr(:error, "invalid \"::\" syntax"))), :(f(::A{T}) where T))

Now, we see, the error arises when `methods_by_execution!` calls `Meta.lower(Foo, :(f(::A{T} where T)))`. Without context `f(::A{T}) where T` evaluates to `error("invalid \"::\" syntax")`. Hence appropriately `Meta.lower` gives `$(Expr(:error, "invalid \"::\" syntax"))`.

For some reason, the error occurs, when the argument to `f` is parametric but not, when I just document `f(::Int)`. Although `f(::Int)` evaluates to the same “invalid "::" syntax” error as `f(::A{T}) where T` does. Curious.

Since working around that bug is rather easy (add an `x` in front of `::A{T}`), I would be happy enough, if I we could provide some useful error message.

fredrikekre:
Just ran into this, thanks for the workaround.

I made a PR to hack around this issue, see #773.

timholy:
Nice detective work @mgkurtz and thanks for #774 @fredrikekre!

PR #773, opened 2023-11-09
Workaround lowering error in docstring method signatures

fredrikekre:
This patch adds a (minimal?) targeted hack to workaround an issue where a method signature for documentation fails lowering (#735).

Specifically, when revising the following code:

```julia
struct Foo{T} end

"docs"
Foo{T}(::Int) where T
```

lowering errors for the expression `:(Foo{T}(::Int) where T)` with `"invalid :: syntax"`. This patch simply ignores the expression when this situation is detected. Note that, just after, the full expression (docstring + signature) is lowered together succesfully. In particular this means that the docstring will still be updated correctly.


fredrikekre:
Okay, I found the root cause, and this is not it.

PR #774, opened 2023-11-09
Filter out more doc signature expressions

### LINKED ISSUE #735: Revising `"docstring" f(::A{T}) where T` fails with semi-helpful error message

fredrikekre:
Revise extracts the documented expression and puts that before the full
doc expression in the revision queue. E.g. in `"docs" f(x) = x` there
will be a revision of `f(x) = x` followed by a revision of the full
expresion (`"docs" f(x) = x`). However, in many cases a docstring is
only attached to a signature (not a function body/struct definition,
etc.) and in this case the revision of the signature is not needed.

Revise already filters out the base case, `"docs" f(x)`. This patch
extends this filtering to also apply to `where`-clauses such as e.g.
`"docs" f(x::T) where T <: Integer`.

Concretely, this patch fixes #735, i.e. revising doc expressions where
the signature doesn't lower without the context of the doc macro. As a
bonus, slightly less work has to be done for e.g. `"docs" f(x::T) where
T` since this is also filtered out.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Revise"
 uuid = "295af30f-e4ad-537b-8983-00126c2a3abe"
-version = "3.5.7"
+version = "3.5.8"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -90,11 +90,20 @@ istrivial(a) = a === nothing || isa(a, LineNumberNode)
 
 isgoto(stmt) = isa(stmt, Core.GotoNode) | isexpr(stmt, :gotoifnot)
 
+function unwrap_where(ex::Expr)
+    while isexpr(ex, :where)
+        ex = ex.args[1]
+    end
+    return ex
+end
+
 function pushex!(exsigs::ExprsSigs, ex::Expr)
     uex = unwrap(ex)
     if is_doc_expr(uex)
         body = uex.args[4]
-        if isa(body, Expr) && body.head !== :call   # don't trigger for docexprs like `"docstr" f(x::Int)`
+        # Don't trigger for exprs where the documented expression is just a signature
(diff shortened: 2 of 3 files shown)


fredrikekre:
CI failures are unrelated -- looks like tests need a bit of work to pass on nightly (probably the result of stdlib excision).

Issue #778, opened 2023-11-27
Solved: Revise does not work on linux 


j-fu:
I quite often encountered a situation that Revise.jl somehow ceased to work. No errors messages, nothing besides revise not working - see e.g. https://github.com/fonsp/Pluto.jl/issues/2094 .
Testing Revise appears to reliably detect this situation.

Tests look like
```
File paths: Test Failed at /home/fuhrmann/.julia/packages/Revise/xRltd/test/runtests.jl:511
Expression: pcwf1() == -1
Evaluated: 1 == -1

Stacktrace:
[1] top-level scope
@ ~/local/julia/julia-1.9.4/share/julia/stdlib/v1.9/Test/src/Test.jl:478
[2] eval(m::Module, e::Any)
@ Core ./boot.jl:370
[3] macro expansion
@ ~/.julia/packages/Revise/xRltd/test/runtests.jl:511 [inlined]
[4] macro expansion
@ ~/local/julia/julia-1.9.4/share/julia/stdlib/v1.9/Test/src/Test.jl:1498 [inlined]
[5] macro expansion
@ ~/.julia/packages/Revise/xRltd/test/runtests.jl:412 [inlined]
[6] macro expansion
@ ~/local/julia/julia-1.9.4/share/julia/stdlib/v1.9/Test/src/Test.jl:1498 [inlined]
[7] top-level scope
@ ~/.julia/packages/Revise/xRltd/test/runtests.jl:72
File paths: Test Failed at /home/fuhrmann/.julia/packages/Revise/xRltd/test/runtests.jl:518
Expression: using_macro_pcwf() == -6
Evaluated: 6 == -6
```

In another round of digging I learned that Revise is tracking files using file system tools and  about the "inotify limits". I found a quite good explanation [here](https://github.com/fonsp/Pluto.jl/issues/2094). On my system (openSUSE Leap 15.5) these limits are by default:
```
$ sysctl fs.inotify
fs.inotify.max_queued_events = 16384
fs.inotify.max_user_instances = 128
fs.inotify.max_user_watches = 524288
```
Increasing  `fs.inotify.max_user_instances` immediatly fixes the problem.
E.g. after setting
```
sudo sysctl fs.inotify.max_user_instances=2048
```
Revise unit tests immediately start to work. This also can be made [permanent](https://www.suse.com/de-de/support/kb/doc/?id=000020048).

I could create a documention PR  on this.

timholy:
Fixed by #779

PR #779, opened 2023-11-27
Update band-aid info on file watching (#778)

j-fu:
At once updated link to Linux resource.

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -123,25 +123,37 @@ string `"1"` (e.g., `JULIA_REVISE_INCLUDE=1` in a bash script).
 Revise needs to be notified by your filesystem about changes to your code,
 which means that the files that define your modules need to be watched for updates.
 Some systems impose limits on the number of files and directories that can be
-watched simultaneously; if this limit is hit, on Linux this can result in a fairly cryptic
-error like
+watched simultaneously; if such a limit is hit, on Linux this can result in Revise silently ceasing to work 
+(albeit with unit tests failing) or in a fairly cryptic error like
 
 ```sh
 ERROR: start_watching (File Monitor): no space left on device (ENOSPC)
 ```
 
-The cure is to increase the number of files that can be watched, by executing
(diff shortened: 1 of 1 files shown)


docs/src/config.md
```diff
@@ -123,25 +123,37 @@ string `"1"` (e.g., `JULIA_REVISE_INCLUDE=1` in a bash script).
...
 
-The cure is to increase the number of files that can be watched, by executing
+The cure is to investigate and possibly increase the number of files that can be watched.
 
+Invoking
 ```sh
-echo 65536 | sudo tee -a /proc/sys/fs/inotify/max_user_watches
+$ sysctl fs.inotify
 ```
+at the linux prompt may e.g. result im
```
timholy, line 139:
```suggestion
at the linux prompt may e.g. result in
```

docs/src/config.md
```diff
@@ -123,25 +123,37 @@ string `"1"` (e.g., `JULIA_REVISE_INCLUDE=1` in a bash script).
...
+```
+fs.inotify.max_queued_events = 16384
+fs.inotify.max_user_instances = 128
+fs.inotify.max_user_watches = 524288
+```
+
+In this case it may make sense to increase one of the limits, e.g.
```
timholy, line 146:
```suggestion
For Revise usage, `max_user_watches >= 65536` is recommended, and more can be helpful; the value of 524288 above is common on modern systems. One can set higher values as needed, e.g.,
```

docs/src/config.md
```diff
@@ -123,25 +123,37 @@ string `"1"` (e.g., `JULIA_REVISE_INCLUDE=1` in a bash script).
...
+```
+$ sudo sysctl fs.inotify.max_user_instances=2048
+```
+and to run unit tests again.
```
timholy, line 150:
```suggestion
After changing these values, it is advised to run Revise's unit tests to see if they pass.
```

timholy (commented):
Thanks, this is terrific.

Issue #734, opened 2023-02-21
Revise not working anymore for files that contain type definition (but the type is not changed)


thofma:
I noticed a regression with files that contain type definitions.

I have a file, which contains a type definition
```
mutable struct A
  x
end
```
and some method definitions
```
function g()
end
```
If the file is tracked with Revise and I change `g`, I get:
```
┌ Error: Failed to revise ...
│   exception =
│    invalid redefinition of constant A
```
which I did not expect, since I did *not* change the type definition. This is with Revise version 3.5.1.

I quickly tried with Revise version 3.4.0 and it is working as expected. This means, that when only `g` is changed but not `A`, then the file is revised properly.

lgoettgens:
The same thing happened to me. However, it only occurs in a subset of the files of the given structure, and I cannot find any way to distinguish them.
Pinning the version to 3.4.0 seems to be a workaround for now, thanks for the tip @thofma .

fingolfin:
@timholy any idea what might be causing this?

timholy:
I need to spend a week or so on Revise. No time right now though.

kahaaga:
Same issue here with Revise v3.5.10. 

I'm tracking changes in a package I'm developing. Any change to any file that contains a type definition triggers the same error as above. 

The following workaround solves it for me (thanks, @fingolfin):

```julia
(@v1.9) pkg> activate . # in whatever project folder you're working
(MyProject) pkg> add JuliaInterpreter@0.9.26
(MyProject) pkg> add Revise@3.4.0
julia> using Revise, MyProject
```

timholy:
At least with Julia 1.10 and LoweredCodeUtils 2.4.0, I cannot replicate this. Here's my setup:
```julia
tim@diva:/tmp/TestRevise$ ls -l
total 12
-rw-r--r-- 1 tim tim  186 Dec 30 10:36 Manifest.toml
-rw-r--r-- 1 tim tim  128 Dec 30 10:35 Project.toml
drwxr-xr-x 2 tim tim 4096 Dec 30 10:35 src
tim@diva:/tmp/TestRevise$ cat src/TestRevise.jl
module TestRevise

export g

mutable struct A
    x
end

function g()
end

end # module TestRevise
```
And then
```julia
julia> using TestRevise
Precompiling TestRevise
  1 dependency successfully precompiled in 1 seconds

julia> g()

# Now edit `g` to return 1
julia> 1+1
2

julia> g()
1
```

[...]

PR #781, opened 2023-11-28
minor updates on old code

aviatesk:
Since the minimum version compat for Revise is now 1.6, we can expect `GotoIfNot` to be defined within `Core` always.


aviatesk:
You need to update the dependencies too.

rben01:
@aviatesk I ran `]update`; no change.

aviatesk:
What happens if you do `pkg> add LoweredCodeUtils@2.3.1`?

fingolfin:
I also get that error; however with Revise 3.4.0, which I pinned as all newer versions seem to suffer from issue #734

PR #784, opened 2023-12-31
Fix tests that were broken by 1.10

timholy:
In particular, the new parser generates different (and much better) error messages; we were testing the content of the error message, so that has to change.

Also skips a portion of the old signature tests due to usage of `include(mapexpr, file)`, which is not (and may never be) supported by Revise.


timholy:
Except nightly, all the core tests are passing, it seems to be an unfortunate level of flakiness in some of the auxiliary CI-only tests.

Issue #797, opened 2024-01-08
Function not getting recompiled after changing struct definition


A-C1:
Hi, not sure if this is a bug or I am using Revise incorrectly. I am following the example [here](https://timholy.github.io/Revise.jl/stable/limitations/) to redefine struct (by renaming it).  The package `MyPkg` was created using
`]generate MyPkg ` and then deving it by `dev ./MyPkg`.

I think the issue is if the function  `MyPkg.processFoo(b)` is called with `FooStruct1` and then `FooStruct1` is changed to `FooStruct2` and then the function is called with `FooStruct2` a new method does not get compiled. I am attaching the complete REPL session
```julia
julia> 
PS C:\Users\adity\Desktop\MyPkg> julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.10.0 (2023-12-25)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using Revise

julia> using MyPkg
Precompiling MyPkg
  1 dependency successfully precompiled in 3 seconds

julia> a = MyPkg.FooStruct(1)
MyPkg.FooStruct1(1)

julia> MyPkg.processFoo(a)
[ Info: 1
Hello Mitali

julia> b = MyPkg.FooStruct(1,1)
┌ Error: Failed to revise C:\Users\adity\Desktop\MyPkg\src\MyPkg.jl
│   exception =
│    invalid redefinition of constant MyPkg.FooStruct1
│    Stacktrace:
│     [1] top-level scope
│       @ C:\Users\adity\Desktop\MyPkg\src\MyPkg.jl:3
│    Revise evaluation error at C:\Users\adity\Desktop\MyPkg\src\MyPkg.jl:3
│ 
└ @ Revise C:\Users\adity\.julia\packages\Revise\FaTes\src\packagedef.jl:722
┌ Warning: The running code does not match the saved version for the following files:
│ 
│   C:\Users\adity\Desktop\MyPkg\src\MyPkg.jl
│ 
│ If the error was due to evaluation order, it can sometimes be resolved by calling `Revise.retry()`.
│ Use Revise.errors() to report errors again. Only the first error in each file is shown.
[...]

PR #798, opened 2024-01-08
Update limitations.md with REPL session

mkitti:
Fix https://github.com/timholy/Revise.jl/issues/797

See Discourse thread:
https://discourse.julialang.org/t/revise-not-updating-functions/108391/12

Basically we replace the initial struct changing session with the following.
1. Use an abstract parent type for all struct versions and method definitions
2. Provide a full blown REPL session and example showing all steps
3. Explicitly use `write` to demonstrate editing steps.
4. Follow through with a Julia session restart
5. Demonstrate `isconst` on `FooStruct`.

### DIFF
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -77,6 +77,8 @@ julia> processFoo(FooStruct(3.5))
 [ Info: 3.5
 ```
 
+Here, note that we made two changes: we updated the "version number" of FooStruct when we changed something about its fields, and we also re-assigned FooStruct to alias the new version. We did not change the definition of any methods that have been typed AbstractFooStruct.
+
 This works as long as the new type name doesn't conflict with an existing name; within a session you need to change the name each time you change the definition.
 
 Once your development has converged on a solution, it's best to switch to the "permanent" name: in the example above, `FooStruct` is a non-constant global variable, and if used internally in a function there will be consequent performance penalties. Switching to the permanent name will force you to restart your session.
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -11,30 +11,113 @@ These kinds of changes require that you restart your Julia session.
 During early stages of development, it's quite common to want to change type definitions. You can work around Julia's/Revise's limitations by temporary renaming:
 
 ```julia
-# 1st version
-struct FooStruct1
-    bar::Int
-end
-FooStruct = FooStruct1
-function processFoo(foo::FooStruct)
-    @info foo.bar
-end
+julia> using Pkg, Revise
+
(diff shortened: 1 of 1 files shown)


docs/src/limitations.md
```diff
@@ -11,31 +11,114 @@ These kinds of changes require that you restart your Julia session.
 During early stages of development, it's quite common to want to change type definitions. You can work around Julia's/Revise's limitations by temporary renaming:
```
timholy, line 11:
```suggestion
During early stages of development, it's quite common to want to change type definitions. You can work around Julia's/Revise's limitations by temporary renaming. We'll illustrate this below, using `write` to be explicit about when updates to the file happen. But in ordinary usage, these are changes you'd likely make with your editor.
```

docs/src/limitations.md
```diff
@@ -11,31 +11,114 @@ These kinds of changes require that you restart your Julia session.
...
+       end
+       """)
+234
+
+julia> FooStruct # make sure FooStruct refers to FooStruct2
+MyPkg.FooStruct2
+
+julia> processFoo(FooStruct(3.5))
+[ Info: 3.5
 ```
```
timholy, line 78:
```suggestion
```
Here, note that we made two changes: we updated the "version number" of `FooStruct` when we changed something about its fields, and we also re-assigned `FooStruct` to alias the new version. We did not change the definition of any methods that have been typed `AbstractFooStruct`.
```

**Review by timholy — COMMENTED** [review]
Thanks, this is a very big improvement!

File: docs/src/limitations.md [hunk 3]
```diff
@@ -8,34 +8,117 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+
```
mkitti, line 79:
```suggestion

Here, note that we made two changes: we updated the "version number" of FooStruct when we changed something about its fields, and we also re-assigned FooStruct to alias the new version. We did not change the definition of any methods that have been typed AbstractFooStruct.

```

docs/src/limitations.md
```diff
@@ -8,34 +8,119 @@ There are some kinds of changes that Revise (or often, Julia itself) cannot inco
...
+[ Info: 1
+
+julia> write("src/MyPkg.jl","""
+       module MyPkg
+
+       export FooStruct, processFoo
+
+       abstract type AbstractFooStruct end
+       struct FooStruct2 <: AbstractFooStruct # change version nuumber
+           bar::Float64 # changed type of the field
```
mkitti, line 62:
```suggestion
           bar::Float64 # change type of the field
```

PR #799, opened 2024-01-11
Add continue-on-error to CI

timholy:
Suggestion from https://github.com/timholy/Revise.jl/pull/769#issuecomment-1886030780


timholy:
Hmm, that didn't work... @Moelf ?

Moelf:
for the normal ones you need `allow_failure: [false]`

- https://github.com/JuliaHEP/UnROOT.jl/blob/77b75d8f8a7d5a6a2b8c408efbcac2c00817e798/.github/workflows/ci.yml#L25C9-L25C31

Moelf:
image


seems to work?

PR #806, opened 2024-02-20
Workaround issue when internal code loading function gives a path to entry file instead of directory to package

KristofferC:
Fixes https://github.com/timholy/Revise.jl/issues/802


felixcremer:
I tested it on an environment where I ran into #802 almost always when I touched the package manager and with this PR  I don't get the warnings anymore and also when I test the methods(mean) the fallback version is still there.

KristofferC:
Good enough for me, thanks! I'll release a new Revise then and hopefully that's it.

timholy:
Thanks @KristofferC !

PR #822, opened 2024-06-20
Track `GlobalRef` consistently

Keno:
Companion PR to https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/107. With both of these and https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/634, basic Revise functionality is working for me again on Julia master.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
           - '1.6'     # LTS
           - '1.9'     # parsing errors branch on 1.10, so test the last pre-1.10 version
           - '1'       # current stable
+          - 'pre'     # next release, if available
         os:
           - ubuntu-latest
           - macOS-latest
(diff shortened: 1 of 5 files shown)


timholy:
For me, here's the current status (with both JuliaInterpreter and LoweredCodeUtils `dev`ved on their current `master` branch):
`nightly`:
image

But more concerningly, on Julia 1.11:
image

Is this working for you?

timholy:
OK, while there were many errors there didn't seem to be many causes.

timholy:
Thanks Shuhei! Are you good with me tagging LoweredCodeUtils 3.0.0 and then finishing this off? See my final comment in https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/99.

timholy:
Thanks again, @Keno!

PR #823, opened 2024-06-30
Fix precompilation warning on Julia 1.11

timholy:


timholy:
Better fix is https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/108

PR #824, opened 2024-07-01
Fix tests broken on 1.11

timholy:
Almost all the real work of updating to 1.11 was in https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/108 and https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/109. But a warning now prints with different text, so accept both variants.


fredrikekre (approved):

timholy:
I'm making a new release just so PkgEval is happy

PR #827, opened 2024-07-25
More 1.12 compat

Keno:
With this and the LCU fixes, Revise should pass tests again on 1.12.

### DIFF
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -1,23 +1,31 @@
+const badfile = (nothing, nothing, nothing, UInt128(0))
 function pkg_fileinfo(id::PkgId)
     origin = get(Base.pkgorigins, id, nothing)
-    origin === nothing && return nothing, nothing, nothing
+    origin === nothing && return badfile
     cachepath = origin.cachepath
-    cachepath === nothing && return nothing, nothing, nothing
+    cachepath === nothing && return badfile
+    local checksum
     provides, includes_requires, required_modules = try
-        @static if VERSION ≥ v"1.11.0-DEV.683" # https://github.com/JuliaLang/julia/pull/49866
+        ret = @static if VERSION ≥ v"1.11.0-DEV.683" # https://github.com/JuliaLang/julia/pull/49866
+            io = open(cachepath, "r")
+            checksum = Base.isvalid_cache_header(io)
+            iszero(checksum) && (close(io); return badfile)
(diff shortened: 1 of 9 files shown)


src/loading.jl
```diff
@@ -29,13 +36,17 @@ function parse_pkg_files(id::PkgId)
...
     if use_compiled_modules()
-        cachefile, includes, reqs = pkg_fileinfo(id)
+        cachefile, includes, reqs, buildid = pkg_fileinfo(id)
         if cachefile !== nothing
             @assert includes !== nothing
             @assert reqs !== nothing
             pkgdata.requirements = reqs
             for chi in includes
-                mod = Base.root_module(id)
+                if VERSION >= v"1.12.0-DEV.764" && haskey(Base.loaded_precompiles, id => buildid)
```
Keno, line 45: @vtjnash is this the expected fix for this?

vtjnash: That probably depends upon what the code is attempting to do. It is supposed to be buildid specific or not?

Keno: This code is adding the files that a .ji file depends on to Revise's watch list, so I believe it is buildid specific.

vtjnash: Ensuring that it uses the same .ji file is probably correct then. We might need to examine whether the package hooks should provide that additional information also, to ensure Revise is accurate about it though?

Keno: Alright, good enough for now then.

Keno:
Updated to bump LCU and cherry-pick the additional fixes from #829

KristofferC:
Interesting that the linux testers don't know about LCU 3.0.1.

Keno:
Probably just hadn't propagated yet. Rerunning.

Keno:
@timholy The last issue here is that `ExponentialUtilities` `v1.9.0` is not compatible with all the julia versions we want to test. Can we switch this test to be more self contained and include some minimal in the test code itself rather than depending on some external package?

Keno:
That's mostly a separate change though. I would propose merging this first and then addressing that last issue.

timholy:
> Can we switch this test to be more self contained and include some minimal in the test code itself rather than depending on some external package?

Yes, that needs to happen. I have been doing that for newer tests, but I haven't migrated old tests. Doesn't need to stand in the way of this PR.

timholy (approved):

PR #829, opened 2024-07-26
Handle change in `Base.active_repl_backend` on 1.12

IanButterworth:
Handles changes to `Base.active_repl_backend` in https://github.com/JuliaLang/julia/pull/55210


Keno:
I also have a version of this in #827, but looks like we both caught some additional cases that the other didn't, so I guess we need both.

IanButterworth:
Feel free to cherry-pick this commit over to yours

PR #831, opened 2024-07-28
Decouple version switching test from ExpontentialUtils

Keno:
And also fix a bug when revising from a package that was previously empty, where it would assume it hadn't been parsed yet because `modexsigs` was empty.


Keno:
@timholy Should be good to go if you want to have a quick look.

test/switch_version.jl
```diff
@@ -0,0 +1,32 @@
+# This test is intended to be run after `setup.sh` runs. ExponentialUtilities
+# should be held at v1.9.0.
+
```
timholy, line 3:
```suggestion
```

timholy (approved):
Wow, thanks! Really nice work!

PR #841, opened 2024-08-18
recheck file exists on deletions to deal with race condition with editor saving

acertain:
This is somewhat hacky & only mostly works for me, it's still possible to trigger a failure with entr + repeated saving. I guess it'd be better to keep tracking deleted files in case they come back?

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -682,11 +682,11 @@ function handle_deletions(pkgdata, file)
             filecontents = try_read(filename)
         end
     end
-    mexsnew = filecontents ? parse_source(filep, topmod, filecontents) : ModuleExprsSigs(topmod)
+    mexsnew = (filecontents !== nothing) ? parse_source(filep, topmod, filecontents) : ModuleExprsSigs(topmod)
     if mexsnew !== nothing
         delete_missing!(mexsold, mexsnew)
     end
-    if !filecontents
+    if filecontents === nothing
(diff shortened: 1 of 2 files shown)


timholy:
Thanks for pointing this out. I assume the race condition is the file disappearing between the `isfile` check and reading the contents.

I don't use vim, so pardon the ignorant questions.

- I imagine this wouldn't be a problem if vim (1) saved the new contents to a temporary filename, and then (2) renamed the new file to the real name. Is there a setting in vim one could configure to make it use that behavior instead? If not, would it be worth filing a vim issue and ask if they could switch the way they save files? I'm asking because I don't see any other options that are particularly palatable.
- emacs leaves `*~` files in place. So if someone uses both vim and emacs, won't `bak_exists` get confused?

If this is only triggered by `entr`, should we handle this there instead? E.g., does putting a `sleep(0.01)` between https://github.com/timholy/Revise.jl/blob/d368547f7f659e0937297576ed3a8612ab550a4e/src/callbacks.jl#L164-L165 make the problem go away? That's also hacky and unreliable, but short of using lock files (which seem OS-dependent, and may not work on NFS) I don't see a great solution.

timholy:
For anyone who finds this, the current recommendation is to set `backupcopy=yes` in your vim settings. This might become the default, but that's unclear. xref https://github.com/vim/vim/issues/15733

PR #844, opened 2024-09-11
Capture log message instead of telling users to ignore it

giordano:
This makes reading test logs slightly less noisy, scary and confusing.


giordano:
Bump

Issue #837, opened 2024-08-09
Possibility of lost events due to use of edge-triggered Condition in entr(...)


frankier:
@JanisErdmanis caught a possible race-condition in `Revise.entr(...)`

In particular in these lines:

https://github.com/timholy/Revise.jl/blob/13a5eb7986ee1239ff938a92043c13fee04579cc/src/callbacks.jl#L163-L166

It looks like any event fired while `Revise.revise()` is running will be lost. I guess something level-triggered like https://docs.julialang.org/en/v1/base/parallel/#Base.Event could be used. This would assume that there is only one listener, which is probably reasonable.

frankier:
I believe this is still a problem even though the file watching tasks this revision_event watching loop should typically end up running in the same thread (the file watching tasks are sticky). The reason is is that the Julia interpreter can choose to yield at any time and particularly it may yield during IO performed in `Revise.revise(...)`.

timholy:
I'm not sure there's a great solution for this short of file-locking. xref #841

frankier:
I'm not sure if locking one of the places the event is fired with the revision_lock solves the issue. The problem is that for `Condition`, `notify(...)` will only wake up all blocking tasks that *are blocking at that moment in time*.

Imagine we have two tasks, and task 1 is running `entr(...)`, then:

```

          ----------------------------------------------------------------------------------------------
Task 1    |  wait(revision_event)  |      Revise.revise()     | wait(revision_event)
          ----------------------------------------------------------------------------------------------

Task 2                             |                       |
                             notify(revision_event)  notify(revision_event)

                                                           ^ Event is lost  ^ Waiting despite changes
```

Another issue is that it appears `Base.Condition` is being used which is not threadsafe, apparently https://docs.julialang.org/en/v1/base/parallel/#Base.Threads.Condition should be used in this case.

The solution is to use something level triggered. https://docs.julialang.org/en/v1/base/parallel/#Base.Event together with autoreset seems like it should work, although requires Julia 1.8. Note that in this case we need to have only a single "listener", which should probably be fine I guess? Or are there situations where there need to be multiple listeners? In the latter case, I guess some kind of pub/sub pattern would be needed where each listener adds its own autoresetting event, and then notify notifies all of them.

timholy:
Ah, I assumed it was because files were being added to `revision_queue` in the middle of `revise` running (which would also be a way for this to happen). So #846 should help, but you're right this doesn't completely solve the problem.

[...]

PR #846, opened 2024-09-24
Add locking to `revise()`

### LINKED ISSUE #845: Revise.revise(...) is not safe to be called from multiple threads simultaneously
`Revise.revise(...)` appears to not be threadsafe. Typically it is only run from a single thread, and only ever one-at-a-time, e.g. in the REPL. However, it is possible it could be called from multiple threads (e.g. in a webserver). In this case, the reading and writing of the global `revision_queue` without locks could cause inconsistent/corrupt reads (build-in data structures are not threadsafe). As a secondary problem, there is a thundering herd issue where work could be duplicated between Revise passes started at the same time.

timholy:
Fixes #845

@frankier, can you test whether this works for you? I'm guessing it's not straightforward to reliably test this.

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -85,6 +85,9 @@ Global variable, a set of `Manifest.toml` files from the active projects used du
 """
 const watched_manifests = Set{String}()
 
+# Locks access to `revision_queue` to prevent race conditions, see issues #837 and #845
+const revise_lock = ReentrantLock()
+
 """
     Revise.revision_queue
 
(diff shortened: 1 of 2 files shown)


frankier:
I looks good to me. Locking everything (apart from sleep) definitely seems the safest approach.

I don't have any test code for this, but it seems correct. Another change related for passing through keyword arguments has snuck into the 2nd commit.

timholy:
I found more places to add locking, now also fixes #837

PR #847, opened 2024-09-24
Add `force=true` kwarg to `revise(mod::Module)`

### LINKED ISSUE #839: Tools for when `FileWatching` doesn't work properly
I do a fair amount of development work on my MacBook Pro with a linux VM that has a shared folder setup so that my native VSCode edits files on my macOS partition, but the Linux guest VM can execute the code.  While this works great overall, Revise doesn't work at all because the networked filesystem doesn't propagate file change notifications to `FileWatching` properly.

To work around this, I ended up cooking up the following recipe:
```julia
function rerevise(mod::Module)
    pkgdata = Revise.pkgdatas[Revise.PkgId(mod)]
    for file in pkgdata.info.files
        push!(Revise.revision_queue, (pkgdata, file))
    end
    Revise.revise()
end
```

Perhaps we can have a `revise(mod::Module; force::Bool = false)` method that does something similar to this, forcing re-evaluation of everything, regardless of what `FileWatching` says?

### LINKED ISSUE #842: Add force_revise function to handle delayed updates on NFS when using JULIA_REVISE_POLL=1
Revise does not always update package files on NFS when setting JULIA_REVISE_POLL=1, in particular sometimes it takes a few minutes for the changes to take effect automatically.

I have been using a workaround for a while and think it would be a great addition to the package, for systems where julias `poll_file` works poorly. This is the workaround:

```
function force_revise(mod)
    id = Revise.PkgId(mod)
    pkgdata = Revise.parse_pkg_files(id)
    files = Revise.srcfiles(pkgdata)
    for file in files
        file = joinpath(Revise.basedir(pkgdata), file)
        push!(Revise.revision_queue, (pkgdata, relpath(file, pkgdata)))
    end
end
``` 
If you think it makes sense I can add it in a PR.

timholy:
Fixes #839
Fixes #842

@staticfloat and/or @danielalcalde can you test whether this works for you? The behavior I think you're wanting is now `revise(mod; force=false)`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Revise"
 uuid = "295af30f-e4ad-537b-8983-00126c2a3abe"
-version = "3.5.18"
+version = "3.6.0"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -849,15 +849,24 @@ end
 revise(backend::REPL.REPLBackend) = revise()
 
 """
-    revise(mod::Module)
+    revise(mod::Module; force::Bool=true)
 
-Reevaluate every definition in `mod`, whether it was changed or not. This is useful
+Revise all files that define `mod`.
+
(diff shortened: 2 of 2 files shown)


danielalcalde:
thank you, seems to work fine :)

PR #853, opened 2024-10-03
Stop using `eval` to find module location

Keno:
I added an explicit accessor for this in
https://github.com/JuliaLang/julia/pull/55963
so Revise doesn't have to rely on internals.


aviatesk (approved):

timholy:
Thanks!

PR #854, opened 2024-10-03
Prepare for upcoming change to `include`

Keno:
Prepares for https://github.com/JuliaLang/julia/pull/55949.

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -33,7 +33,7 @@ function is_some_include(@nospecialize(f))
             f = f.value
         end
         if isa(f, Function)
-            if isdefined(Base, :IncludeInto) && isa(f, Base.IncludeInto)
+            if @static isdefined(Base, :IncludeInto) && isa(f, Base.IncludeInto)
                 return true
             end
             mod = Base.typename(typeof(f)).module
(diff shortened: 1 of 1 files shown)


src/lowered.jl
```diff
@@ -33,6 +33,9 @@ function is_some_include(@nospecialize(f))
             f = f.value
         end
         if isa(f, Function)
+            if isdefined(Base, :IncludeInto) && isa(f, Base.IncludeInto)
```
aviatesk, line 36:
```suggestion
            if @static isdefined(Base, :IncludeInto) && isa(f, Base.IncludeInto)
```

timholy: That changes doesn't seem necessary, but might as well have it (if further edits get made it might theoretically avoid some invalidation risk since `Base.IncludeInto` gets inferred as `Any` on existing Julias).

timholy (approved):
Thanks! I'll approve, then merge when the Julia PR merges.

Keno:
> Thanks! I'll approve, then merge when the Julia PR merges.

Needs to be the other way around, since Julia CI tests Revise now, so Julia PR can't merge unless Revise works.

timholy:
Good point. OK, merge at will.

PR #856, opened 2024-10-05
avoid private access to loaded_precompile

vtjnash:


src/loading.jl
```diff
@@ -43,8 +43,8 @@ function parse_pkg_files(id::PkgId)
             @assert reqs !== nothing
             pkgdata.requirements = reqs
             for chi in includes
-                if isdefined(Base, :loaded_precompiles) && haskey(Base.loaded_precompiles, id => buildid)
-                    mod = Base.loaded_precompiles[id => buildid]
+                if isdefined(Base, :maybe_loaded_precompile) && Base.maybe_loaded_precompile(id, buildid)
```
Keno, line 46: I'm confused by this being used as a Boolean here and a module below. Is there a comparison missing?

vtjnash: Yep, good point

src/loading.jl
```diff
@@ -43,8 +43,8 @@ function parse_pkg_files(id::PkgId)
             @assert reqs !== nothing
             pkgdata.requirements = reqs
             for chi in includes
-                if isdefined(Base, :loaded_precompiles) && haskey(Base.loaded_precompiles, id => buildid)
-                    mod = Base.loaded_precompiles[id => buildid]
+                if isdefined(Base, :maybe_loaded_precompile) && Base.maybe_loaded_precompile(id, buildid) isa Module
+                    mod = Base.maybe_loaded_precompile(id, buildid)
```
topolarity, line 47:
```suggestion
                if isdefined(Base, :maybe_loaded_precompile) && Base.maybe_loaded_precompile(id, buildid) isa Module
                    mod = Base.maybe_loaded_precompile(id, buildid)
                elseif isdefined(Base, :loaded_precompiles) && haskey(Base.loaded_precompiles, id => buildid)
                    mod = Base.loaded_precompiles[id => buildid]
```

To avoid breaking 1.11

KristofferC:
Weird it hasn't been updated in two days.

vtjnash:
I think Pkg is broken on v1.11, so the upload jobs have been failing for a couple days, if I read the logs right

topolarity:
I've tested nightly locally and it passes - Good to merge?

PR #863, opened 2024-11-14
Support two-argument version of Core.include

Keno:
`Core.include` does not support mapexpr, so the existing error was wrong. It does however suppoer the two-argument version with module (which we did support for `Base.include`), so wire that up appropriately.


aviatesk:
Thank you for fixing this 🙇

PR #869, opened 2024-12-05
Look through SSAValue to find `include`

Keno:
There's no guarantee that all include calls will be literal GlobalRefs and in fact https://github.com/JuliaLang/julia/pull/56746 will likely make them never GlobalRef.

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -24,6 +24,7 @@ add_dependencies!(methodinfo::MethodInfo, be::CodeEdges, src, isrequired) = meth
 add_includes!(methodinfo::MethodInfo, mod::Module, filename) = methodinfo
 
 function is_some_include(@nospecialize(f))
+    @assert !isa(f, Core.SSAValue) && !isa(f, JuliaInterpreter.SSAValue)
     if isa(f, GlobalRef)
         return f.name === :include
     elseif isa(f, Symbol)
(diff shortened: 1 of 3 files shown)


Keno:
Nightly CI failure will be https://github.com/JuliaLang/julia/pull/56756

src/lowered.jl
```diff
@@ -44,17 +45,21 @@ function is_some_include(@nospecialize(f))
...
-function hastrackedexpr(stmt; heads=LoweredCodeUtils.trackedheads)
+function hastrackedexpr(stmt, code; heads=LoweredCodeUtils.trackedheads)
     haseval = false
     if isa(stmt, Expr)
         haseval = matches_eval(stmt)
         if stmt.head === :call
             f = stmt.args[1]
+            while isa(f, Core.SSAValue) || isa(f, JuliaInterpreter.SSAValue)
+                f = code[f.id]
+            end
```
aviatesk, line 56: We have quite a bit of similar code in JuliaInterpreter and LoweredCodeUtils, so it might be a good idea to consolidate it into a single utility.  
Other code typically doesn’t use `while` like this but instead only follows a single level of `SSAValue` references. I’ll handle the refactoring later.

aviatesk (approved):

src/lowered.jl
Keno: There's lots of hacks all over the place. I think we need to take a much wider look at refactoring this stack next year, probably once JuliaLowering is ready. It's just all way too ad hoc and fragile

aviatesk: Yeah overhauling the whole stack at once with integrating JuliaLowering would be very reasonable.

timholy: Yes, once JuliaLowering is available we can make everything sane.

Issue #872, opened 2024-12-12
Revise tests clones general registry which is quite slow


KristofferC:
When running the Revise tests I see:

```
  Activating new project at `C:\Users\KRISTO~1\AppData\Local\Temp\jl_6nmNSw`
  Generating  project TestPackage:
    TestPackage\Project.toml
    TestPackage\src\TestPackage.jl
vvvvvvvvvvvvvvvvvvvvv
  Installing known registries into `C:\Users\Kristoffer\.julia`
     Cloning registry from "https://github.com/JuliaRegistries/General.git"
       Added registry `General` to `C:\Users\Kristoffer\.julia\registries\General`
^^^^^^^^^^^^^^^^^^
     Cloning git-repo `C:\Users\Kristoffer\AppData\Local\Temp\jl_vx8FqQ\TestPackage`
    Updating git-repo `C:\Users\Kristoffer\AppData\Local\Temp\jl_vx8FqQ\TestPackage`
    Updating registry at `C:\Users\Kristoffer\.julia\registries\General`
┌ Info: The General registry is installed via git. Consider reinstalling it via
│ the newer faster direct from tarball format by running:
│   pkg> registry rm General; registry add General
└ 
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
```

This is a little bit ironic considering

https://github.com/timholy/Revise.jl/blob/d21c32f397d115dd037b2d02f7fb7e24283d9286/test/runtests.jl#L3058

PR #881, opened 2025-01-12
Tests: Remove faulty registry install skip given registry install is fast now

IanButterworth:
Fixes https://github.com/timholy/Revise.jl/issues/872


IanButterworth:
I don't have repo permissions to request reviewers (or just merge). @KristofferC would you mind this and https://github.com/timholy/Revise.jl/pull/882

PR #894, opened 2025-03-13
[WIP] Fix struct/const revision

timholy:
The aim here is to provide full support for struct/const revision. This first commit just adds simple tests, but it already reveals something that needs to be addressed: when a `struct` gets revised, methods defined for that `struct` aren't automatically re-evaluated for the new type. Specifically, in the added tests, after revising the definition of `Point` we get
```julia
julia> methods(StructConst.mynorm)
# 1 method for generic function "mynorm" from StructConst:
 [1] mynorm(p::@world(StructConst.Point, 40598:40738))
     @ /tmp/jW7BpD8RL7/StructConst/src/StructConst.jl:12
```
and there is no defined method for a currently-valid `StructConst.Point`.

@Keno, if you have a moment, let me check my understanding of the current situation:

- method applicability is assessed by type-intersection, as usual, but now types can have a world-age
- the MethodInstances and CodeInstances involving the old type are still valid (their world age is not capped, but of course they require inputs with "old" types)
- there is no invalidation step needed for changes of the kind tested with the revision of Point (`invalidate_code_for_globalref!` is not called?)
- we can't count on `binding.backedges` to list everything that uses this binding
- the right solution is for Revise to:
  + detect that we're about to redefine a struct
  + search the entire system for any methods where that type appears in a signature
  + parse the source files, if necessary
  + redefine the struct
  + re-evaluate the method definitions

Presumably, waiting to do the last step as the last stage of a revision would be wise, as it is possible that more than one struct will be revised at the same time, and one might as well do each evaluation only once.


Keno:
> * method applicability is assessed by type-intersection, as usual, but now types (via their bindings) can have a world-age

I think that's a valid way to say it, but just to clarify, nothing about the type object itself changes, just the name by which you have to access it

> * the MethodInstances and CodeInstances involving the old type are still valid (their world age is not capped, but of course they require inputs with old types)

yes

> * there is no invalidation step needed for changes of the kind tested with the revision of Point (`invalidate_code_for_globalref!` is not called?)

I don't understand what you're asking

> * we can't count on `binding.backedges` to list everything that uses this binding

It lists all cross-module edges to get everything, you also need to iterate all methods in the same module as the binding.

> * the right solution is for Revise to:
>   
>   * detect that we're about to redefine a struct
>   * search the entire system for any methods where that type appears in a signature
>   * parse the source files, if necessary
>   * redefine the struct
>   * re-evaluate the method definitions

In the fullness of time, the correct way to implement Revise is to run the toplevel expressions in a non-standard evaluation mode that tracks dependency edges for any evaluated constants. However, that should maybe wait until after the JuliaLowering changes and the interpreter rewrite. I think your proposed strategy is reasonable in the meantime although not fully complete.

timholy:
> > there is no invalidation step needed for changes of the kind tested with the revision of Point (`invalidate_code_for_globalref!` is not called?)
> 
> I don't understand what you're asking

I was wondering whether some kind of callback in there could be helpful in short-circuiting the process of determining which methoddefs need to be re-evaluated, but if it's not called then that won't help. From what I can tell, that's not going to work anyway because it only traverses the module containing the type definition. But your comment

> It lists all cross-module edges to get everything, you also need to iterate all methods in the same module as the binding.

seems to be what I was looking for.

This seems fairly straightforward, thanks for the excellent groundwork.

Keno:
> I was wondering whether some kind of callback in there could be helpful in short-circuiting the process of determining which methoddefs need to be re-evaluated

No, there's no edges of any kind for signatures. You need to scan the entire system. However, an earlier version of that code did the same whole-system scan, so you can steal some code for whole-system scanning.

Keno:
There's two kinds of edges that are tracked there. One is explicit import/using. The other is to lowered code of method definitions (but only after the first inference). At no point is an edge ever added for an evaluation of a binding, only for compilation of code that references the binding.

timholy:
Just to cross-check:
```
                module StructConstUser
                using StructConst: Fixed, Point
                struct PointWrapper
                    p::Point
                end
                scuf(f::Fixed) = 33 * f.x
                scup(p::Point) = 44 * p.x
                end
```
yields a backedge for the `using StructConst: Fixed, Point` statement, but I don't see anything related to `scup` even if I evaluate it:
```julia
julia> e = only(b.backedges)
Binding StructConstUser.Point
   40598:∞ - explicit `using` from StructConst.Point
   1:0 - undefined binding - guard entry
```

Keno:
> but I don't see anything related to `scup` even if I evaluate it:

That's because `scup` doesn't use any bindings in its lowered code (as I said, no edges for signatures). However, even if you had like `mkscup() = Point()`, that binding would be to StructConstUser. To actually get a cross-module edge is not that easy, but you could do `@eval mkscup() = $(GlobalRef(StructConst, :Point))()`. The most common case to get cross-module edges is bindings. Also, not a lot of edges is good, that is the design goal, since we don't want to store them :).

timholy:
This is getting close, but there's a philosophical question to answer: should Revise try to hew to Base behavior as closely as possible, or should it add/modify behaviors for what might be a nicer interactive experience?

In this case, the issue is the following: when I redefine `Point`, in Base you can still call previously-defined functions on "old data." However, in this PR, calling functions on old data throws (or is intended to throw) a `MethodError`. Why might you want that? Because I imagine that a common user-error will be to forget to refresh your data with updated types, and if your whole code pipeline runs (using all the old definitions), I imagine people will bang their heads against walls trying to figure out why their edits aren't having the intended effect. In contrast, if they get a `no method foo(::@world(Point, m:n))`, they'll pretty quickly learn how you fix it.

I should say that initially I wasn't aiming in this direction, and this current proposal arose from noticing some unexpected subtleties about the order in which Revise's cache files get set up: the order in which you parse the source, lower the code, and rebind the name matters quite a lot, and some of the choices in this version are designed to compensate for the fact that Revise normally doesn't parse a file unless it's been edited. (Now, we'll also have to parse any file with a method whose signature references an updated type.) I think it's possible to hew to the Base behavior, if we decide that's better, but the initial bugs I faced led me to ask what behavior we actually want, and I came to the conclusion that it's likely a better user experience if we invalidate methods that only work on old types.

src/packagedef.jl
```diff
@@ -874,6 +882,33 @@ function revise(; throw=false)
...
+                m = pop!(reeval_methods)
+                methinfo = get(CodeTracking.method_info, m.sig, missing)
+                methinfo === missing && continue
+                if length(methinfo) != 1
+                    @warn "Multiple definitions for $(m.sig) found, skipping reevaluation"
+                    continue
+                end
+                Base.delete_method(m)  # ensure that "old data" doesn't get run with "old methods"
+                _, ex = methinfo[1]
+                eval_with_signatures(m.module, ex; mode=:eval)
```
timholy, line 898: Note to self: I was surprised that dependent types (e.g., `PointWrapper`) don't experience binding invalidation until their constructors are re-evaluated. `Core.eval(m.module, ex)` triggers the binding invalidation and gets us farther in the tests, but it doesn't add the appropriate signatures to `CodeTracking.method_info`. My sense is that JuliaInterpreter/JuliaLowering need to trigger binding invalidation from `LoweredCodeUtils.methoddef!`.

Keno:
> However, in this PR, calling functions on old data throws (or is intended to throw) a `MethodError`. Why might you want that?

I think that's fine. Conceptually Revise does both adding new methods and deleting old ones. Base is always "append only".

timholy:
The collateral damage to tests is something I'm looking into; if one runs all the tests, there appear to be some failures that predate this effort, including an incomplete updating of the ecosystem to https://github.com/JuliaLang/julia/pull/52415. I'll work on that.

But meanwhile, this seems to be working:

```julia
tim@diva:~/.julia/dev/Revise$ ~/src/juliaw/julia --startup=no --project -e 'using Pkg; Pkg.test(; test_args=["struct/const revision"], coverage=false)'
     Testing Revise
      Status `/tmp/jl_4EbO8B/Project.toml`
# Pkg output deleted
     Testing Running tests...
Test Summary: | Pass  Total   Time
Revise        |   30     30  10.3s
beginning cleanup
```

and may be ready for anyone who wants to review it.

aviatesk:
This is interesting.

So it looks like this PR tracks edges not only from binding to method signature, but also from binding to method body?

I think a similar mechanism will be needed when developing a new language server. If it's already implemented in Revise, I'd like to reuse it.

I haven't had a chance to look at this PR closely yet, but I'll read it later.

timholy:
I'll be curious to see whether it improves the efficiency of hacking on inference.

Keno:
> There's a separate bit of novelty for `Core.Compiler`: this PR also checks whether any of the entry points to type inference have been invalidated, and if so calls `bootstrap!`

I don't know if I like that. We already have `@activate Compiler[:codegen]` for this use case.

timholy:
I didn't know about `@activate`. Is there a demo of the workflow somewhere?

I can back that commit out, if there are other mechanisms to achieve the same aim.

KristofferC:
I tired this. On:

- Revise (this PR)
- LoweredCodeUtils master
- Julia 1.12.0-beta1

What I did was:

- Load Revise, Tensors.
- Make the following revision to Tensors: 
  ```diff
  diff --git a/src/Tensors.jl b/src/Tensors.jl
  index 4cd677b..5f04312 100644
  --- a/src/Tensors.jl
  +++ b/src/Tensors.jl
  @@ -63,6 +63,7 @@ julia> Tensor{1,3,Float64}((1.0, 2.0, 3.0))
   """
   struct Tensor{order, dim, T, M} <: AbstractTensor{order, dim, T}
       data::NTuple{M, T}
  +    x::Int
       Tensor{order, dim, T, M}(data::NTuple) where {order, dim, T, M} = new{order, dim, T, M}(data)
   end
  ```
- Press 1 in the REPL

with the following result:

```julia
julia> using Revise, Tensors

# updates Tensors file here

julia> 1
ERROR: UndefVarError: `Broadcasted` not defined in `StaticArrays.StableFlatten`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/StaticArrays/LSPcF/src/broadcast.jl:180
Revise evaluation error at /Users/kristoffercarlsson/.julia/packages/StaticArrays/LSPcF/src/broadcast.jl:180

Stacktrace:
  [1] methods_by_execution!(recurse::Any, methodinfo::Revise.CodeTrackingMethodInfo, docexprs::Dict{…}, mod::Module, ex::Expr; mode::Symbol, disablebp::Bool, always_rethrow::Bool, kwargs::@Kwargs{})
    @ Revise ~/JuliaPkgs/Revise.jl/src/lowered.jl:306
  [2] eval_with_signatures(mod::Module, ex::Expr; mode::Symbol, kwargs::@Kwargs{})
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:556
  [3] eval_with_signatures
    @ ~/JuliaPkgs/Revise.jl/src/packagedef.jl:553 [inlined]
  [4] instantiate_sigs!(modexsigs::OrderedCollections.OrderedDict{…}; mode::Symbol, kwargs::@Kwargs{})
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:564
  [5] instantiate_sigs!
    @ ~/JuliaPkgs/Revise.jl/src/packagedef.jl:560 [inlined]
  [6] maybe_extract_sigs_for_meths(meths::Set{Method})
    @ Revise ~/JuliaPkgs/Revise.jl/src/pkgs.jl:162
  [7] (::Revise.var"#107#108"{Bool})()
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:941
  [8] lock(f::Revise.var"#107#108"{Bool}, l::ReentrantLock)
    @ Base ./lock.jl:335
  [9] #revise#104
    @ ~/JuliaPkgs/Revise.jl/src/packagedef.jl:844 [inlined]
 [10] revise()
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:842
 [11] top-level scope
    @ REPL:1

caused by: UndefVarError: `Broadcasted` not defined in `StaticArrays.StableFlatten`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
  [1] lookup_var
    @ ~/.julia/packages/JuliaInterpreter/J7J9G/src/interpret.jl:5 [inlined]
  [2] step_expr!(recurse::Any, frame::JuliaInterpreter.Frame, node::Any, istoplevel::Bool)
    @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/J7J9G/src/interpret.jl:44
  [3] signature(recurse::Any, frame::JuliaInterpreter.Frame, stmt::Any, pc::Int64)
    @ LoweredCodeUtils ~/.julia/packages/LoweredCodeUtils/BIVzf/src/signatures.jl:50
  [4] methoddef!(recurse::Any, signatures::Vector{Any}, frame::JuliaInterpreter.Frame, stmt::Any, pc::Int64; define::Bool)
    @ LoweredCodeUtils ~/.julia/packages/LoweredCodeUtils/BIVzf/src/signatures.jl:595
  [5] methoddef!
    @ ~/.julia/packages/LoweredCodeUtils/BIVzf/src/signatures.jl:534 [inlined]
  [6] methods_by_execution!(recurse::Any, methodinfo::Revise.CodeTrackingMethodInfo, docexprs::Dict{…}, frame::JuliaInterpreter.Frame, isrequired::Vector{…}; mode::Symbol, skip_include::Bool)
    @ Revise ~/JuliaPkgs/Revise.jl/src/lowered.jl:354
  [7] methods_by_execution!
    @ ~/JuliaPkgs/Revise.jl/src/lowered.jl:317 [inlined]
  [8] methods_by_execution!(recurse::Any, methodinfo::Revise.CodeTrackingMethodInfo, docexprs::Dict{…}, mod::Module, ex::Expr; mode::Symbol, disablebp::Bool, always_rethrow::Bool, kwargs::@Kwargs{})
    @ Revise ~/JuliaPkgs/Revise.jl/src/lowered.jl:296
  [9] eval_with_signatures(mod::Module, ex::Expr; mode::Symbol, kwargs::@Kwargs{})
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:556
 [10] eval_with_signatures
    @ ~/JuliaPkgs/Revise.jl/src/packagedef.jl:553 [inlined]
 [11] instantiate_sigs!(modexsigs::OrderedCollections.OrderedDict{…}; mode::Symbol, kwargs::@Kwargs{})
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:564
 [12] instantiate_sigs!
    @ ~/JuliaPkgs/Revise.jl/src/packagedef.jl:560 [inlined]
 [13] maybe_extract_sigs_for_meths(meths::Set{Method})
    @ Revise ~/JuliaPkgs/Revise.jl/src/pkgs.jl:162
 [14] (::Revise.var"#107#108"{Bool})()
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:941
 [15] lock(f::Revise.var"#107#108"{Bool}, l::ReentrantLock)
    @ Base ./lock.jl:335
 [16] #revise#104
    @ ~/JuliaPkgs/Revise.jl/src/packagedef.jl:844 [inlined]
 [17] revise()
    @ Revise ~/JuliaPkgs/Revise.jl/src/packagedef.jl:842
 [18] top-level scope
    @ REPL:1
Some type information was truncated. Use `show(err)` to see complete types.
```

aviatesk:
I updated against the latest master branch.

I tested this branch with the simple case, but it seems that for `export`ed bindings, additional edge tracking is needed? 
In particular, the world ages of bindings that are `using`ed need to be updated, it seems.
For example, let's say we have an Example package like this:
```julia
module Example
export hello, Hello

struct Hello
    who::String
end

hello(x::Hello) = hello(x.who)

"""
    hello(who::String)

Return "Hello, `who`".
"""
hello(who::String) = "Hello, $who"

end
```
We get the following error:
```julia-repl
julia> using Revise

julia> using Example

julia> hello(Hello("world"))
"Hello, world"

julia> # Apply the following diff
       # ```diff
       # diff --git a/src/Example.jl b/src/Example.jl
       # index 65c7eae..c631814 100644
       # --- a/src/Example.jl
       # +++ b/src/Example.jl
       # @@ -2,10 +2,10 @@ module Example
       #  export hello, Hello
       #  
       #  struct Hello
       # -    who::String
       # +    who2::String
       #  end
       #  
       # -hello(x::Hello) = hello(x.who)
       # +hello(x::Hello) = hello(x.who2 * " (changed)")
       #  
       #  """
       #      hello(who::String)
       # ```

julia> hello(Hello("world"))
ERROR: MethodError: no method matching @world(Example.Hello, 38355:38383)(::String)
The type `@world(Example.Hello, 38355:38383)` exists, but no method is defined for this combination of argument types when trying to construct it.
Stacktrace:
 [1] top-level scope
   @ REPL[5]:1

julia> hello(Example.Hello("world")) # but this works
"Hello, world (changed)"
```

timholy:
@aviatesk your example works now.

timholy:
@KristofferC your example looks like a selective evaluation issue:
https://github.com/JuliaArrays/StaticArrays.jl/blob/e7c5e0162fe829f317b1d49aea7f192eb8c58eae/src/broadcast.jl#L171-L175
doesn't run because of the Julia version number, but somehow Revise still tries to step through it.

I've found that the problematic frame looks like this:
```

julia> Revise.LoweredCodeUtils.print_with_code(stdout, frame.framecode.src, isrequired)
  1 t 1 ─ %1   = StaticArrays.StableFlatten.:>=
  2 t │   %2   = StaticArrays.StableFlatten.VERSION
  3 t │   %3   =   dynamic (%1)(%2, v"1.11.0-DEV.103")
  4 t └──        goto #3 if not %3
  5 t 2 ─ %5   = StaticArrays.StableFlatten.Broadcast
  6 t │   %6   =   dynamic Base.getproperty(%5, :flatten)
  7 t │          $(Expr(:const, :(StaticArrays.StableFlatten.broadcast_flatten), :(%6)))
  8 f │          $(Expr(:latestworld))
  9 f └──        return %6
 10 f 3 ─        using Base: tail
  ⋮
```

The `return` statement is not marked. @aviatesk, any chance that https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/117 might fix it?

aviatesk:
Thanks for fixing my example!

> The `return` statement is not marked

Sure, I will look into it at least by tomorrow.

aviatesk:
> I didn't know about `@activate`. Is there a demo of the workflow somewhere? I've seen the docs, and they're good, but I suspect there's a bit more wisdom around effective usage than they convey.
> 
> I can back that commit out, if there are other mechanisms to achieve the same aim. Are you saying that `Revise.track(Core.Compiler)` should be deprecated?

Sorry for the delay, but I'd like to add some details about this.

Currently, when you use `InteractiveUtils.@activate Compiler`, the Compiler.jl implementation specified in your environment gets loaded. This also switches the Compiler.jl implementation used by reflection utilities like `code_typed` to the newly loaded one.
For the actual Compiler.jl implementation, besides a dummy implementation that just redirects to `Base.Compiler` (https://github.com/JuliaLang/BaseCompiler.jl), we expect the implementation to be used is a local checkout of `JuliaLang/julia/Compiler` or a specific remote `JuliaLang/julia` commit.

So when you're hacking on Compiler.jl, using `InteractiveUtils.@activate Compiler` is generally enough, and I think `Revise.track(Core.Compiler)` can be deprecated. In most cases, I don't see any benefit in applying changes to the `Base.Compiler` module and observing the changes for the Compiler development. There might still be situations where it still is potentially useful, like for debugging inference issues that only happen during bootstrap. But this problem itself is quite tricky, and I feel it's better to use simple reflections like `Core.eval` or `Core.println` for debugging it, rather than advanced tools like Revise. So in conclusion, I think `Revise.track(Core.Compiler)` can be deprecated.

@Keno Please let me know if you have anything to add.

aviatesk:
For some reason, @KristofferC's example seems to be working fine with the latest combination of Revise/LoweredCodeUtils/Tensors.

```julia-repl
julia> using Revise

julia> using Tensors

julia> fieldnames(Tensors.Tensor)
(:data,)

julia> fieldnames(Tensors.Tensor)
(:data, :x)

julia> VERSION
v"1.12.0-rc1.23"
```

timholy:
Not for me (or at least not on nightly).

timholy:
Amusingly/terrifyingly, I have even occasionally triggered the message `"Compiling the compiler. This may take several minutes ..."` just from using the REPL and making revisions 😱

timholy:
OK, I have to correct what I said above: it does work, both with the released version of LCU and with https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/117 (I'm not sure why it didn't before).

But with both I get
```
Compiling the compiler. This may take several minutes ...
Base.Compiler ──── 11.776275873184204 seconds
```
and obviously that needs to be fixed.

timholy:
Well, that was easy. This branch may now be worth trying more seriously (once tests get back to passing).

Keno:
I thought I had commented, but I guess not

> I think `Revise.track(Core.Compiler)` can be deprecated

This is very occasionally still useful, when the `Compiler` in your environment is not the same as the base compiler and you want to test things. So I don't think we need to remove it, we can just teach people about the new, better workflow for Compiler development, and `Revise.track` can remain an option for power users.

timholy:
I restored the ability for Revise to call `Compiler.bootstrap!()` but made it conditional on Revise being sure that it was triggered by changes made by Revise. This is conservative, and automatic revision will stop if methods get invalidated just by new method definitions (e.g., loading packages).

timholy:
struct revision tests are now passing again (the failure is only in `extra tests`).

In principle, this could be ready to merge. Please test in real workflows (nightly, as rc1 is currently broken).

MasonProtter:
> Please test in real workflows (nightly, as rc1 is currently broken).

Can you suggest a branch that does work?

timholy:
`nightly`

MasonProtter:
My bad, I completely misread the comment as saying that both `nightly` and `1.12` were broken. Sorry for the noise.

aviatesk:
You can also build from the latest `backports-1.12` branch if you need v1.12.

MasonProtter:
One issue I'm encountering on `nightly` is that using this branch, with a file `foobar.jl`
```julia
struct Foo
end

function f((; x)::Foo)
    x + 1
end
```
I get
```
julia> using Revise

julia> includet("/tmp/foobar.jl")
WARNING: Detected access to binding `Main.Foo` in a world prior to its definition world.
  Julia 1.12 has introduced more strict world age semantics for global bindings.
  !!! This code may malfunction under Revise.
  !!! This code will error in future versions of Julia.
Hint: Add an appropriate `invokelatest` around the access to this binding.
To make this warning an error, and hence obtain a stack trace, use `julia --depwarn=error`.
```
That warning message is not printed when I use the `master` branch of `Revise`.

timholy:
Ah, thanks for the reminder, I've seen that warning but focused so far on fixing "real bugs."

timholy:
Oh weird:
```
tim@hypnotic:/tmp/rvs$ julia +nightly -q
julia> includet("foobar.jl")
WARNING: Detected access to binding `Main.Foo` in a world prior to its definition world.
  Julia 1.12 has introduced more strict world age semantics for global bindings.
  !!! This code may malfunction under Revise.
  !!! This code will error in future versions of Julia.
Hint: Add an appropriate `invokelatest` around the access to this binding.
To make this warning an error, and hence obtain a stack trace, use `julia --depwarn=error`.

julia>
tim@hypnotic:/tmp/rvs$ julia +nightly -q --depwarn=error
julia> includet("foobar.jl")

julia>
```

?!?

nickrobinson251:
> Oh weird:

https://github.com/JuliaLang/julia/issues/58648 (although it's closed)

timholy:
Should be fixed @MasonProtter

lassepe:
I seem to have found a bug (commit [0158326](https://github.com/timholy/Revise.jl/pull/894/commits/0158326d6635b0207b30843733ba424736435fdd), Julia nightly 303b5d8c64d (2025-08-13 00:34 UTC))

1. Create Foo.jl file with content:
```julia
module Foo
export Bar

struct Bar end
end
```
2. Start a Julia REPL, run `using Revise; Revise.includet("Foo.jl"); using .Foo; Bar()`
3. Modify Foo.jl file by adding a field to `Bar`
```julia
module Foo
export Bar

struct Bar
  x
end
end
```
4. In the existing Julia REPL run `Bar(1)`:
```julia
julia> Bar(1)
ERROR: MethodError: no method matching @world(Main.Foo.Bar, 39034:39037)(::Int64)
The type `@world(Main.Foo.Bar, 39034:39037)` exists, but no method is defined for this combination of argument types when trying to construct it.
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1
```

KristofferC:
> You can also build from the latest backports-1.12 branch if you need v1.12.

Or just juliaup add the PR (prXYZ).

timholy:
@lassepe, this behavior observed at the end of your demo is quite surprising:

```julia
julia> fieldnames(Bar)
()

julia> fieldnames(Foo.Bar)
(:x,)

julia> Foo.Bar(1)
Main.Foo.Bar(1)
```

@Keno, is this a Julia bug failing to propagate the updated binding to `Main`?

timholy:
https://github.com/JuliaLang/julia/issues/59272

timholy:
@lassepe your issue is fixed in `julia +1.12-rc2` (just released)

xlxs4:
```julia-repl
┌ Revise
│  ┌ Warning: precompile directive
│  │     precompile(Tuple{mbody.sig.parameters[1], Symbol, Bool, Bool, Iterators.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:skip_include,), Tuple{Bool}}}, typeof(methods_by_execution!), Compiled, MI, Module, Expr})
│  │ failed. Please report an issue in Revise (after checking for duplicates) or remove this directive.
│  └ @ Revise ~/.julia/packages/Revise/FV0Wc/src/precompile.jl:65
└
```

<details>

<summary>REPL output</summary>

```julia-repl
julia> Pkg.add(name="Revise"; rev="teh/struct_revision")
    Updating git-repo `https://github.com/timholy/Revise.jl.git`
   Resolving package versions...
    Updating `~/.julia/environments/v1.12/Project.toml`
  [295af30f] + Revise v3.9.0 `https://github.com/timholy/Revise.jl.git#teh/struct_revision`
    Updating `~/.julia/environments/v1.12/Manifest.toml`
  [da1fd8a2] + CodeTracking v2.0.0
  [807dbc54] + Compiler v0.1.1
  [aa1ae85d] + JuliaInterpreter v0.10.5
  [6f1432cf] + LoweredCodeUtils v3.4.3
  [bac558e1] + OrderedCollections v1.8.1
  [ae029012] + Requires v1.3.1
  [295af30f] + Revise v3.9.0 `https://github.com/timholy/Revise.jl.git#teh/struct_revision`
  [56f22d72] + Artifacts v1.11.0
  [2a0f44e3] + Base64 v1.11.0
  [7b1f6079] + FileWatching v1.11.0
  [b77e0a4c] + InteractiveUtils v1.11.0
  [ac6e5ff7] + JuliaSyntaxHighlighting v1.12.0
  [76f85450] + LibGit2 v1.11.0
  [8f399da3] + Libdl v1.11.0
  [d6f4376e] + Markdown v1.11.0
  [ca575930] + NetworkOptions v1.3.0
  [de0858da] + Printf v1.11.0
  [3fa0cd96] + REPL v1.11.0
  [9a3f8284] + Random v1.11.0
  [ea8e919c] + SHA v0.7.0
  [6462fe0b] + Sockets v1.11.0
  [f489334b] + StyledStrings v1.11.0
  [cf7118a7] + UUIDs v1.11.0
  [4ec0a83e] + Unicode v1.11.0
  [e37daf67] + LibGit2_jll v1.9.0+0
  [29816b5a] + LibSSH2_jll v1.11.3+1
  [458c3c95] + OpenSSL_jll v3.5.1+0
Precompiling packages finished.
  1 dependency successfully precompiled in 12 seconds. 20 already precompiled.
  1 dependency had output during precompilation:
┌ Revise
│  ┌ Warning: precompile directive
│  │     precompile(Tuple{mbody.sig.parameters[1], Symbol, Bool, Bool, Iterators.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:skip_include,), Tuple{Bool}}}, typeof(methods_by_execution!), Compiled, MI, Module, Expr})
│  │ failed. Please report an issue in Revise (after checking for duplicates) or remove this directive.
│  └ @ Revise ~/.julia/packages/Revise/FV0Wc/src/precompile.jl:65
└
```

</details>

<details>

<summary>versioninfo</summary>

```julia-repl
julia> versioninfo()
Julia Version 1.12.0-rc2
Commit 72cbf019d04 (2025-09-06 12:00 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 5700U with Radeon Graphics
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver2)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 16 virtual cores)
```

</details>

<details>

<summary>Pkg status</summary>

```julia-repl
(@v1.12) pkg> st
Status `~/.julia/environments/v1.12/Project.toml`
  [295af30f] Revise v3.9.0 `https://github.com/timholy/Revise.jl.git#teh/struct_revision`

(@v1.12) pkg> st -m
Status `~/.julia/environments/v1.12/Manifest.toml`
  [da1fd8a2] CodeTracking v2.0.0
  [807dbc54] Compiler v0.1.1
  [aa1ae85d] JuliaInterpreter v0.10.5
  [6f1432cf] LoweredCodeUtils v3.4.3
  [bac558e1] OrderedCollections v1.8.1
  [ae029012] Requires v1.3.1
  [295af30f] Revise v3.9.0 `https://github.com/timholy/Revise.jl.git#teh/struct_revision`
  [56f22d72] Artifacts v1.11.0
  [2a0f44e3] Base64 v1.11.0
  [7b1f6079] FileWatching v1.11.0
  [b77e0a4c] InteractiveUtils v1.11.0
  [ac6e5ff7] JuliaSyntaxHighlighting v1.12.0
  [76f85450] LibGit2 v1.11.0
  [8f399da3] Libdl v1.11.0
  [d6f4376e] Markdown v1.11.0
  [ca575930] NetworkOptions v1.3.0
  [de0858da] Printf v1.11.0
  [3fa0cd96] REPL v1.11.0
  [9a3f8284] Random v1.11.0
  [ea8e919c] SHA v0.7.0
  [6462fe0b] Sockets v1.11.0
  [f489334b] StyledStrings v1.11.0
  [cf7118a7] UUIDs v1.11.0
  [4ec0a83e] Unicode v1.11.0
  [e37daf67] LibGit2_jll v1.9.0+0
  [29816b5a] LibSSH2_jll v1.11.3+1
  [458c3c95] OpenSSL_jll v3.5.1+0
```

</details>

timholy:
Thanks for the report. I need to fix that, but FYI it's nothing to worry about in terms of Revise working properly.

jakobnissen:
I can't get Revise to work at all with this PR. MWE:
* Make a new package `Foo` with the content (in `src/Foo.jl`):
```julia
module Foo

struct Foo
    x::Int
end

foo() = 2

end # module Foo
```
* Launch Julia 1.12.0-rc2, with an environment only containing `Foo` and Revise with this PR (`https://github.com/timholy/Revise.jl.git#teh/struct_revision`), no startup file
* In the REPL, type `julia> using Revise, Foo; Foo.foo()`. It works correctly
* Change `foo() = 2` to `foo() = 1`
* Now, in the REPL:
```julia
julia> Foo.foo()
┌ Error: Failed to revise /tmp/Foo/src/Foo.jl
│   exception =
│    invalid redefinition of constant Foo
│    Stacktrace:
│     [1] queuenext!(iter::JuliaInterpreter.ExprSplitter)
│       @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/378J1/src/construct.jl:492
└ @ Revise ~/.julia/packages/Revise/FV0Wc/src/packagedef.jl:785
┌ Warning: The running code does not match the saved version for the following files:
│
│   /tmp/Foo/src/Foo.jl
│
│ If the error was due to evaluation order, it can sometimes be resolved by calling `Revise.retry()`.
│ Use Revise.errors() to report errors again. Only the first error in each file is shown.
│ Your prompt color may be yellow until the errors are resolved.
└ @ Revise ~/.julia/packages/Revise/FV0Wc/src/packagedef.jl:957
2
```

lassepe:
@jakobnissen, isn't the problem simply that both your module and the struct are called `Foo`? I think that never worked.

lassepe:
A modified version of my original setting above still seems to be broken (in a different way) on julia 1.12-rc2 and the lateset version of this branch


**Steps to reproduce**

1. Generate a package: `]generate StructRevision`
2. Add the following content to `src/StructRevision`
```julia
module StructRevision

struct Foo
  x::Int
end

bar(::Foo) = 1

end # module StructRevision
```
3. With julia 1.12-rc2 from the REPL:
```julia
❯ julia +1.12 --startup-file=no --project
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.12.0-rc2 (2025-09-06)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org release
|__/                   |

julia> using Revise, StructRevision

julia> foo1 = StructRevision.Foo(1)
StructRevision.Foo(1)

julia> StructRevision.bar(foo1)
1

julia> # after changing Foo.x from Int to Float64:

julia> foo2 = StructRevision.Foo(1)
StructRevision.Foo(1.0)

julia> StructRevision.bar(foo1)
1

julia> StructRevision.bar(foo2)
ERROR: MethodError: no method matching bar(::StructRevision.Foo)
The function `bar` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  bar(::@world(StructRevision.Foo, 38519:38547))
   @ StructRevision ~/worktree/BugReports/StructRevision.jl/src/StructRevision.jl:7

Stacktrace:
 [1] top-level scope
   @ REPL[7]:1
```

timholy:
Hmm, there are working tests that are awfully close to your example, e.g., https://github.com/timholy/Revise.jl/blob/0158326d6635b0207b30843733ba424736435fdd/test/runtests.jl#L2453-L2458
I'll have to check and see what the difference may be.

jakobnissen:
Sorry for my noise before. I found another MWE to reproduce my issue:
```julia
module Foo

abstract type AbstractBar end

struct Bar
    x::Int
end

Bar() = Bar(1)

end # module Foo
```
Now, change `Bar` to subtype `AbstractBar`. The zero-arg constructor's only method `Bar()` is deleted by Revise, but the updated method is not added back.

timholy:
@lassepe fixed in 962c7cb
@jakobnissen fixed in 1583289

timholy:
@xlxs4 fixed in 9e3fd08

timholy:
PSA: anyone who is using this via `Pkg.add(name="Revise"; rev="teh/struct_revision")` will need to re-execute that line to get the latest.

lassepe:
Nice! I stress-tested it a bit more, and all the stuff I tried before now works flawlessly. ~~The only minor oddity I found is that, when a struct is modified from parametric to non-parametric, the old methods don't get deleted and the old instances + dispatches still work. I am not sure if that is considered a bug because everything from the current world age still works as expected (in the sense that it refuses to dispatch on the outdated `bar` method table)~~
There seems to be another edge case that occurs when one removes a type parameter and adds it back:

1. Generate a new package with `]generate StructRevision`
2. Add the following to `src/StructRevision.jl`
```julia
module StructRevision

export Foo, bar

struct Foo{T}
  x::T
end

bar(::Foo{T}) where {T} = "parametric with $T"

end # module StructRevision
```
3. Run the following from the REPL with modification instructions in the comments:
```julia
worktree/BugReports/StructRevision.jl is 📦 v0.1.0 via ஃ v1.10.10 took 5s
❯ julia +1.12 --project=. --startup-file=no
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.12.0-rc2 (2025-09-06)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org release
|__/                   |

julia> using Revise, StructRevision

julia> foo = Foo(1)
Foo{Int64}(1)

julia> bar(foo)
"parametric with Int64"

julia> # change Foo to: `struct Foo x::Int end`

julia> foo2 = Foo(1)
Foo(1)

julia> bar(foo)
"parametric with Int64"

julia> bar(foo2)
ERROR: MethodError: no method matching bar(::Foo)
The function `bar` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  bar(::@world(Foo, 38519:38547){T}) where T
   @ StructRevision ~/worktree/BugReports/StructRevision.jl/src/StructRevision.jl:9

Stacktrace:
 [1] top-level scope
   @ REPL[6]:1

julia> # now change Foo back to its original definition

julia> foo = Foo(1)
Foo{Int64}(1)

julia> bar(foo)
ERROR: MethodError: no method matching bar(::Foo{Int64})
The function `bar` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  bar(::@world(Foo, 38519:38547){T}) where T
   @ StructRevision ~/worktree/BugReports/StructRevision.jl/src/StructRevision.jl:9

Stacktrace:
 [1] top-level scope
   @ REPL[8]:1
```

jkrumbiegel:
I tried it on Makie and it immediately failed. With CairoMakie and Makie dev'ed I did `using CairoMakie; lines(cumsum(randn(100)))` to get a first plot. Then I went to `src/makielayout/types.jl` and renamed the field `reset_timer` on `ScrollZoom` to `reset_time` just to see what happens. Executing another plot command then printed:

```julia
┌ Error: Failed to revise /Users/krumbiegel/.julia/dev/Makie/Makie/src/makielayout/types.jl
│   exception =
│    MethodError: no method matching getfile(::Nothing)
│    The function `getfile` exists, but no method is defined for this combination of argument types.
│    
│    Closest candidates are:
│      getfile(::JuliaInterpreter.Frame)
│       @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/378J1/src/utils.jl:421
│      getfile(::JuliaInterpreter.FrameCode, ::Any)
│       @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/378J1/src/utils.jl:416
│      getfile(::JuliaInterpreter.Frame, ::Any)
│       @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/378J1/src/utils.jl:421
│      ...
│    
│    Stacktrace:
│     [1] top-level scope
│       @ ~/.julia/dev/Makie/Makie/src/makielayout/types.jl:190
│    Revise evaluation error at /Users/krumbiegel/.julia/dev/Makie/Makie/src/makielayout/types.jl:190
│    
└ @ Revise ~/.julia/packages/Revise/il844/src/packagedef.jl:783
┌ Warning: The running code does not match the saved version for the following files:
│ 
│   /Users/krumbiegel/.julia/dev/Makie/Makie/src/makielayout/types.jl
│ 
│ If the error was due to evaluation order, it can sometimes be resolved by calling `Revise.retry()`.
│ Use Revise.errors() to report errors again. Only the first error in each file is shown.
│ Your prompt color may be yellow until the errors are resolved.
└ @ Revise ~/.julia/packages/Revise/il844/src/packagedef.jl:957
```

I then tried to `Revise.retry()` to see what would happen and that printed something else

```julia
julia> Revise.retry()
┌ Error: Failed to revise /Users/krumbiegel/.julia/dev/Makie/Makie/src/makielayout/types.jl
│   exception =
│    UndefVarError: `#1217###mixin#307` not defined in `Makie`
│    Suggestion: check for spelling errors or missing imports.
│    Stacktrace:
│     [1] top-level scope
│       @ ~/.julia/dev/Makie/Makie/src/colorsampler.jl:14
│    Revise evaluation error at /Users/krumbiegel/.julia/dev/Makie/Makie/src/colorsampler.jl:14
│    
│    Stacktrace:
│     [1] top-level scope
│       @ ~/.julia/dev/Makie/Makie/src/makielayout/types.jl:190
│    Revise evaluation error at /Users/krumbiegel/.julia/dev/Makie/Makie/src/makielayout/types.jl:190
│    
└ @ Revise ~/.julia/packages/Revise/il844/src/packagedef.jl:783
```

lassepe:
Ah, here **is** a bug: if in my example above you go "full circle" by adding back the parameter `T`, new instances won't be accepted by `bar`. I updated my comment above accordingly.

timholy:
@jkrumbiegel, thanks for the example. I might need your help, because Makie's recipe system makes it a bit hard for me to look up how types are defined. I created a [gist](https://gist.github.com/timholy/1aaaa2c281f22116f61a7f6185f07b3e) (because it exceeded GitHub's message limit) with a log consisting of:
- modules and expressions it's "scanning" to extract method signatures
- types that it thinks have changed their definition, e.g., "### REVISED TYPE ###\n\nAxis"
- methods that take that type in their signature constraints (and which therefore need to be re-evaluated), e.g., "#### HERE IS A METHOD FOR THAT TYPE #####\nm = kwcall(::NamedTuple, ::typeof(hidedecorations!), la::Axis) @ Makie ~/.julia/environments/v1.12/dev/Makie/Makie/src/makielayout/blocks/axis.jl:1156

So it seems to start trying to start redefining Axis (does that wrap `ScollZoom`?) and methods that have `::Axis` in their signatures. Does that make sense? The type that ends up causing it to error is `Plot{PlotFunc, T}` and the method `get(x::Plot, key::Symbol, default) @ Makie ~/.julia/environments/v1.12/dev/Makie/Makie/src/compute-plots.jl:29`. It chokes when it gets to `@recipe Image` and it may have something to do with inconsistent gensyms from your whole `DocumentedAttributes`/`mixin` system? Revise has to re-evaluate code and if that means gensyms are created, they probably/certainly won't be created with the same numbers as the first time.

Let me know if these somehow make sense or if Revise is just going off the rails somehow.

jkrumbiegel:
> So it seems to start trying to start redefining Axis (does that wrap ScollZoom?) and methods that have ::Axis in their signatures. Does that make sense? 

Hm I'm not sure if it makes sense, the type Axis shouldn't change if ScrollZoom changes. ScrollZoom is used when instantiating an Axis via `initialize_block!` and `register_events!` in there but that should only cause recompilation of those functions, no? Not everything that uses Axis in its arguments, as Axis itself should be the same.

Maybe Makie's system where structs are defined via complicated macros just ties too many things together so that one small change ripples outward further than it should. The thing with the mixin kind of makes me think that, as it was definitely not designed with revise-ability in mind and is maybe a bit "weird" in places. Like this macro-within-macro thing here https://github.com/MakieOrg/Makie.jl/blob/46350344714203b165c696947e37051708bf332a/Makie/src/recipes.jl#L506-L508 in which there is indeed a gensym with `mixin` happening here https://github.com/MakieOrg/Makie.jl/blob/46350344714203b165c696947e37051708bf332a/Makie/src/recipes.jl#L319

timholy:
Thanks for the perspective, that's extraordinarily helpful. Merely knowing that I should start at the "head" of this log (why did it redefine `Axis`?) rather than the "tail" (what to do about `mixin`?) might save me a couple hours.

Re the gensyms and revisability, do keep in mind that the only reason Revise needs to go to all these shenanigans is because we don't yet have JuliaLowering: to track changes in line numbers and thus "repair" the line numbers in stacktraces, Revise has to map source text ↔ signatures in a manner more dynamic than plain line numbers (in case you insert code earlier in the file). Currently there isn't a way to do that besides reparse/lower/almost-evaluate all the source files, but JuliaLowering should provide a much better way. Since hopefully 🤞 JuliaLowering will land sometime in the Julia 1.13 or 1.14 timeframe, one option is to just wait it out.

MasonProtter:
So this has been more or less working fine for me day to day for the past few weeks. Maybe we should just merge this into `main` and register is and then work on fixing the test suite later? Or are there known serious regressions from this?

timholy:
> Or are there known serious regressions from this?

Yes (e.g., https://github.com/timholy/Revise.jl/pull/894#issuecomment-3279854151). When it fails, it's strictly worse than just using `master` (and not having good struct revision).

hz-xiaxz:
> JuliaLowering will land sometime in the Julia 1.13 or 1.14 timeframe, one option is to just wait it out.

Now that it's merged in 1.14, what will happen after that. Revise.jl would need a major rewrite to depend on it?

KristofferC:
Only the repo is merged into the Julia repo. Nothing has changed other than that.

ufechner7:
Any chance that this will be merged for Julia 1.12.x ?

timholy:
I certainly hope so!

aviatesk:
Ah, it's been a long journey. This work is finally wrapping up.

I think Revise can now handle many cases for updating `struct` definitions. On my end, I've confirmed that all the issues brought up in this long thread are resolved, so it would be great if you could give it another try when you have some time.

Starting today, I'll spend a few days testing it, and the plan is to merge it around the new year.

@timholy I've re-evaluated and implemented the invalidations Revise needs to handle to enable struct revision in most of realistic cases. This ended up adding nearly 1000 new lines of changes, and the logic has been significantly rewritten from the initial struct revision implementation, so I'd like to leave a few points to note for future reference:
- To add information to the `pkgdatas` cache about which type definition an expression is associated with, I've generalized the `ModuleExprsSigs` type to `ModuleExprsInfos`, and its leaf type has been changed from the previous `SigInfo` to `Union{SigInfo,TypeInfo}`.
    - However, for `TypeInfo`, I haven't prepared a lightweight cache on the CodeTracking side that corresponds to `CodeTracking.method_info` yet. Revise didn't need it for a correct implementation, and it would be a hassle to maintain compatibility with CodeTracking, so I decided it's not worth doing right now. But it might be useful for optimization, so I've left a TODO comment in `packagedef.jl` just in case.
- Thanks to Julia's binding partition support, Revise could, in principle, correctly propagate the effects associated with binding redefinition to the active session (regardless of whether the binding is `const` or not), not just the struct revision implemented in this PR, by cleverly using `delete_binding` and similar functions. This is a superset of the "struct revision" implemented by this PR, which I'm calling "binding revision.", meaning it enables revision in cases like the following:
  ```julia
    global n::Int = 10 # and then modified to `n::Int = 5`
    for i = 1:n
        @eval func(::Val{$i}) = $i
    end
    ```
    However, to handle this correctly, we'd need to explicitly manage access to bindings accompanied with top-level code execution and track their edges, which is quite technically challenging. I might do that in the future, but I don't know when. For now, this PR only achieves "struct revision," so I've updated the PR description based on that.
- Regarding the invalidation logic that happens with type redefinition, you can understand it by looking at `record_invalidations_for_type_deletion!`. Tests were added also.
- I've removed the logic that automatically performs Compiler.jl's bootstrap. I wasn't entirely confident if the logic was correct, and I also prefer calling `bootstrap` manually. Also, I didn't think there was a strong need to add such additional revision logic specifically for Compiler.jl, so I've removed it for now.

- One of the minor things I may finish before merging is the revision ordering problem. Currently, revision follows a simple flow like this:
    - First, execute the code modified by the user.
    - Redefine type definitions related to that code.
    - Redefine methods related to that code.
  However, this approach might cause errors if the user's modified code includes methods that should be defined using new types. We might need to implement a revision order that first defines all related type definitions, including those changed by the user, and then redefines related method definitions. But for now, the current logic works fine in the scope of my testing, so I'm keeping this naive logic for the time being.

aviatesk:
@jkrumbiegel As far as I've tested on my end, the issue you found seems to be resolved. It would be helpful if you could try it again on your side too.
```julia-repl
julia> using Revise
Precompiling Revise finished.
  1 dependency successfully precompiled in 6 seconds

julia> using CairoMakie

julia> fieldnames(CairoMakie.Makie.ScrollZoom)
(:speed, :reset_timer, :prev_xticklabelspace, :prev_yticklabelspace, :reset_delay)

julia> lines(cumsum(randn(100)))

julia> fieldnames(CairoMakie.Makie.ScrollZoom) # after edit (reset_timer -> reset_time)
(:speed, :reset_time, :prev_xticklabelspace, :prev_yticklabelspace, :reset_delay)

julia> lines(cumsum(randn(100)))
```

jkrumbiegel:
@aviatesk, thanks for testing although I did make some changes to Makie which resolved the main issue. So this was not necessarily Revise's fault (it did not trip it up before of course). If you want to see whether your changes also made a difference you'd have to compare with a version before https://github.com/MakieOrg/Makie.jl/pull/5402 was merged. The warnings I mention over there are gone in your example so that would be a big improvement anyway.

aviatesk:
I ran the same test on Makie and CairoMakie with https://github.com/MakieOrg/Makie.jl/pull/5402 reverted, and it seems to be working without warnings.

topolarity:
This is working pretty well for me locally so far! Nice work @aviatesk . 

> it enables revision in cases like the following:
>  ```julia
>    global n::Int = 10 # and then modified to `n::Int = 5`
>    for i = 1:n
>        @eval func(::Val{$i}) = $i
>    end
>  ```
  
I suspect there are a lot of complexities associated with trying to respond to updates to mutable bindings so I wouldn't expect that to work (yet?), but I did almost immediately run into the case that `const` updates do not cascade:
```julia
const IdVector = Vector{Int} # edit and `Graph` is not updated
struct Graph
    allocated::IdVector
    conflicted::IdVector
    resolved::IdVector
end
```

topolarity:
Same limitation seems to apply for type-aliases:
```julia
IdArray{N} = Array{Int64,N} # change and `Graph` is not updated
struct Graph
    allocated::IdArray{1}
    conflicted::IdArray{1}
    resolved::IdArray{1}
end
```
Is that expected?

This one is a little weird, because it _is_ a type you're re-defining at top-level. It's just not a `struct` definition.

aviatesk:
> Same limitation seems to apply for type-aliases:
> 
> ```julia
> IdArray{N} = Array{Int64,N} # change and `Graph` is not updated
> struct Graph
>     allocated::IdArray{1}
>     conflicted::IdArray{1}
>     resolved::IdArray{1}
> end
> ```
> 
> Is that expected?
> 
> This one is a little weird, because it _is_ a type you're re-defining at top-level. It's just not a `struct` definition.

Thanks for testing this! And yeah, Revise doesn't handle this case right now. It's what I'd call a 'binding revision' case, rather than a 'struct revision'. But I think we could support it by adding a simple 'ad-hoc hack'. The same goes for the first case you mentioned; we could add that before merging this PR (added a new TODO item for the OP).
Generalizing the hack would require some pretty advanced techniques.

aviatesk:
Gave it a shot, but it turned out to be tricky to handle even simple cases where const binding or type alias is used in struct fields with just a simple ad-hoc hack.  believe we need a more comprehensive approach to support "binding revision", but I don't think it should be implemented in this PR. So I've removed it from this PR's blocking items.

lassepe:
I've been using this for ~6h in a side project of mine since the latest set of commits was pushed yesterday, and I get a lot of value out of it! Thanks for this great work! I didn't do any dedicated "let's try to break this" but rather just used it organically as part of my regular workflow. I've found it to be very robust in that context.

aviatesk:
@lassepe, thanks for testing it. Glad it worked for your case.

aviatesk:
Alright, let's merge this.
After merging, I'll update NEWS.md and release v3.13.

timholy:
Thanks ever so much for all the work you put into finishing this, @aviatesk !! It was indeed a long and somewhat tortuous journey.

Issue #900, opened 2025-04-07
Revise should precompile more methods


PatrickHaecker:
With
```
if isinteractive()
    using Revise
end
```
in `~/.julia/config/startup.jl` and Julia 1.11.4 I get
```
> julia --trace-compile=stderr -ie "exit()"
precompile(Tuple{typeof(Revise.iswritable), String})
precompile(Tuple{typeof(Base.first), Array{Any, 1}})
precompile(Tuple{typeof(Base.iszero), Float64})
precompile(Tuple{Revise.var"#118#119"})
precompile(Tuple{typeof(Revise.active_repl_backend_available)})
precompile(Tuple{typeof(FileWatching.uvfinalize), FileWatching.FileMonitor})
```

while without `using Revise` there is no `precompile` done in the above test case.

PatrickHaecker:
Ok, I need help here. After reading the precompilation tutorial, I thought that introducing a backedge would do the trick like in
```julia
    __first() = Base.first(Any[])
    @warnpcfail precompile(__first, ())
```
but that was not successful. Calling the backedge explicitly is also not successful:
```julia
    __first() = Base.first(Any[1])
    __first()
```

So who knows how to get rid of the remaining four precompilations during runtime?

PatrickHaecker:
Status with Julia 1.12.0-rc1 and Revise v3.8.0:
```
precompile(Tuple{typeof(Base.indexed_iterate), Tuple{String, Array{Base.CacheHeaderIncludes, 1}, Array{Base.PkgId, 1}, UInt128}, Int64})
precompile(Tuple{typeof(Base.indexed_iterate), Tuple{String, Array{Base.CacheHeaderIncludes, 1}, Array{Base.PkgId, 1}, UInt128}, Int64, Int64})
precompile(Tuple{typeof(Base.setproperty!), Revise.PkgData, Symbol, Array{Base.PkgId, 1}})
precompile(Tuple{typeof(Base.iterate), Array{Base.CacheHeaderIncludes, 1}})
precompile(Tuple{typeof(Base.getproperty), Base.CacheHeaderIncludes, Symbol})
precompile(Tuple{Type{Pair{A, B} where B where A}, String, Revise.FileInfo})
precompile(Tuple{typeof(Base.iterate), Array{Base.CacheHeaderIncludes, 1}, Int64})
precompile(Tuple{typeof(Base._atexit), Int32}) # recompile
precompile(Tuple{typeof(Base.iszero), Float64})
[...]

PR #901, opened 2025-04-07
Get rid of two precompiles from trivial Julia execution

PatrickHaecker:
With only using Revise in the startup.jl and this commit, there are two precompiles less with
`julia --trace-compile=stderr -ie "exit()"`

This partially addresses #900.


timholy:
Thanks, BTW!

PR #904, opened 2025-04-18
Correctly revise methods defined in external method tables

serenity4:
Fixes https://github.com/timholy/Revise.jl/issues/646.

Requires:
- https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/680
- https://github.com/timholy/CodeTracking.jl/pull/140
- https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/125

To do before merging:
- [x] Test this "in the wild" with some large packages (SPIRV.jl and GPUCompiler-dependent packages)
- [x] Update documentation, docstrings and comments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
         version: ${{ matrix.version }}
         show-versioninfo: ${{ matrix.version == 'nightly' }}
     - uses: julia-actions/cache@v2
+    - run: julia --project -e 'using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/serenity4/CodeTracking.jl#parametrize-by-mt https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"'
     - uses: julia-actions/julia-buildpkg@latest
     # Revise's tests need significant customization
     # Populate the precompile cache with an extraneous file, to catch issues like in #460
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
         version: ${{ matrix.version }}
         show-versioninfo: ${{ matrix.version == 'nightly' }}
     - uses: julia-actions/cache@v2
-    - run: julia --project -e 'using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/timholy/CodeTracking.jl#master https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"'
+    - run: julia --project -e 'using Pkg; using Pkg; Pkg.add([PackageSpec(; url="https://github.com/serenity4/JuliaInterpreter.jl", rev="codetracking-v2"), PackageSpec(; name = "CodeTracking", rev="master"), PackageSpec(; url = "https://github.com/serenity4/LoweredCodeUtils.jl", rev = "support-external-methodtables")])'
     - uses: julia-actions/julia-buildpkg@latest
     # Revise's tests need significant customization
     # Populate the precompile cache with an extraneous file, to catch issues like in #460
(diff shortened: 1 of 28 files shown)


src/lowered.jl
```diff
@@ -16,7 +16,7 @@ end
 # This defines the API needed to store signatures using methods_by_execution!
 # This default version is simple and only used for testing purposes.
 # The "real" one is CodeTrackingMethodInfo in Revise.jl.
```
timholy, line 18: Now `packagedef.jl` I guess... (xref #540)

src/packagedef.jl
```diff
@@ -1178,15 +1178,15 @@ function get_def(method::Method; modified_files=revision_queue)
...
     # So that we don't call it again, store missingness info in CodeTracking
-    CodeTracking.invoked_setindex!(CodeTracking.method_info, method.sig, missing)
+    CodeTracking.invoked_setindex!(CodeTracking.method_info, CodeTracking.method_info_key(method), missing)
     return false
 end
 
 function get_def(method, pkgdata, filename)
     maybe_extract_sigs!(maybe_parse_from_cache!(pkgdata, filename))
-    return get(CodeTracking.method_info, method.sig, nothing)
+    return get(CodeTracking.method_info, CodeTracking.method_info_key(method), nothing)
```
timholy, line 1189: Perhaps we should add `method_info_key` to `using CodeTracking: ...`

It's slightly idiosyncratic, but I like the module-scoping of `CodeTracking.method_info` because it emphasizes who keeps the data. But `method_info_key` seems like we could just make it a known function in the Revise namespace. But if you aren't a fan of this proposal, it's fine the way it is.

timholy (approved):
This looks great. I don't think that #815 offers anything that this doesn't, so this PR could close that one unless you disagree.

src/packagedef.jl
serenity4: That makes perfect sense, I like this style.

serenity4:
I looked into supporting the `mapexpr`-form of include, and it doesn't strike me that it would require changes to either CodeTracking or LoweredCodeUtils. I'm not done with that, but it's getting a bit involved and I'm a bit low on time these days.

To avoid leaving this PR rotting for too long I think it might be good to move forward with the release of CodeTracking 2.0 and LoweredCodeUtils 4.0 after https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/125, so we can merge this PR. Then, if the `mapexpr` support still eventually requires a new breaking release for either, I would hope the disruption to not be too impactful.

@timholy what do you think?

.github/workflows/ci.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         version: ${{ matrix.version }}
         show-versioninfo: ${{ matrix.version == 'nightly' }}
     - uses: julia-actions/cache@v2
+    - run: julia --project -e 'using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/timholy/CodeTracking.jl#master https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"'
```
serenity4: To be removed before merge.

.github/workflows/ci.yml
```diff
@@ -80,6 +81,7 @@ jobs:
         # We also need to pick up the Git tests, but for that we need to `dev` the package
         echo "Git tests"
         julia --code-coverage=user -e '
+          using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/timholy/CodeTracking.jl#master https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"
```
serenity4: To be removed before merge.

serenity4:
Once LoweredCodeUtils v3.4.3 is available in the registry (https://github.com/JuliaRegistries/General/pull/135445), if all checks pass this is good to merge!

PR #907, opened 2025-04-19
bump LCU version to 3.2

aviatesk:


src/lowered.jl
```diff
@@ -54,7 +54,7 @@ end
 
 
 # This is not generally used, see `is_method_or_eval` instead
-function hastrackedexpr(stmt, code; heads=LoweredCodeUtils.trackedheads)
+function hastrackedexpr(@nospecialize(stmt), code)
```
aviatesk, line 57: There is a duplicated definition in LCU side. Maybe we should refactor them?

timholy (approved):

PR #934, opened 2025-07-12
Use `invokelatest` when printing errors after `includet`

Keno:
Matches the behavior in the REPL. Useful in general in case included code is required for proper printing and will also be required in the tests after https://github.com/JuliaLang/julia/pull/58976.


Keno:
nightly/pre CI fails on main also (but was fine three weeks ago). Needs to be investigated, but unrelated to this PR.

PR #938, opened 2025-07-15
separate `RelocatableExpr` into a subpackage

aviatesk:
I want to use the `RelocatableExpr` definition in a new language server project, so we are separating it as a subpackage of Revise. There are no functional changes when using Revise.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,7 @@
 name = "Revise"
 uuid = "295af30f-e4ad-537b-8983-00126c2a3abe"
 version = "3.8.0"
+authors = ["Tim Holy <tim.holy@gmail.com>"]
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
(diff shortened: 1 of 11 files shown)


timholy (approved):

timholy:
Is this still something you want, @aviatesk? Feel free to merge (though it will need rebasing) whenever.

aviatesk:
Thanks for your comment, @timholy .

As I mentioned in the description of https://github.com/timholy/Revise.jl/pull/962,

> We considered other approaches such as implementing their own Revise-like system on the JET[LS] side (e.g. [#938](https://github.com/timholy/Revise.jl/pull/938)), but concluded that implementing incremental analysis by making Revise itself extensible would be better in terms of implementation reliability and future maintenance costs.

Instead of implementing a "Revise-like system" externally, I'd like to proceed with the integration into LS by making Revise itself extensible. So, since I'm planning a more general refactoring than what this PR covers, I'll close this PR.

Just for your information, my current refactoring plan is to split Revise into:
- ReviseCore (diff and patch algorithm and data structures)
- Revise (file watching and REPL integration, user hooks)

Let me discuss the details of this in a separate PR.

Issue #634, opened 2021-06-18
Revise does not apply functions in the two-argument form of `include`


ranocha:
The two-argument form `include(mapexpr::Function, path::AbstractString)` allows transforming the code in the file at `path`. However, Revise does not seem to apply the same transformation `mapexpr` when re-loading code that is `include`d using this form. For example
```julia
julia> using Pkg; Pkg.activate(temp=true); Pkg.add("Revise")
  Activating new environment at `/tmp/jl_D95G2q/Project.toml`
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General`
   Resolving package versions...
    Updating `/tmp/jl_D95G2q/Project.toml`
  [295af30f] + Revise v3.1.17
[...]

julia> "BrokenModule" |> mkdir |> cd

julia> "src" |> mkdir |> cd

julia> open("BrokenModule.jl", "w") do io
           println(io, """
               module BrokenModule

               include(expr -> quote @fastmath begin \$expr end end, "broken_a.jl")

               end
           """)
       end

julia> open("broken_a.jl", "w") do io
           println(io, """
               foo(a, b, c) = a * b + c
           """)
       end

julia> cd(dirname(pwd()))

julia> open("Project.toml", "w") do io
           println(io, """
               name = "BrokenModule"
               uuid = "79e551b4-d023-11eb-261f-b9feb8f119ff"
               authors = ["Someone"]
               version = "0.0.1"

               [compat]
               julia = "1.6"
           """)
       end

julia> run(`git init`)
Initialized empty Git repository in ~/.julia/dev/BrokenModule/.git/
Process(`git init`, ProcessExited(0))

julia> run(`git add .`)
Process(`git add .`, ProcessExited(0))

julia> run(`git commit -m 'initial commit'`)
[master (root-commit) 78b2121] initial commit
 3 files changed, 16 insertions(+)
 create mode 100644 Project.toml
 create mode 100644 src/BrokenModule.jl
 create mode 100644 src/broken_a.jl
Process(`git commit -m 'initial commit'`, ProcessExited(0))

julia> Pkg.add(url=pwd())
    Updating git-repo `~/.julia/dev/BrokenModule`
   Resolving package versions...
    Updating `/tmp/jl_D95G2q/Project.toml`
  [79e551b4] + BrokenModule v0.0.1 `~/.julia/dev/BrokenModule#master`
[...]

PR #939, opened 2025-07-24
Preliminary support for `include(mapexpr, file)`

### LINKED ISSUE #820: Revise and DispatchDoctor
I have been playing around with https://github.com/MilesCranmer/DispatchDoctor.jl/ (v0.4.2) and found it quite useful. It also seems to work well with `Revise.includet`:
```
shell> cat test.jl
@stable f(x) = x == 1 ? 1 : 1.0

julia> Revise.includet("/Users/thofma/test2.jl")

julia> f(1)
ERROR: TypeInstabilityError: Instability detected in `f` defined at test.jl:1 with arguments `(Int64,)`. Inferred to be `Union{Float64, Int64}`, which is not a concrete type.
```
 If I then change test.jl, i.e. change `f` from unstable -> stable, things are properly revised (even changing `f` from unstable  -> unstable works.)

But it does not seem to work well within a module. If I have a module with the following content:
```
module A
using DispatchDoctor
include("test.jl")
end
``` 
Now if I do `using Revise, A` etc, and change `f` from unstable -> unstable, the error is not reported anymore.

As far as I understand, DispatchDoctor.jl replaces `A.include` with https://github.com/MilesCranmer/DispatchDoctor.jl/blob/569aac2fc3a904aa6e499a6a0b406aaa3a3ed849/src/stabilization.jl#L196-L205, and it seems this does not work properly in connection with Revise.

I am not sure how realistic it is to have both packages work nicely together. So feel free to close this issue, if it is technically not possible.

CC: @MilesCranmer

timholy:
In conjunction with [`CodeTracking.MapExprFile`](https://github.com/timholy/CodeTracking.jl/pull/144), this allows Revise to track
changes in files that are included with `include(mapexpr, file)`. This
PR supports the implementation of revisions for such files, properly
applying the `mapexpr` transformation to the included file's
expressions.

What's missing:
- Revise can't correctly populate the internal data structures needed to
  initiate `mapexpr` without help. The problem is that Revise only
  parses files that get edited, but discovering that `file2.jl` gets
  included from `file1.jl` via a `mapexpr` requires parsing `file1.jl`,
  which may not have been edited and therefore never scanned via Revise.
  The right way to fix this is in Julia itself, writing additional info
  to the package cache header. This will not land before Julia 1.13.
- `include(mapexpr, file)` is well-suited to include the same file more
  than once with different `mapexpr`s. While it hasn't been tested, this
  seems unlikely to work for now, and the anticipated problems stem all
  the way back to Julia itself (again, issues that should be fixed in
  the package cache header).

Package authors who want to use this in their packages can add a call to
`Revise.parseall(@__MODULE__)` in their `__init__` function to
ensure that the `mapexpr` files are parsed and tracked by Revise.

Fixes #820, #634

Note this is built on #904 so the diff will be confusing unless you look at only the final commit.

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -11,6 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - run: julia --project -e 'using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/serenity4/CodeTracking.jl#parametrize-by-mt https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"'
+      - run: cd docs && julia --project -e 'using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/serenity4/CodeTracking.jl#parametrize-by-mt https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"'
       - uses: julia-actions/julia-buildpkg@latest
       - uses: julia-actions/julia-docdeploy@latest
         env:
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: julia --project -e 'using Pkg; pkg"add https://github.com/serenity4/JuliaInterpreter.jl#codetracking-v2 https://github.com/serenity4/CodeTracking.jl#parametrize-by-mt https://github.com/serenity4/LoweredCodeUtils.jl#support-external-methodtables"'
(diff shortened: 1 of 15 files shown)


serenity4:
That appears much less intrusive than what I started at https://github.com/timholy/CodeTracking.jl/pull/142#issuecomment-2989392306, that's great!

IMHO the design remains somewhat less intuitive (treating the `mapexpr`-included files as a new kind of file), as opposed to an include-centric architecture (where we actually care about inclusions, not files directly - only indirectly). But given that for standard uses (like 99% of the time) we don't care about the distinction, I believe that keeping Revise file-centric is better and supporting `mapexpr` inclusions as a kind of file is a very good workaround.

alexandergunnarson:
@serenity4 curious whether this is mergeable so we can get [DispatchDoctor](https://github.com/MilesCranmer/DispatchDoctor.jl) working with Revise! 🎉

serenity4:
We decided together with @timholy to go ahead with the approach started in https://github.com/serenity4/Revise.jl/commit/b169d6e0e81dd929a38779570e361fbe12fc8537 instead, but I haven't had time to finish this up yet. It is still on my radar though.

alexandergunnarson:
Sounds good — thank you!!

timholy:
In principle I'm also happy to help, though time has been short lately. Originally I was hoping to get this in before switching to focusing on 1.12, but at this point I'm abandoning that idea. #894 is my main priority until it's done.

alexandergunnarson:
Understood @timholy — fixing #894 will be a game-changing improvement for Revise.jl.

timholy:
Superseded by #1102

PR #941, opened 2025-07-26
Remove manifest from docs & bump version

serenity4:
I had made the changes locally, and while the Documenter CI happened to also succeed in #904, I actually forgot to commit these :facepalm: 

This also bumps the patch version, to release #904 along with #909 and a few other changes, which I believe are essentially a set of non-breaking bug fixes.


serenity4:
@timholy gave his approval to merge the changes to the docs project and to release once I made sure nothing appears to be breaking, so I'll go ahead with merging and tagging a new version.

PR #942, opened 2025-07-28
Improve the exit-check

### LINKED ISSUE #887: loading revise on 1.11 releases leads to very slow exiting of REPL
I'm having a strange issue on 1.11.2 and 1.11.3 that when I load Revise (3.7.2) in a REPL (directly via `using` or via the recommended startup.jl try/catch block) I find exiting the REPL can take as much as 40 seconds. If I don't load revise I do not experience any lag. This happens even if I just open a REPL, `using revise`, and `exit()`. 

I have a M3 Max laptop on MacOS 14.6.1.

Are there any suggestions on how I can avoid this issue?

timholy:
Recent versions of Julia have changed the expression to

```
julia> 1+1
:($(Expr(:toplevel, :(#= REPL[5]:1 =#), quote
    $(Expr(:softscope, true))
    1 + 1
end)))
2
```

Fixes #887

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1233,15 +1233,17 @@ end
 # reduces the overhead of using Revise.
 function revise_first(ex)
     # Special-case `exit()` (issue #562)
-    exu = unwrap(ex)
-    if isexpr(exu, :block, 2)
-        arg1 = exu.args[1]
-        if isexpr(arg1, :softscope)
-            exu = exu.args[2]
+    if isa(ex, Expr)
+        exu = unwrap(ex)
+        if isexpr(exu, :block, 2)
+            arg1 = exu.args[1]
+            if isexpr(arg1, :softscope)
+                exu = exu.args[2]
+            end
         end
-    end
-    if isa(exu, Expr)
-        exu.head === :call && length(exu.args) == 1 && exu.args[1] === :exit && return ex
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Thanks for the fix!! (Maybe worth having a regression test in case it changes again? Though I know I'm asking for extra work... Just might be useful to prevent needing to debug this in the future. Might end up saving you more time...)

timholy:
It's not easy to test...I'd have to set up a fake REPL. Contributions welcome.

timholy:
If you do ever notice it regressing, just add `display(ex)` to `revise_first` and see what changed, then submit a PR.

MilesCranmer:
I think the Linux `expect` tool is great for this kinda thing. For example one could solve this with:

1. an environment variable set by the tests,
2. => read as a static `const DEBUGGING_REPL_EXIT = get(ENV, ...)`,
3. => enable some kind of tracking around the exit code that gets removed by the compiler otherwise

Then, a REPL test could use Linux expect to: (1) boot up the REPL, then once the prompt appears; (2) trigger an enormous number of code changes (somehow); (3) write "exit()" in the REPL and time (?) the exit time. Or something

In any case, I'm also facing  an extreme time deficit so I suppose we shall need to wait for a hero to see this thread and take it up...

PR #944, opened 2025-07-28
Add pretty-printing for logs

timholy:
The logs are hard to read. In preparation for #894 (which seems likely
to cause problems), make them easier to understand while also preserving
all the info.


timholy:
I don't understand the Documenter failure, and cannot currently replicate it locally. (I did see it when I didn't run it on julia 1.12.) @fredrikekre, any ideas?

timholy:
Oh, for the love of whitespace (85a8b2e). Problem fixed (I hope).

PR #946, opened 2025-07-31
remove no longer used `methodinfo` argument from `minimal_evaluation!`

aviatesk:


aviatesk:
The failures on 1.12 would be fixed by https://github.com/JuliaLang/julia/pull/59182, which is already backported.

timholy:
This seems to be the first commit that has a scary Documenter test failure. Is this the cause or is it something else?

aviatesk:
No worries, the documentor failure also seems to be fixed by https://github.com/JuliaLang/julia/pull/59182 (not sure why we suddenly started to see this error, which is really scary though)

PR #955, opened 2025-09-03
Docs: Suggest `import Revise` (instead of `using Revise`) for most users

DilumAluthge:
It seems like `import Revise` is probably sufficient for most users, since most people probably aren't using `includet()`.

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -13,7 +13,7 @@ Users should weigh this penalty against whatever benefit they may derive from no
 This can be as simple as adding
 
 ```julia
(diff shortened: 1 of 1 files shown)


timholy:
I wish that were true, but I'm not certain. ~15% of all issues mention `includet`.

I also like `using Revise: Revise` better than `include Revise` since the latter allows you to add method specializations without qualifications, but that's a minor point.

Let's see if others weigh in on this issue.

timholy:
I think we'll keep it the way it is.

PR #957, opened 2025-09-25
Add test showcasing failure when removing then adding back type parameters

lassepe:
This PR adds a test for the failure case reported in https://github.com/timholy/Revise.jl/pull/894#issuecomment-3279832342

Note that this PR is against https://github.com/timholy/Revise.jl/pull/894, not against `master`.


timholy:
Thanks!

PR #959, opened 2025-10-04
Force revise() after Revise.track

IanButterworth:
Fixes
```
% ./julia -E 'using Revise; Revise.track(Base); Base.Threads.foo()' 
ERROR: UndefVarError: `foo` not defined in `Base.Threads`
```

PR
```
% ./julia -E 'using Revise; Revise.track(Base); Base.Threads.foo()'
1
```

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -1,14 +1,14 @@
 """
-    Revise.track(Base; revise_throw::Bool=false)
-    Revise.track(Core.Compiler; revise_throw::Bool=false)
-    Revise.track(stdlib; revise_throw::Bool=false)
+    Revise.track(Base; revise_throw::Bool=!isinteractive())
+    Revise.track(Core.Compiler; revise_throw::Bool=!isinteractive())
+    Revise.track(stdlib; revise_throw::Bool=!isinteractive())
 
 Track updates to the code in Julia's `base` directory, `base/compiler`, or one of its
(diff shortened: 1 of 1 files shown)


timholy (approved):

IanButterworth:
@timholy I realised in the examples above it's better for the revision to throw if there are issues, so I added a kwarg.

IanButterworth:
Maybe the default for that kwarg should be `!isinteractive()`...

KristofferC:
I didn't get why you would not want to revise after tracking?

IanButterworth:
Sorry I don't follow.

KristofferC:
Sorry, nevermind, I was confused about what the kwarg actually did.

IanButterworth:
I'd be open to removing the kwarg and just making it `revise(throw=!isinteractive())`

timholy:
Can you expand on what's triggering the throwing? Bad code? Something else? In the OP I'm assuming the error stems from the command (`Base.Threads.foo()`) but that the keyword is targeted at something else?

IanButterworth:
Yeah, bad code. 

Previously 

- Edit code badly i.e. syntax errors. 
- Run `Revise.track(Base); include("test.jl")`
- revise will soft fail, logging the parsing error and then continue to run test.jl with the un-revised Base, which you then have to interrupt

It's better for `revise` to throw.

timholy:
https://github.com/JuliaRegistries/General/pull/139976

PR #962, opened 2025-10-21
implement basic extension system for `pkgdatas`

aviatesk:
We are currently implementing an incremental analysis system for JET and JETLS by integrating them with Revise. We considered other approaches such as implementing their own Revise-like system on the JET[LS] side (e.g. timholy/Revise.jl#938), but concluded that implementing incremental analysis by making Revise itself extensible would be better in terms of implementation reliability and future maintenance costs.

This commit implements the first step towards this goal by extending Revise's most important data structure `pkgdatas` to hold custom data and providing utilities for interacting with it. Specifically, this commit extends the leaf nodes of the `pkgdatas` data structure, `MethodInfoKey(method_table => signature)`, to a new data structure `SigInfo(method_table, signature, ext_data::ExtendedData)`.

`SigInfo` is defined as follows:
```julia
struct SigInfo
    mt::Union{Nothing,MethodTable}
    sig::Type
    ext::ExtendedData
end

struct ExtendedData
    owner::Symbol
    data::Any
    next::ExtendedData
    ExtendedData(owner::Symbol, data, next::ExtendedData) = new(owner, data, next)
    ExtendedData(owner::Symbol, data) = new(owner, data)
    ExtendedData() = new(:Revise, nothing)
end
```

As seen from the type definition, `SigInfo` can hold `ExtendedData` in addition to the information that the conventional `MethodInfoKey` holds. `ExtendedData` is implemented as a linked list, enabling safe data management per owner through the `owner` field (similar to `Core.CodeInstance` design).

We also implement the following utility functions for getting or replacing `ext_data`:
- `get_extended_data(ext::ExtendedData, owner::Symbol) -> ext::Union{ExtendedData,Nothing}`
- `get_extended_data(siginfo::SigInfo, owner::Symbol) -> ext::Union{ExtendedData,Nothing}`
- `replace_extended_data(ext::ExtendedData, owner::Symbol, data) -> new_ext::ExtendedData`
- `replace_extended_data(siginfo::SigInfo, owner::Symbol, data) -> new_siginfo::SigInfo`

This commit implements these simple extension setup only. To actually use this extension system, external packages need to decide when to inspect `Revise.pkgdatas` by themselves and use functions like `replace_extended_data` to hold custom data. While packages using this extension system and Revise's revision pipeline are not tightly synchronized, stale `SigInfo` objects are automatically removed, enabling minimal external incremental systems. Specifically, using this commit, we can replace `JET.report_package` with a Revise-based implementation and make it incremental (aviatesk/JET.jl#763).

In the future, we may need to set up some callbacks in Revise's revision pipeline to allow external packages to customize in sync with Revise's revision pipeline. However, since the necessary requirements are still unclear at the current stage of extending `JET.report_package`, this commit does not implement them.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Revise"
 uuid = "295af30f-e4ad-537b-8983-00126c2a3abe"
-version = "3.10.0"
+version = "3.11.0"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
--- a/docs/src/dev_reference.md
+++ b/docs/src/dev_reference.md
@@ -17,6 +17,7 @@ Revise.tracking_Main_includes
 
 ```@docs
 Revise.juliadir
+Revise.expected_juliadir
 Revise.basesrccache
 Revise.basebuilddir
 ```
@@ -152,4 +153,13 @@ Revise.git_repo
 
 ```@docs
 Revise.init_worker
+```
+
+## Extension system
+
+```@docs
+Revise.SigInfo
+Revise.ExtendedData
+Revise.get_extended_data
+Revise.replace_extended_data
 ```
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -10,10 +10,10 @@ end
 # This default version is simple and only used for testing purposes.
 # The "real" one is CodeTrackingMethodInfo in Revise.jl.
 const MethodInfo = IdDict{MethodInfoKey,LineNumberNode}
-add_signature!(methodinfo, @nospecialize(sig), ln) = push!(methodinfo, sig=>ln)
-push_expr!(methodinfo, mod::Module, ex::Expr) = methodinfo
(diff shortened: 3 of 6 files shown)


aviatesk:
Alright, I've completed all the necessary changes on the aviatesk/JET.jl#763 side, which is the first user of this feature.
@timholy I'm afraid for the long description in this PR, but if you have time, I would appreciate your comments.

On the other hand, as mentioned in aviatesk/JET.jl#763, moving forward with the integration of JET and JETLS with Revise will require fairly significant refactoring, and considering these factors, I also have a desire to speed up the merge cycle for this PR and future ones I will submit to this repository.

So, while I think this refactor should ideally wait for your opinion, if there's no response by tomorrow, I would like to take responsibility and merge this change myself.
I will also rebase `teh/revise_struct` branch accordingly after merge.
At least for this PR, all test cases are passing, and given the nature of the code, I don't believe it will cause any performance degradation.

PR #963, opened 2025-10-21
Update limitations section in documentationConst/struct redefinition …

PallHaraldsson:
…possible(?)

### DIFF
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -2,7 +2,6 @@
 
 There are some kinds of changes that Revise (or often, Julia itself) cannot incorporate into a running Julia session:
 
-- changes to type definitions or `const`s
 - conflicts between variables and functions sharing the same name
 - removal of `export`s
 


PallHaraldsson:
I believe consts were the last thing Julia couldn't redefine, until 1.12, so I assume holds with revise too. Feel free to change this PR, to "as of 1.12", but also since I think this was forgotten, ar the other limitations still valid:

https://timholy.github.io/Revise.jl/stable/limitations/
* conflicts between variables and functions sharing the same name
* removal of exports

And while I'm here, I'm not in the habit of using Revise, though I should be... I was looking a bit into off by default mode:
https://timholy.github.io/Revise.jl/stable/config/#Configuring-the-revise-mode

since I saw a tip for a new Julia user. But should this or some more be on by default?

I'm not sure where that came from:
>Clarified limitations regarding changes that require a session restart.

My chosen title got also changed... and this added.

PallHaraldsson:
I though this limitation might have been simply gone, and just forgotten. I now see in the 1.12 blog post:

>There is also work in progress in Revise.jl to automatically redefine functions on replaced bindings.

Maybe docs here should be clarified? I believe consts can be redefined already (without or even with Revise), Functions are also sort of "consts" so that by now the only exception? I'm not sure how do word this, and since this isn't simply forgotten I think I^'ll close here and leave to others.

PR #967, opened 2025-10-31
Update syntax highlight tags for example code blocks

abhro:
Mostly add `julia-repl` for more specific highlighting.


aviatesk (approved):
LGTM. Thanks!

PR #970, opened 2025-11-19
Making Revise relocatable

JanisErdmanis:
Currently, Revise hardcodes paths at compilation time, which prevents its pkgimage cache from being relocatable. This creates issues when forming Julia distributions like Jumbo that need relocatable package caches.

This PR addresses the issue by converting the hardcoded path constants to uninitialized globals that are assigned in the `__init__` function instead. Limited testing shows that the relocated Revise works fine with this approach.

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -262,7 +262,7 @@ See also [`Revise.silence`](@ref).
 """
 const dont_watch_pkgs = Set{Symbol}()
 const silence_pkgs = Set{Symbol}()
-const depsdir::String
+global depsdir::String
 const silencefile = Ref{String}()  # Ref so that tests don't clobber
 
 ##


src/packagedef.jl
```diff
@@ -1345,6 +1345,13 @@ function __init__()
         return nothing
     end
 
+    # Setting up the paths relative to package module location
+
+    global juliadir = find_juliadir()
+    global basesrccache = normpath(joinpath(expected_juliadir(), "base.cache"))
+    global depsdir = joinpath(isnothing(pkgdir(@__MODULE__)) ? dirname(@__DIR__) : pkgdir(@__MODULE__), "deps")
+    silencefile[] = joinpath(depsdir, "silence.txt")
```
KristofferC: Tangentially, the whole `silence` functionality should be rewritten using Preferences.jl instead of mutating the package folder.

JanisErdmanis: I am currently overwhelmed and refactoring to use Preferences.jl is not on my plate, but I support the change.

timholy:
Thanks!

timholy:
https://github.com/JuliaRegistries/General/pull/143000

PR #974, opened 2025-11-26
Adjust for base paths possibly being non-normalized

Keno:
Fixes test failures with https://github.com/JuliaLang/julia/pull/60251. Also fixes an unrelated test where Revise was using the wrong PkgId for Core.Compiler (failure not seen on CI, because it's explicitly disabled there).


Keno:
I'm not entirely happy with this in the sense that I don't think we have a principle'd answer as to how we're treating paths, but it's passing tests on both the old and the new version, which I think is good enough for now. This is one of those topics that I want to revisit after JuliaLowering.

PR #977, opened 2025-12-19
Remove Requires and Unicode as dependencies

JamesWrigley:
Requires is never actually loaded by Revise, Revise only defines a function that can be called by Requires. Unicode is also not used at all.


aviatesk (approved):
Thanks!

PR #978, opened 2025-12-19
Move REPL support into a package extension

JamesWrigley:
The main reason for this is to make Revise a bit more light-weight for packages that only need the revision API, and to reduce the overhead of loading Revise by default outside of the REPL. The load time when loaded without the REPL is almost halved:
```shell-session
# Master
$ julia --startup-file=no --project -e '@time import Revise'                                                                                                                                                           
  0.319723 seconds (404.29 k allocations: 27.255 MiB)

# PR
$ julia --startup-file=no --project -e '@time import Revise'                                                                                                                                                             
  0.182584 seconds (314.75 k allocations: 21.336 MiB)
```

I also renamed the Distributed extension in 8d4a07f to make it easier to distinguish where the extension module comes from.
Written with help from Claude :robot:

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,14 +14,15 @@ LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
 LoweredCodeUtils = "6f1432cf-f94c-5a45-995e-cdbf5db27b0b"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
-REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [weakdeps]
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 
 [extensions]
 ReviseDistributedExt = "Distributed"
+ReviseREPLExt = "REPL"
 
 [compat]
 CodeTracking = "3"
@@ -31,6 +32,7 @@ JuliaInterpreter = "0.10.8"
 LoweredCodeUtils = "3.5"
 OrderedCollections = "1"
 Preferences = "1.5.0"
+REPL = "1"
 # Exclude Requires-1.1.0 - see https://github.com/JuliaPackaging/Requires.jl/issues/94
 Requires = "~1.0, ^1.1.1"
 julia = "1.10"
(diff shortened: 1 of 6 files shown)


JamesWrigley:
(rebased to fix merge conflicts)

JamesWrigley:
(bump)

JamesWrigley:
(bump)

aviatesk:
Loading Revise even when not starting a REPL seems to be a startup-file issue on the user side. So I agree with the optimization in this PR, but I have concerns about whether the added complexity required to support package extensions is actually justified by the benefits.

JamesWrigley:
My motivation isn't actually the startup time, it's making Revise more lightweight to use in other libraries that don't need the REPL (i.e. mine :stuck_out_tongue:). But I do understand the complexity concern. Having to load the REPL would be unfortunate but not a deal breaker for me, so I leave the final decision to you :)

aviatesk:
Actually, I need to do something similar to your use case (calling Revise from a library that doesn't need REPL). Specifically, I'll be calling Revise's subcomponents from JETLS, so I'll need this kind of mechanism in the But if we use the package extension method, these codes would be automatically loaded by packages that load REPL no matter if the package want to use this REPL extension, which might cause a different problem. Or, maybe we should just refactor Revise subcomponents into something like ReviseCore.jl, by defining reasonable subcomponents of Revise that can be used independently. This change wouldn't affect its design.

JamesWrigley:
Having a ReviseCore also sounds good to me :+1: I'm happy to take a shot at it if you prefer that over a package extension.

JamesWrigley:
Shall I make a ReviseCore?

timholy:
Sorry for my inattention here.

I worry that what might be needed by ReviseCore would need very careful scoping. I will note that the Makie developers recently deprecated MakieCore and eliminated its use in Makie 0.24 because there was no clear boundary between what went in the core and what was in the rest. One can argue that CodeTracking is already "the core" in the sense that it holds key data and permits other libraries to make queries of Revise data.

If you can't solve your concerns by using CodeTracking, then we can talk about what you actually need. But until that's clear I'm mildly opposed to splitting out a ReviseCore.

timholy:
tagging this "DO NOT MERGE" just because I'm starting to use agents to help with maintenance and I don't want them doing something stupid! ("I see CI is green, so I'll merge!")

JamesWrigley:
I see, thanks for the info. In this case I was thinking that ReviseCore would be literally everything except the REPL bits, so Revise would end up being ReviseCore + the code that's in the REPL extension in this PR. Would that be a good boundary?

timholy:
I guess it's one that can't be argued with functionally. But if that's the boundary, then I wonder whether going back to the extension idea is the better way to go. @aviatesk, since you're the one who brought up ReviseCore, what are your thoughts on this issue?

KristofferC:
FWIW, I approve of the direction of this PR (I was going to make the same PR and saw this). Various packages depend on Revise for various reasons but there is no point in forcefully loading REPL for these. Pkg also has a REPL extension for this reason.

JamesWrigley:
Happy to resurrect it :) I don't have a strong opinion on whether it should be a package extension or separate package.

PatrickHaecker:
Paralleling Pkg sounds like the right thing to do, except if there are reasons not to do it that way.

JamesWrigley:
(rebased to fix merge conflicts)

PR #992, opened 2026-01-08
preference to disable initial cache of fieldtypes(#991)

### LINKED ISSUE #991: unexpected background thread after Revise 1.13.1
After Revise 1.13.1, due to https://github.com/timholy/Revise.jl/pull/987, loading the package causes a background thread to run for a long time. This impacts my performance benchmarks even though I don't explicitly load Revise (it's loaded by JET).
Would you consider a PR adding an env var to disable loading the cache?

thanks

dpinol:
Fixes #991 adding a Preference.jl optional to disable the initial fieltypes cache population

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -185,3 +185,15 @@ string `"1"` (e.g., `JULIA_REVISE_POLL=1` in a bash script).
     NFS stands for [Network File System](https://en.wikipedia.org/wiki/Network_File_System) and is typically only used to mount shared network drives on *Unix* file systems.
     Despite similarities in the acronym, NTFS, the standard [filesystem on Windows](https://en.wikipedia.org/wiki/NTFS), is completely different from NFS; Revise's default configuration should work fine on Windows without polling.
     However, WSL2 users currently need polling due to [this bug](https://github.com/JuliaLang/julia/issues/37029).
+
+
+### Disabling Automatic Fieldtype Caching
+
+When Revise is loaded, it spawns a background task to cache all fieldtypes. This process ensures faster code reloading later but can introduce several seconds of initial startup overhead.
+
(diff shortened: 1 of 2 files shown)


dpinol:
> Can we document this somewhere?

done, thanks

timholy:
I'm starting to make steps towards "coming back," but I'm hesitating a bit on this. Once we merge this we need to stick with it until the next breaking release.

I think there's some question whether the better thing to do is have an option to disable type revision. Without a pre-populated cache, type revision is so slow that it's often faster to just quit the session. If we just do it this way, then if you change any types it will start on the slow task of populating the cache.

JamesWrigley:
I don't have a strong opinion on whether the preference should be on the type cache or type revision entirely, but I'm definitely +1 on having some option that ends up disables the caching. It's kinda annoying to have Revise hog a thread for 30s every time Julia starts.

lassepe:
+1 for allowing to disable struct vision all together

aviatesk:
We had a segfault issue with this part on v1.12 and Apple Silicon environments, as reported in https://github.com/JuliaLang/julia/issues/60721 (due to a problem on the Julia side, just to be clear). For now, wIve added a fix in 3.12.2 to disable this pre cache-population when it's v1.12 and `Sys.isapple()`.

lassepe:
I think it's still good to have an option to disable struct revision until the performance issues are resolved

JamesWrigley:
Gentle bump, would be great to have an option for this.

aviatesk:
I think it's okay to add this option, as far as we put an explicit "experimental" warning on this preference, and make it clear that this configuration might change or even be removed without announcement, depending on our plans for struct revision-related issues.

pfitzseb:
Should this be an opt-in instead, given that the performance issues can be fairly severe?

timholy:
I went with #1008 instead

PR #1008, opened 2026-02-11
Add a preference for enabling struct revision

JamesWrigley:
This disables it by default since the overhead is quite high. It's a slightly more invasive alternative to #992. Also fixed some links in 8a93737.

BTW since this doesn't really need to be a compile-time preference I would like to use this PR in the future once it's merged and released: https://github.com/JuliaPackaging/Preferences.jl/pull/90.

Made with help from Claude :robot:


JamesWrigley:
(bump)

JamesWrigley:
Bump, rebased to fix merge conflicts.

timholy:
Thanks for doing this!

timholy:
I'll fix it so that the tests run with `__bpart__[] = true`

PR #1012, opened 2026-02-24
test: ensure struct revision tests run on CI and test the preference mechanism

timholy:
The revise_structs preference defaults to false, so all struct revision tests were silently skipped on CI. Fix by forcing `__bpart__[] = true` before those tests on supported Julia versions.

Add subprocess-based tests for the LocalPreferences mechanism, since the preference is read in __init__ and cannot be tested within the same process.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ docs/site
 *.log
 docs/src/figures/diagram.pdf
 Manifest.toml
+Manifest-v*.toml
 src/aliasnewstruct.jl
 .vscode
 test/envs/use_exputils/Project.toml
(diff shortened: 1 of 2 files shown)


timholy:
CC @JamesWrigley (xref #1008)

JamesWrigley:
Oops thanks, didn't realize that. The nightly failure looks new, I'm guessing something changed in Julia? On #1008 nightly did pass.

timholy:
It seems the `entr` test is failing on master for every PR. Since it's not caused by this PR, we can merge.

Issue #988, opened 2026-01-04
Struct revision takes several minutes for packages with large type trees


lassepe:
<details><summary>Original (but outdated) issue description</summary>
I have a medium-sized project (5k LOC) and noticed that, since #894 was merged,  some revisions cause Revise to lock up (I'm not sure if it will terminate eventually because I've terminated it after 5min). In my project, I can most reliably trigger this situation by changing some default values in a `@kwdef` struct, e.g., changing

```julia
Base.@kwdef struct Foo
  a::Int = 1
end
```

to

```julia
Base.@kwdef struct Foo
  a::Int = 2
end
```

Unfortunately, this change in isolation does not trigger the issue and I have yet to find a good way to isolate a compact MWE. Still, I thought it's worth opening this issue in case others observe similar issues or have ideas for how to narrow this down.
</details>

## Updated Issue Description

I have a medium-sized project and noticed that, since #894 was merged,  revisions cause Revise to take very long. My project itself has only about 5k LOC but it has a few heavy dependencies (including CUDA and Plots.jl). I added some logging to Revise (see https://github.com/timholy/Revise.jl/issues/988#issuecomment-3719297218) and found that `collect_all_subtypes(Any)` reaches 62k subtypes (almost all of which are from dependencies). Indexing those takes about 16min.

Looking at those logs, it looks like the initial cache that is built on the background thread hits only about 13k types (because it is loaded before everything else). Upon the first change to any code, we hit `collect_all_subtypes(Any)` again which now has to index another ~49k subtypes reachable through the newly loaded (transient) dependencies.

So, in sum, this slowness seems to be the compounding effect of three issues:

1. Changes of default arguments in `@kwdef` trigger struct revision (not just traditional non-struct revision)
2. `alltypes` gets way too large (perhaps because it also includes all subtypes of `Function`)
3. The initial call to `collect_all_subtypes` hits only a small portion of the type tree because is typically loaded before everything else

alexandergunnarson:
FWIW I've also found this to be the case when changing types in a struct. I have a reasonably large project, and it takes at least as long as precompilation does (45 seconds). Probably more like ~3 minutes.

lassepe:
@alexandergunnarson does it eventually terminate for you? I didn't even see it terminate in my case and I waited considerably longer than the precompile time for that package

alexandergunnarson:
[...]

PR #1013, opened 2026-02-25
Speed up `fieldtypes_cached`

timholy:
xref #988

### DIFF
--- a/src/visit.jl
+++ b/src/visit.jl
@@ -63,7 +63,7 @@ function fieldtypes_cached(@nospecialize(type))
         # the equivalent `get!(types_cache, type) do ... end` form is not used because on 1.12 it triggers recompilation
         cache = get(types_cache, type, missing)
         cache !== missing && return
-        types_cache[type] = cache = fieldtypes_array(type)
+        types_cache[type] = fieldtypes_array(type)
         return
     end
 end


timholy:
@lassepe, any chance you can try this branch? On my machine it speeds up 1.12 quite substantially even without https://github.com/JuliaLang/julia/pull/61156

lassepe:
Nice, with julia 1.12.5 and this PR, revision time in my example from https://github.com/timholy/Revise.jl/issues/988#issuecomment-3846363436 is down to 20s (from >5min). I would still not consider #988 fixed with this but it is a huge improvement

lassepe:
https://github.com/JuliaLang/julia/pull/61156https://github.com/JuliaLang/julia/pull/61156 has all sorts of weird precompilation errors for me so I cannot test with that combination (which may be due to the fact that my MWE uses GPU-related packages that don't currently work on julia master)

timholy:
Great! I just tried your example and saw similar speedup (>800s to 11s).

Do you think type-revision should still be off by default? It's not clear that https://github.com/JuliaLang/julia/pull/61156 will make it any faster on top of this (it's still a useful change, though). So if it needs to be faster to be viable, we will need additional optimizations. https://github.com/JuliaLang/julia/pull/60736 may provide a path forward (one could potentially restrict to names created after the revised type), but it may take a while to get backported to 1.12.

lassepe:
I would want to do more tests in a real-world use case to make that call. My guess is that it can be enabled by default but that we shouldn't spawn a background thread to build the cache. I think many people implicitly rely on Revise being close to zero-overhead when they don't edit any code (so that it is not devastating to have Revise loaded during "production") but that stops being true if we have a background thread.

timholy:
Well, a background thread that completes its work fairly quickly would be less intrusive, no?

PR #1014, opened 2026-02-26
Update type-cache after each user interaction

### LINKED ISSUE #988: Struct revision takes several minutes for packages with large type trees

timholy:
The intent here is to keep the type-cache up to date, so that if and when a type is redefined, there isn't a large latency before the revision can start.

This is the last piece of the puzzle needed to
Fix #988

Note this also re-enables type revision by default. CC @lassepe, @JamesWrigley . If you haven't followed, #1013 yielded a dramatic speedup. This PR should keep the cache up-to-date after each REPL command, which I hope will mean that it will be quite rare to hit long latencies on type-revision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
                                            Base.run_main_repl(true, true, false, true, false))
           isdefined(Base, :errormonitor) && Base.errormonitor(t)
           while (!isdefined(Base, :active_repl_backend) || isnothing(Base.active_repl_backend)) sleep(0.1) end
-          pushfirst!(Base.active_repl_backend.ast_transforms, Revise.revise_first)
+          pushfirst!(Base.active_repl_backend.ast_transforms, Revise.revise_first_scan_last)
           include(joinpath("test", "runtests.jl"))
           if Base.VERSION.major == 1 && Base.VERSION.minor >= 9
               REPL.eval_user_input(:(exit()), Base.active_repl_backend, Main)
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1483,12 +1483,9 @@ function maybe_set_prompt_color(color)
     return nothing
 end
 
-# `revise_first` gets called by the REPL prior to executing the next command (by having been pushed
+# `revise_first_scan_last` gets called by the REPL prior to executing the next command (by having been pushed
 # onto the `ast_transform` list).
-# This uses invokelatest not for reasons of world age but to ensure that the call is made at runtime.
-# This allows `revise_first` to be compiled without compiling `revise` itself, and greatly
-# reduces the overhead of using Revise.
-function revise_first(ex)
+function revise_first_scan_last(ex)
     # Special-case `exit()` (issue #562)
     if isa(ex, Expr)
         exu = unwrap(ex)
@@ -1517,8 +1514,28 @@ function revise_first(ex)
             end
         end
     end
-    # Check for queued revisions, and if so call `revise` first before executing the expression
-    return Expr(:toplevel, :($isempty($revision_queue) || $(Base.invokelatest)($revise)), ex)
+    return revise_first_scan_last_expr(ex)
+end
+
+function revise_first_scan_last_expr(ex)
+    # Modify the user-supplied expression `ex` to:
+    # 1. Check for queued revisions, and if any call `revise` first
+    # 2. Execute the user's expression and store the result in a temporary variable
+    # 3. Spawn a task to update the type cache after the expression has been executed
+    # 4. Return the result of the user's expression
+    result = gensym("result")
+    return Expr(:toplevel,
+        # This uses invokelatest not for reasons of world age but to ensure that the call is made at runtime.
+        # This allows `revise_first_scan_last` to be compiled without compiling `revise` itself, and greatly
+        # reduces the overhead of using Revise.
+        :($isempty($revision_queue) || $(Base.invokelatest)($revise)),
+        quote
+            let $result = $ex
+                Base.Threads.@spawn :default $(repopulate_typecache)()
+                $result
+            end
+        end
+    )
 end
 
 steal_repl_backend(_...) = @warn """
@@ -1633,10 +1650,10 @@ function __init__()
 
     mode = get(ENV, "JULIA_REVISE", "auto")
     if mode == "auto"
-        pushfirst!(REPL.repl_ast_transforms, revise_first)
+        pushfirst!(REPL.repl_ast_transforms, revise_first_scan_last)
         # #664: once a REPL is started, it no longer interacts with REPL.repl_ast_transforms
         if active_repl_backend_available()
-            push!(Base.active_repl_backend.ast_transforms, revise_first)
+            push!(Base.active_repl_backend.ast_transforms, revise_first_scan_last)
         else
             # wait for active_repl_backend to exist
             # #719: do this async in case Revise is being loaded from startup.jl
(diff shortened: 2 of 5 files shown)


JamesWrigley:
Very cool :tada: Slightly offtopic, but would it be possible to put that expression in a separate function? I was thinking that then IJulia could call it directly instead of reimplementing it: https://github.com/JuliaLang/IJulia.jl/blob/8e17d35ec29c89a7b51b75fd649021a94c647c3c/ext/IJuliaReviseExt.jl#L18

lassepe:
In my test case from https://github.com/timholy/Revise.jl/issues/988#issuecomment-3846363436 , populating the type cache takes 20s every time, even after the cache has already been populated and no changes to any types have been made. Hence, I'm not sure if this extra background work is worth it. Perhaps we can limit the invocation of `Revise.repopulate_typecache` to scenarios where a new package is being loaded or a new type is being defined?

timholy:
It only takes 20s because you've just loaded a ton of packages. Once they are "processed" it will be ~15ms. You can try this manually: after loading your package with all its dependencies, call
```julia
julia> @time Revise.foreach_subtype(Any) do @nospecialize type
    Revise.fieldtypes_cached(type)
end
```
and you should see it take ~20s. Then do the exact same thing again in this same session (without loading a whole bunch of new packages), and it should be very fast.

In other words, the way this fixes the slow-revision problem is that it starts scanning loaded packages in a background thread as soon as they load. That process should be finished (presuming it will take the user some to make some edits in their editor) before the next type revision occurs and the user will experience little latency.

JamesWrigley:
20s hogging a thread just by loading packages seems a bit much to enable by default :eyes:

timholy:
@JamesWrigley 

> but would it be possible to put that expression in a separate function?

Done

timholy:
> 20s hogging a thread just by loading packages seems a bit much to enable by default 👀

Fair enough. It's a pretty big package, though (not quite "OmniPackage" but getting there), and just loading it is also on par with that amount of time.

Not sure what to do about the default. I might leave it on a bit and play with it myself for a while. I still haven't abandoned 1.10 for real work but I think now is the time...

lassepe:
> It only takes 20s because you've just loaded a ton of packages. Once they are "processed" it will be ~15ms. You can try this manually: after loading your package with all its dependencies, call
> 
> ```julia
> julia> @time Revise.foreach_subtype(Any) do @nospecialize type
>     Revise.fieldtypes_cached(type)
> end
> ```
> 
> and you should see it take ~20s. Then do the exact same thing again in this same session (without loading a whole bunch of new packages), and it should be very fast.

I don't find this to be true in my tests. For testing purposes, I've modified the invocation of `repopulate_typecache` in `revise_first` to track timings:

```julia
    return Expr(:toplevel,
        :($isempty($revision_queue) || $(Base.invokelatest)($revise)),
        quote
            let $result = $ex
                # Base.Threads.@spawn :default $(repopulate_typecache)()
                t = time()
                println("Repopulating type cache...")
                $repopulate_typecache()
                println("Repopulation done after ", round(time() - t, sigdigits=3), " seconds.")
                $result
            end
        end
    )
```

With those changes, here's what I get with the test package from https://github.com/timholy/Revise.jl/issues/988#issuecomment-3846363436

```julia
❯ JULIA_LOAD_PATH="@." julia --startup-file=no
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.12.5 (2026-02-09)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org release
|__/                   |

julia> using Revise

julia> @time Revise.foreach_subtype(Any) do @nospecialize type
           Revise.fieldtypes_cached(type)
       end
  0.577651 seconds (245.60 k allocations: 176.847 MiB, 2.10% gc time, 2.21% compilation time)
Repopulating type cache...
Repopulation done after 0.57 seconds.

julia> using Foo
Repopulating type cache...
Repopulation done after 12.4 seconds.

julia> Foo.Bar()
Repopulating type cache...
Repopulation done after 12.6 seconds.
Foo.Bar(4.0, 5)

julia> Foo.Bar()
Repopulating type cache...
Repopulation done after 12.5 seconds.
Foo.Bar(4.0, 5)
```

lassepe:
Similarly:

```julia
Revise.jl-struct-revision-kwdef on  main [!?] is 📦 v0.1.0 via ஃ v1.12.5 took 2m6s
❯ JULIA_LOAD_PATH="@." julia --startup-file=no
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.12.5 (2026-02-09)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org release
|__/                   |

julia> using Revise

julia> @time Revise.foreach_subtype(Any) do @nospecialize type
           Revise.fieldtypes_cached(type)
       end
  0.563205 seconds (245.60 k allocations: 176.633 MiB, 2.12% gc time, 2.26% compilation time)
Repopulating type cache...
Repopulation done after 0.582 seconds.

julia> @time Revise.foreach_subtype(Any) do @nospecialize type
           Revise.fieldtypes_cached(type)
       end
  0.560553 seconds (236.09 k allocations: 176.144 MiB, 2.12% gc time, 1.86% compilation time)
Repopulating type cache...
Repopulation done after 0.556 seconds.

julia> using Foo
Repopulating type cache...
Repopulation done after 12.6 seconds.

julia> @time Revise.foreach_subtype(Any) do @nospecialize type
           Revise.fieldtypes_cached(type)
       end
 12.659920 seconds (3.93 M allocations: 3.295 GiB, 3.33% gc time, 0.07% compilation time)
Repopulating type cache...
Repopulation done after 12.7 seconds.

julia> @time Revise.foreach_subtype(Any) do @nospecialize type
           Revise.fieldtypes_cached(type)
       end
 12.824380 seconds (3.93 M allocations: 3.295 GiB, 4.56% gc time, 0.08% compilation time)
Repopulating type cache...
Repopulation done after 12.6 seconds.
```

Note that I did not make any modifications to source code in any of these tests.

timholy:
Ah shoot, I must have benchmarked on a small system. (Or maybe on julia master?) Just the call to `names` takes most of the time.

OK, I agree we probably have to turn this off by default, at least until https://github.com/JuliaLang/julia/pull/60736 can be backported.

timholy:
Note to anyone who might be tempted to merge this: it's not really feasible at present, need to wait for more performance work on the Julia side.

timholy:
Superseded by #1076

PR #1017, opened 2026-03-04
release: v3.14

pfitzseb:
The opt-in for struct revision got merged last week, so we should probably tag a new version too. Not sure if only bumping the patch is appropriate, given that the defaults change, but I thought I'd open this anyways to get the ball rolling.


timholy:
I was dragging my feet because I was hoping to make it fast enough that we could make type-revision the default. But I ran out of time, so let's get this out there.

PR #1019, opened 2026-03-04
Fix preferences test

### LINKED ISSUE #1018: Preference test only works if Revise is the active project.
Make a temp environment and dev Revise then run the tests:

```
revise_structs preference: Test Failed at /Users/kc/JuliaPkgs/Revise.jl/test/runtests.jl
  Expression: read(`$julia --project=$test_proj_dir -e $check_bpart`, String) == "true"
   Evaluated: "false" == "true"
```

I am not sure of the details but I don't think the LocalPreferences.toml is picked up.

topolarity:
If the Pkg sandbox tries to install a preferences file it will prefer the name "JuliaLocalPreferences.toml", which would supersede the "LocalPreferences.toml" the test installs here.

Resolves #1018.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2471,7 +2471,7 @@ end
         @testset "revise_structs preference" begin
             # The preference is read in __init__, so we have to test it via subprocesses.
             test_proj_dir = dirname(Base.active_project())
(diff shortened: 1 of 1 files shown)


KristofferC:
CI errors on previous commit to nightly as well

Issue #341, opened 2019-08-28
Revise not picking up on "Rename" + "Change" file events


curtd:
Revise currently isn't able to handle "Renamed" + "Changed" file events, only "Changed" file events. The context for this is that I'm running Julia code in a volume-mounted Docker container on remote server, where rsync syncs the local files to the remote directory. Rsync's default behaviour is to create a temporary file on the server and then rename it to the existing server file. This causes "Renamed + Changed" events to fire on the filesystem, which are not picked up by Revise. When using rsync's `--inplace` option, the files are modified directly, causing the "Changed" event to fire and Revise to behave as expected. It would be desirable to have Revise pick up on the former behaviour as well as the latter behaviour.

![Revise Issues](https://user-images.githubusercontent.com/771028/63897648-bfbd7a80-c9aa-11e9-98ab-d6fca1463d00.png)

timholy:
My suspicion is that the write of the temporary file is updating Revise's timestamp on the directory, so that when the rename happens the file looks "old" even though it's newer than the file it replaces. Can you `dev Revise` and then make this change:
```diff
tim@diva:~/.julia/dev/Revise$ git diff
diff --git a/src/pkgs.jl b/src/pkgs.jl
index a6c3fff..7d22581 100644
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -239,7 +239,7 @@ function watch_files_via_dir(dirname)
                 push!(latestfiles, file)
             end
         end
-        updatetime!(wf)
+        isempty(latestfiles) || updatetime!(wf)
     end
     return latestfiles, stillwatching
 end
```
and see if it fixes the problem?

curtd:
That change appears to fix the problem, thank you! 
![revise_fix](https://user-images.githubusercontent.com/771028/63959089-44a6a380-ca40-11e9-9f13-ae5c3dde9a6f.png)

PR #1031, opened 2026-04-07
Use per-file ctime (fixes #1025)

timholy:
PR #1003 fixed a possible data race but re-opened issue #341. This commit adds a more rigorous test but also adds a robust fix: store the ctime of each tracked file. This eliminates the whole class of possible bugs that arise from tracking only a directory-wide modification time.

This fix allows us to continue to only monitor directories for changes (important to limit resource usage on platforms like OSX) while having the granularity to robustly detect changes to tracked files.


timholy:
@vtjnash why is `ctime` is the correct choice even on Windows? https://stackoverflow.com/a/39521489 I guess our `ctime` is change time and not creation time?

pfitzseb:
libuv abstracts over that and ensures that ctime on windows is _not_ creation time, which is instead recorded as "birth time".

PR #1036, opened 2026-05-13
Use autoreset Base.Event for revision_event (fix #837)

### LINKED ISSUE #837: Possibility of lost events due to use of edge-triggered Condition in entr(...)

timholy:
## Summary

- `revision_event` was a non-autoreset `Base.Event` with an explicit `reset(revision_event)` between `wait` and `revise()` in `entr`'s loop. A `notify` that arrives in the window between `wait` returning and `reset` is silently cleared, so the corresponding file change is not serviced until another notify happens.
- Switch to `Base.Event(true)` (autoreset, available since Julia 1.8 ≤ LTS 1.10) and drop the explicit `reset`. Pushes to `revision_queue` already happen under `revision_queue_lock` *before* `notify`, so any notify fired while `revise()` is running latches the event and wakes the next `wait`.
- Closes the lost-wakeup window described in #837 without changing external behaviour.

Fixes #837

## Test plan

- [x] `entr`, `entr with modules`, `entr with all files` testsets (21 assertions) pass.
- [x] CI green on all platforms.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -4,8 +4,10 @@
     Revise.revision_event::Base.Event
 
 This event is used to notify `entr` that one of the watched files has changed.
+It is created with `autoreset=true` so that a `notify` fired between a waiter's
+return from `wait` and its next call to `wait` is not lost — see issue #837.
 """
-const revision_event = Base.Event()
+const revision_event = Base.Event(true)
 
 """
     Revise.user_callbacks_queue
@@ -159,8 +161,7 @@ function entr(f::Function, files, modules=nothing; all=false, postpone=false, pa
     end
     try
         while true
-            wait(revision_event)
-            reset(revision_event)
+            wait(revision_event)  # autoreset; see issue #837
             revise(throw=true)
         end
     catch err
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -4114,6 +4114,26 @@ do_test("New files & Requires.jl") && @testset "New files & Requires.jl" begin
     pop!(LOAD_PATH)
 end
 
+do_test("revision_event autoreset") && @testset "revision_event autoreset (issue #837)" begin
+    # `revision_event` must be an autoreset `Base.Event`, so that `wait`
+    # clears the bit. Without autoreset, a `notify` fired between the time
+    # `entr`'s loop returns from `wait` and the time it would call
+    # `reset(revision_event)` is silently dropped, and the corresponding
+    # file change is not serviced until another notify happens.
+    ev = Revise.revision_event
+    @test ev isa Base.Event
+    # Drain any stale state left by earlier testsets.
+    notify(ev); wait(ev)
+    # After `wait`, the bit must be cleared: a second wait with no
+    # intervening `notify` must block. Without autoreset it would return
+    # immediately because the bit set by the previous `notify` would still
+    # be latched.
+    t = @async wait(ev)
+    @test timedwait(() -> istaskdone(t), 0.2; pollint=0.02) === :timed_out
+    notify(ev)
+    wait(t)
+end
+
 do_test("entr") && @testset "entr" begin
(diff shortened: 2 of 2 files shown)


timholy:
Some thoughts on testability, since race tests are notoriously hard to write robustly:

**Option A — pin the API choice (easy, shallow).**

```julia
@testset "revision_event uses autoreset (issue #837)" begin
    # The bit must auto-clear on `wait`, so a stale notify cannot leak
    # into the next iteration of entr's loop.
    notify(Revise.revision_event)
    wait(Revise.revision_event)            # autoreset → bit cleared here
    @test timedwait(() -> false, 0.05; pollint=0.05) === :timed_out
    @test_throws Exception wait(Revise.revision_event; timeout=...)  # would block
end
```

Pure inspection. Catches an accidental revert to `Base.Event()` or the re-introduction of an explicit `reset(revision_event)` (the latter would not type-error but would re-open the race). It does **not** prove the original bug is gone — only that the chosen mitigation is in place.

**Option B — reproduce the race end-to-end via `entr` (hard, flaky).**

The fix actually matters at three of the four `notify(revision_event)` call sites which fire **outside** `revision_queue_lock`:

- `src/packagedef.jl:657` (user-callback path, dir watcher)
- `src/packagedef.jl:699` and `:711` (user-callback path / fall-through, file watcher)
- `src/pkgs.jl:520` (active-project change)

The remaining notify (`packagedef.jl:664`) fires under the queue lock, where revise() holding the lock would serialise the push and the notify and the race is harder to hit.

A test that drives `entr` with the *user-callback* path can in principle hit the lost-wakeup window:

1. `entr` with `pause=0`, callback `f()` sleeps ~100 ms.
2. Touch file A → `f()` starts.
3. While `f()` is sleeping, touch file B.
4. After `f()` returns, assert `f()` is invoked again within a timeout.

On master this **can** drop the second invocation; on this PR it should not. But the race window between `wait` returning and the explicit `reset` is microseconds, so master would also pass most of the time. To make the test discriminative you'd have to either:
- run hundreds of iterations and accept some statistical FAIL/PASS asymmetry, or
- intercept the `wait → reset` transition with a controlled `yield()` injection. That requires either parameterising `entr` for tests or duplicating its loop body.

**Option C — emulate `entr`'s loop in a tight test (medium, deterministic).**

Build a minimal consumer with the same `wait/(reset)/work` pattern but with deterministic yields, then check that a notify fired during `work` is serviced. This bypasses the file system entirely:

```julia
ev = Base.Event(true)               # what we ship
serviced = Threads.Atomic{Int}(0)
in_work = Base.Event(false)         # signals "now inside work"
done_work = Base.Event(false)       # released by the test to end work

consumer = @async begin
    while serviced[] < 2
        wait(ev)
        notify(in_work); wait(done_work); reset(done_work); reset(in_work)
        Threads.atomic_add!(serviced, 1)
    end
end

notify(ev)                  # first event
wait(in_work)               # consumer is now inside work
notify(ev)                  # second event arrives DURING work — must latch
notify(done_work)           # let first work finish
wait(in_work)               # consumer must enter work again on its own
@test serviced[] >= 1
notify(done_work)
wait(consumer)
@test serviced[] == 2
```

With `Base.Event(true)` this passes. Swap in `Base.Event()` + an explicit `reset` after `wait` and the second notify is racy. Deterministic except for the OS scheduling of `@async`, which in practice is reliable enough.

**Recommendation.** Option A is worth landing in this PR — one short testset that catches the most likely regression. Option C is a reasonable addition for the curious reviewer. Option B I would skip: real-file races at this granularity are below the noise floor of CI.

PR #1038, opened 2026-05-14
`@kwdef`: split methods and type (fix #1022)

timholy:
Editing only the default value inside a `@kwdef` struct used to trigger the full struct-revision path, walking the entire subtype tree of `Any` — many seconds (or minutes, on large dependency graphs) per edit. The struct itself is unchanged; only a keyword constructor differs.

One relatively straightforward fix would have been to do the analysis of the `@kwdef` block at the `Expr` level, splitting the `struct` expression from the method definitions that instantiate the default values. Based on past experience, I disliked this approach because it risks missing a lot, particularly for code that is generated by metaprogramming. Consequently, this PR takes a much more ambitious approach by doing its decision-making based on lowered code. The main reason this PR is so large is that it required a significant refactoring due to an order-of-operations constraint. Historically, we've handled deletion early (pre-lowering) to avoid risk that in moving code from one file to another, the changes to the file from which the code is being removed would be evaluated last and thus result in deletion from the running session. The more ambitious appproach taken with this PR requires that we get through much of lowering before we make any decisions about deletion.

This adds a non-destructive `:predict` mode to `methods_by_execution!` that captures the candidate `(typname, partial DataType, fieldtypes svec)` from each `Core._typebody!` call without executing it. `revise()` runs this prediction in a new Phase 1a, before `delete_missing!`, and feeds the result through `PredictedChanges`. `delete_missing!` then skips `handle_type_deletion!` when the new code redefines a type whose fingerprint matches the existing in-memory binding (compared via a new `structurally_equivalent` helper that handles UnionAll/TypeVar renaming).

`reuse_existing_type!` (introduced in #1035) is extended to also fire in `:eval`/`:evalmeth` mode behind the same structural-equivalence check, so the binding partition for the type is preserved and subtype caches don't churn. A paired `:const` skip helper recognises the case where the RHS would just rebind an unchanged value.

Gated on `__bpart__[]` (Julia 1.12+ + `revise_structs = true`): the predict pass relies on `_typebody!` interception, which doesn't exist on older Julia versions and would otherwise inadvertently define the struct as a side effect of prediction.

Verified end-to-end: a `@kwdef` default-value edit on a session with ~12k loaded subtypes drops from ~700ms to ~17ms (≈40× speedup), while genuine field/type/mutability changes still go through the full revision path. Adds five regression tests (positive + two negative shape-change cases + a method-deletion guard) under "struct revision (issue #1022)".

Fixes #1022

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


timholy:
Now that https://github.com/JuliaLang/julia/issues/61813 has been closed, let's try re-running this.

timholy:
Not yet on nightly, though (CC @adienes, nightly is a few days behind RN)

timholy:
Green! Thanks @adienes, this is a very nice change to be able to merge.

I'll hold just a bit (likely a few days) to give it one more review; this is a really intrusive change and I want to be sure it's worth it.

adienes:
yay! glad to hear it worked out

PR #1044, opened 2026-05-18
Give `entr` a real trailing-edge debounce

timholy:
I've been doing some work to try to make Revise's tests less flaky. #1037 was easy and seems to have been quite effective in getting rid of one whole class of sporadic failures. It's been much more difficult to get rid of flakiness in the `entr` tests (see #1039, #1040, #1041, #1042), despite the fact that these are a tiny fraction of Revise's tests.

My current sense is that a better approach is to change `entr` itself. I don't use it myself and it's always been community-maintained, so it's a little risky for me to be the one doing this. But with Claude as my sidekick and some careful supervision, it seems worth a try. I'd be really grateful if @ffevotte and/or @frankier could review this and/or try it out under real workloads.

Here is Claude's commit message:

`entr`'s coalescing of rapid file changes was an accident of locking: `process_user_callbacks!` runs callbacks under `@sync`, and the registered callback was `() -> (sleep(pause); f())`, so the `@sync` blocked for `pause` seconds while holding `revision_queue_lock`. A second change landing in that window had its queued key wiped by the trailing `empty!`; a change landing after it started a fresh callback. Whether two changes coalesced thus depended on their relative filesystem-event detection latency, which no OS bounds.

Replace this with a genuine trailing-edge debounce: the watch callback only records an extendable deadline and lazily spawns a single debounce task, which sleeps until `pause` seconds of quiet and then runs `f()`. The callback now returns immediately, so it no longer holds `revision_queue_lock` for `pause` seconds. Errors from `f()` are relayed to `entr`'s loop so a throwing `f()` still terminates `entr`.

Rework the `entr` tests accordingly: the "quick succession" checks now use a sustained burst rather than two lone events whose relative detection timing is unbounded, and the watch-arming probes are spaced more than `pause` apart so they no longer defeat the debounce. The directory-watch block is skipped under `JULIA_REVISE_POLL=1`, since poll mode watches a directory via its own `stat` and cannot detect in-place modification of a file inside it.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -137,10 +137,10 @@ any module tracked by Revise.
 
 `entr` will process updates (and block your command line) until you press Ctrl-C.
 Unless `postpone` is `true`, `f()` will be executed also when calling `entr`,
-regardless of file changes. The `pause` is the period (in seconds) that `entr`
-will wait between being triggered and actually calling `f()`, to handle
-clusters of modifications, such as those produced by saving files in certain
-text editors.
+regardless of file changes. The `pause` is the quiet period (in seconds) that
+`entr` waits after the most recent change before calling `f()`; a cluster of
+modifications less than `pause` apart — such as those produced by saving files
+in certain text editors — therefore triggers only a single call.
 
 # Example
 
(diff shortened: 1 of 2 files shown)


timholy:
Nice to see all green without re-running failing jobs! A first hint of possible success...

timholy:
Because I'm tired of flaky tests, I'm going to merge this. Can be revisited later if it proves problematic.

PR #1052, opened 2026-05-29
Recolor only the `julia>` prompt for revision errors

### LINKED ISSUE #755: REPL coloring gets confused by non-julia modes
Screenshot from 2023-08-04 23-04-58

timholy:
`maybe_set_prompt_color` operated on `repl.mistate.current_mode`, the mode active at the moment it ran. When a queued revision error first surfaced while the REPL was in shell mode, it saved the shell prompt's red into the single global save slot and recolored the shell prompt; when the error later resolved, the `:ok` branch wrote that saved red onto the then-active `julia>` prompt, leaving it red instead of restoring green.

Always target the `julia>` prompt (`repl.interface.modes[1]`, the accessor Julia's own REPL.jl uses) for both save and restore, so the yellow warning lands on `julia>` and restoration returns it to its own original color regardless of which mode was active.

Fixes #755

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1476,25 +1476,27 @@ end
 # Set the prompt color to indicate the presence of unhandled revision errors
 const original_repl_prefix = Ref{Union{String,Function,Nothing}}(nothing)
 function maybe_set_prompt_color(color)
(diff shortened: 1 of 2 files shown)


timholy:
Just one flaky test

PR #1053, opened 2026-05-29
Adjust to upcoming type system refactor

Keno:
JuliaLang/julia#61915 changes Type{T} to be represented as TypeEq{T}. Teach the old-type scanner to recurse through TypeEq via Base.type_parameter when available, so constructor signatures like Tuple{Type{T}, Any} are invalidated when T is redefined. Written by GPT 5.5.


timholy:
Wow. That's a big change. Long-desired, too.

timholy (approved):

PR #1056, opened 2026-05-30
Re-alias orphaned function bindings on revision

### LINKED ISSUE #239: Restart reason: Change in binding resolution
Say you accidentally write something like this in a package:
```
foo() = iterate(...)
iterate(a::Foo) = ...
```
The second definition is wrong. What was intended was to extend `Base.iterate`. Now, if I add the missing `Base.`, revise will delete the method, but the binding in `foo` will still resolve to the (now methodless) iterate function in my module. Not sure this can be addressed in the near term, but wanted to document it as a reason I had to restart my session.

timholy:
When a method is accidentally defined as `f(x::Foo) = ...` (creating a new module-local `f` that shadows an imported `Base.f`) and then corrected to `Base.f(x::Foo) = ...`, deleting the mistaken method left the now-methodless local `f` binding in place. Unqualified references elsewhere in the module kept resolving to it, so the package stayed broken until a session restart.

This was previously a hard Julia limitation: function bindings are `const` and could not be reassigned. Binding partitions (Julia 1.12+) permit a `const` redefinition, so when one revision both empties a module-owned function binding and adds a method to a different function of the same name, we read that as the method having moved and alias the orphaned binding onto the function the user actually extended. The repair is surfaced with an `@info`, since it is a binding mutation Revise infers rather than one the user wrote directly.

The change applies only on Julia 1.12+; on earlier versions the behavior is unchanged. This is not subject to the LocalPreferences.toml `revise_structs` setting. If it were, the name alone would be misleading. Moreover, the underlying purpose of that setting is to handle the fact that walking the type-tree is currently too slow to be on-by-default. This change does not involve the type-tree and should be fast enough to be on by default.

Fixes #239

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1153,6 +1153,121 @@ function redefine_bindings!(revision_errors::Vector{Tuple{PkgData,String}}, reev
     return revision_errors
 end
 
+# Extract the singleton function instance from a method signature `Tuple{typeof(f), ...}`.
+# Returns `nothing` for constructors (`Type{T}`) or callable objects with no singleton instance.
+function sig_function(@nospecialize(sig))
+    ft = Base.unwrap_unionall(sig)
+    ft isa DataType && !isempty(ft.parameters) || return nothing
+    F = Base.unwrap_unionall(ft.parameters[1])
+    (F isa DataType && isdefined(F, :instance)) || return nothing
+    return F.instance
+end
+
+# Collect the set of method signatures defined by the `SigInfo`s in `exs_infos`.
+function sigset(exs_infos::ExprsInfos)
+    sigs = Set{Any}()
+    for (_, exinfos) in exs_infos
+        exinfos === nothing && continue
+        for exinfo in exinfos
+            exinfo isa SigInfo || continue
+            push!(sigs, exinfo.sig)
+        end
+    end
+    return sigs
+end
+
+# Issue #239. An accidental definition such as
+#     foo() = iterate(x)        # intends to call Base.iterate
+#     iterate(x::Foo) = ...     # OOPS: defines a *new* `MyMod.iterate`, shadowing Base.iterate
+# creates a module-local `iterate` that shadows the imported one, and `foo`
+# binds to it. When the user corrects it to `Base.iterate(x::Foo) = ...`, Revise
+# deletes the wrong method, but the now-methodless `MyMod.iterate` binding
+# lingers; unqualified references keep resolving to it, so the package stays
+# broken until the session is restarted.
+#
+# This shadowing can only happen for a name that was implicitly in scope; an
+# explicit import errors when code like the above is encountered. Thus if an
+# edit (a) *empties* a module-owned function binding and (b) adds a method to a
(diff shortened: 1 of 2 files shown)


src/packagedef.jl
```diff
@@ -970,6 +970,92 @@ function redefine_bindings!(revision_errors::Vector{Tuple{PkgData,String}}, reev
...
+                isempty(methods(fdel)) || continue       # still has methods => not orphaned
+                parentmodule(fdel) === mod || continue    # only realias bindings this module owns
+                n = nameof(fdel)
+                fadd = get(added, n, nothing)
+                (fadd === nothing || fadd === fdel) && continue
+                # Alias the dead local binding onto the function the user actually extended.
+                # TODO: replicate to distributed workers (cf. `delete_method_by_sig`), e.g. by
+                # broadcasting `const n = GlobalRef(parentmodule(fadd), nameof(fadd))`.
+                try
+                    Core.eval(mod, Expr(:const, Expr(:(=), n, fadd)))
```
Keno: This works, but it'd be better to just delete the binding`, (Base.delete_binding), so that ordinary `using` resolution will pick the correct one as if the method had never been defined.

src/packagedef.jl
```diff
@@ -970,6 +970,92 @@ function redefine_bindings!(revision_errors::Vector{Tuple{PkgData,String}}, reev
...
+                s in newsigs && continue
+                fdel = sig_function(s)
+                fdel === nothing && continue
+                isempty(methods(fdel)) || continue       # still has methods => not orphaned
+                parentmodule(fdel) === mod || continue    # only realias bindings this module owns
+                n = nameof(fdel)
+                fadd = get(added, n, nothing)
+                (fadd === nothing || fadd === fdel) && continue
+                # Alias the dead local binding onto the function the user actually extended.
+                # TODO: replicate to distributed workers (cf. `delete_method_by_sig`), e.g. by
```
Keno: My opinion on this is that this is a Base limitation and that for an interactive distributed job, the binding table should be a globally coherent data structure (but that's a whole project).

src/packagedef.jl
timholy: Hmm, I'm realizing this is more complicated than I originally envisioned. If we could trust that every module executed a bare `using SomeDependency`, then we could trust that any exports of `SomeDependency` should have been imported; its lack due to a name conflict could then be safely corrected with the `using SomeDependency: somename` statement. But in a world of ExplictiImports.jl, many of us are switching to `using SomeDependency: somename`-style imports, and those statements are typically non-local to the change in the methods. Thus, to make it safe for Revise to execute the `using` statement, my original thought is that we'd have to scan the entire package and see what it imported. Now, in the case of `Base` specifically, we know all of its exports are imported by default unless it's a `baremodule`. But that's a bit of a special case.

I was thinking this `const` solved the problem neatly, but it doesn't really; it's in fact a more intrusive change to the intended meaning of the names in the module.

timholy: Ah, looks like the explicit import would cause this to be an error. So it's safer than I thought. Fix coming.

timholy:
CI failure is a flake, see #1064 🤞

PR #1057, opened 2026-06-01
Tolerate transient absence of watched paths

### LINKED ISSUE #523: "Revise is not watching" errors
Recently I'm seeing "Revise is no watching" errors again.
I don't have a reproducing example, unfortunately, so I'm not sure how to proceed looking into it. It sometimes pops up when I've not actually revised anything. Usually (or perhaps always) the error refers to some `]dev`ed package and to https://github.com/JuliaPlots/Plots.jl/blob/c392cc122707851927039ba361a6d1f71952085c/src/utils.jl#L152.
This example happened when doing a "Goto symbol" (ctrl+j,ctrl+g) in Juno:
```
julia> Revise.LogRecord(Error, evaluation error starting at C:\Users\sternlab\.julia\packages\Plots\qcY0P\src\utils.jl:152, lowered, Revise_e943ed8b, "C:\Users\sternlab\.julia\packages\Revise\ucYAZ\src\lowered.jl", 106, (mod=Plots, ex=begin┌ Warning: C:\Users\sternlab\.julia\dev\Dierckx\deps is not an existing directory, Revise is not watching
└ @ Revise C:\Users\sternlab\.julia\packages\Revise\ucYAZ\src\Revise.jl:636

    #= toplevel:151 =#
    for i = 2:4
        #= C:\Users\sternlab\.julia\packages\Plots\qcY0P\src\utils.jl:152 =#
        #= C:\Users\sternlab\.julia\packages\Plots\qcY0P\src\utils.jl:152 =# @eval begin
                #= C:\Users\sternlab\.julia\packages\Plots\qcY0P\src\utils.jl:153 =#
                RecipesPipeline.unzip(v::Union{AVec{<:Tuple{Vararg{T, $i} where T}}, AVec{<:_Point{$i}}}) = begin
                        #= C:\Users\sternlab\.julia\packages\Plots\qcY0P\src\utils.jl:154 =#
                        $(Expr(:tuple, (:([t[$j] for t = v]) for j = 1:i)...))
                    end
            end
    end
end)MethodError: no method matching (::Plots.var"#533#534")(::Int64)
The applicable method may be too new: running in world age 27243, while current world is 27881.
Closest candidates are:
  #533(::Any) at none:0 (method too new to be called from this world context.)
Stacktrace:
 [1] iterate(::Base.Generator{UnitRange{Int64},Plots.var"#533#534"}) at .\generator.jl:47
```
Using Revise 2.7.5, Julia 1.4.2

timholy:
A watched directory or file can disappear briefly and then return: a `Pkg.build`, a git checkout, an environment switch, or an editor's atomic-rename save can remove and recreate it within moments. Previously the watcher tasks reacted to any disappearance by emitting an alarming "is not watching" warning to stderr and then permanently abandoning the watch, so a benign blip both startled the user and silently stopped tracking the path until the session was reloaded.

Both watcher loops now wait up to `watch_reappear_grace[]` (default 5s) for the path to return. If it reappears, watching resumes silently; if it is deregistered (the package moved or was removed), the task stops quietly; only if it stays gone past the grace period do we warn and stop, as before. Genuine-deletion handling is unchanged: the deletion is still queued by the prior loop iteration, so only the final warn-and-stop is deferred.

Fixes #523

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -791,6 +791,12 @@ This is generally called via a [`Revise.TaskThunk`](@ref).
                     with_logger(SimpleLogger(stderr)) do
                         @warn "$dirname is not an existing directory, Revise is not watching"
                     end
+                    # Drop the watch registration as we stop. Otherwise, if `dirname`
+                    # is recreated later (e.g. switching back to a branch that has it),
+                    # `init_watching` would see the stale entry, assume a watcher is
+                    # already running, and never start a replacement — so edits to the
+                    # reappeared files would go unnoticed.
+                    @lock revise_lock delete!(watched_files, dirname)
                 end
                 break
             end
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -750,6 +750,29 @@ function init_watching(pkgdata::PkgData, files=srcfiles(pkgdata))
 end
 init_watching(files) = init_watching((@lock revise_lock pkgdatas[NOPACKAGE]), files)
 
+# Maximum time (in seconds) a watched path may be absent before Revise concludes
+# it was genuinely removed. A shorter absence is treated as transient: a
+# `Pkg.build`, a git checkout, an environment switch, or an editor's
+# atomic-rename save can briefly remove and recreate a directory or file, and in
+# those cases the watch should resume silently rather than stop and warn (#523).
+const watch_reappear_grace = Ref(5.0)
+
+# Block while a watched path is missing. `exists(path)` reports whether it is
+# currently present; `watchkey` is its entry in `watched_files`. Returns:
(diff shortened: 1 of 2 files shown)


pfitzseb:
I suspect I've seen Revise deleting all methods from a file and then never re-adding them when it reappeared while switching branches. This PR obviously doesn't address that issue, but maybe you've seen something similar while testing this?

timholy:
This does not address that, it's just a grace period. You'd suggest putting it in stasis and re-activating it if needed?

EDIT: fix coming...

timholy:
This should fix that behavior too: now it has 5s of grace period, and then stops watching *and* deletes the directory from the list of package paths. That will allow it to be re-added if the user switches back to a branch that has the directory.

PR #1072, opened 2026-06-10
Skip the struct-revision walk for unchanged types

timholy:
This is a replacement for #1038. I was unhappy with how intrusive that was; this is still intrusive at the level of the `packagedef` orchestration (`revise()` pipeline), but the intrusive changes to types, lowering, and `visit` are essentially gone with this implementation. The key realization was that this really shouldn't require anything beyond a type-equivalence check; the only thing that makes it complicated is that disentangling the type from surrounding code can't be robust without lowering. So the "predict" mechanism from #1038 is retained, but the rest is far more straightforward while also being more correct (e.g., leaning on Julia internals for the type-equivalence machinery rather than trying to decide it within Revise). I think the overall line-count may be larger, but that's mostly due to expanded tests which capture a couple of bugs in the implementation of #1038.

I plan to merge this version as soon as tests pass. Below is the full commit message.

Closes #1038.

----

Editing only the default value inside a `@kwdef` struct used to trigger the full struct-revision path, walking the entire subtype tree of `Any` and re-evaluating every dependent. This is a lot of overhead for something that could be implemented as a simple method redefinition.

Before applying deletions, `revise` now predicts which already-defined types the new source re-creates unchanged, and skips `handle_type_deletion!` for those. The prediction interprets only the lowered statements feeding each `Core._typebody!` call (plus `eval` calls, so metaprogrammed definitions are covered), skipping anything that would publish a binding or define a method. If they check out as equivalent to the existing type, the deletion is skipped; if not, the deletion proceeds as normal. When deletion is omitted, the normal "is the redefined type equivalent?" machinery works as usual.

Because predictions are made before any queued change is applied, a same-revision change to a binding that the type's structure depends on (a field-type alias, a supertype, a parameter bound) can falsify them. Each skipped deletion is therefore re-checked after evaluation.

The prediction pass runs only when a type deletion is actually pending, only on Julia 1.12+ (`__bpart__[]`), and spans all queued files so a type moved between files is still recognized.

Fixes #1022

### DIFF
--- a/src/lowered.jl
+++ b/src/lowered.jl
@@ -642,3 +642,175 @@ function skip_declare_const(interp::Interpreter, frame::Frame, stmt::Expr)
     name isa Symbol || return false
     return @invokelatest isdefined(mod_arg, name)
 end
+
+## Prediction of type preservation (issue #1022)
+#
+# Before deletions are applied, `revise` asks whether the new source re-creates
+# each old type unchanged. If so, the subtype-tree walk in
+# `handle_type_deletion!` is unnecessary. Problems arise from `struct`
+# definitions like those in #1022, which disguise the fact that the type is
+# unchanged and only surrounding code is changed. We fix this with a two-phase
+# evaluation: first, a limited evaluation that reconstructs (from the lowered
+# code) the *proposed* new type (henceforth called the "prediction"). This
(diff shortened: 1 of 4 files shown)


timholy:
CC @adienes, @cstjean, @lassepe

Issue #945, opened 2025-07-30
Turn Revise off for a specific folder


asinghvi17:
Our usecase is that we have a tool that generates Julia files (the Dyad compiler) into a `YourPackage.jl/generated` folder, whose contents get `include`d into `src/YourPackage.jl`.  The Dyad compiler changes a bunch of files and while it’s working, we would like for Revise to not reload things. 

If you happen to run Revise by typing something into the REPL while Dyad compilation is running, you can run into an issue where files that have an include statement no longer exist, or some functions may not exist, because the Dyad compiler has removed them while preparing to overwrite them.

It would be nice to have Revise pick up a `.no-revise` file in the folder.  If it sees this it should not revise anything in that directory.

Some concerns that @StefanKarpinski brought up were:

> what happens if there are multiple processes that are trying to do this

Any process that writes Julia code could be thought of as a compiler of a sort.  You wouldn't run `cc` in multiple windows, simultaneously, on the same directory, and expect a sensible result.  So this can be considered undefined and bad behaviour.

More generally, any software should detect the presence of a `.no-revise` and be smart enough to handle that if it sees it.

> what happens if deleting the file doesn’t happen? Then you have this file laying about and it never gets deleted and revise effectively stops working until someone/something deletes it. 

The process that created the .no-revise would have to handle it and delete it.  Barring segfaults etc. this seems broadly doable.

If this doesn't make sense or there are other solutions, would love to hear about them!  For what it's worth this is running in a separate process, so I don't think we can use #943 here (though that would have been a great solution if we could).

pankgeorg:
> or there are other solutions, would love to hear about them!

Would a similar effect be achieved if you debounced the FileWatching events? A debounce of 1-2 seconds would probably be good enough for your usecase, especially if you create your files somewhere and "atomically" copy them over (atomic copy = rename in the same device)

asinghvi17:
The delays in file generation can be longer than that though.  And there is never really an atomic operation on the filesystem...especially if you have 1k+ files.

pankgeorg:
> And there is never really an atomic operation on the filesystem...especially if you have 1k+ files.

Moving an inode is atomic, with these guarantees https://man7.org/linux/man-pages/man2/rename.2.html.
[...]

PR #1075, opened 2026-06-12
Queue event-named files unconditionally

timholy:
The directory watcher decides whether a tracked file changed by comparing its ctime against the stored value. The kernel stamps inodes with its coarse clock, which advances at timer-tick resolution (~10 ms at HZ=100; ticks stretch further under virtualization — a probe on WSL2 showed 500 rm+write cycles over 12 ms receiving only 3 distinct ctimes, in exact 9.97 ms steps). A delete-and-recreate landing within one tick of the recorded stamp therefore produces an identical ctime, and no other stat field rescues the comparison: ext4 immediately reuses the freed inode number, and a same-length rewrite preserves the size. The watcher concludes "unchanged" and the revision is silently and permanently lost. The rapid rewrite loop from #945 (bradcarman/ReviseTest.jl distilled: `rm` + rewrite + immediate `revise()`) hit this about once per 500–700 iterations; Windows' ~15.6 ms tick makes it likelier there.

The filesystem events themselves are the reliable signal, and `watch_folder` already reports the name of the changed entry — `wait_changed_dir` previously discarded it. Now the names are collected and any tracked file named in an event is queued regardless of the timestamp comparison. A spurious queue is cheap (`revise` parses and diffs ASTs, so an unchanged file is a no-op); a false negative is unrecoverable.

The timestamp sweep is retained for files the events did not name, and everything falls back to it when the names are unknown: polling mode, a torn-down monitor, or an event arriving without a name. Names only add queueing, never suppress a check, so a platform whose event names didn't match tracked entries would degrade to exactly the old behavior.

The per-file decision moves into `scan_changed_files`, unit-tested with a fabricated stored-ctime collision (timestamp tricks can't be staged deterministically through the filesystem). Empirically: 12×300-iteration runs of the rewrite loop (3,588 cycles) show zero losses, where master averaged ~5.

Stacked on #1074 (shared watcher-code edits); will rebase onto master once that merges, so the diff to review here is just the head commit.

### DIFF
--- a/src/pkgs.jl
+++ b/src/pkgs.jl
@@ -376,14 +376,13 @@ function wait_changed_dir(dirname::AbstractString)
     end
     changed = Set{String}()
     complete = true
-    note!(name) = isempty(name) ? (complete = false) : (push!(changed, name); true)
     try
         name, _ = watch_folder(dirname)              # block for the next buffered event
-        note!(name)
+        isempty(name) ? (complete = false) : push!(changed, name)
         while true                                   # drain a burst delivered in one wakeup
             name, event = watch_folder(dirname, 0)
             event.timedout && break
-            note!(name)
+            isempty(name) ? (complete = false) : push!(changed, name)
         end
     catch e
         # EOFError: monitor torn down; let caller re-check state. issue #459: Ctrl-C.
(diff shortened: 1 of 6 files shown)


src/pkgs.jl
```diff
@@ -368,63 +372,98 @@ end
...
     end
-    return
+    return complete ? changed : nothing
 end
 
-function watch_files_via_dir(dirname::AbstractString)
-    wait_changed_dir(dirname)  # block until the directory changes (buffered on notifying filesystems)
+# Content hash for disambiguating events whose ctime is unchanged. Reads the
+# file, so call it only on event-named files, not in the per-directory sweep.
+filehash(path::AbstractString) = hash(read(path))
```
pfitzseb:
```suggestion
filehash(path::AbstractString) = open(crc32c, path)
```
may be a worthwhile optimization here

PR #1076, opened 2026-06-13
Build the type universe in a single binding sweep

### LINKED ISSUE #925: "world prior to its definition" error when modifying struct
I was working on some code in Julia 1.12.0-beta3, using Revise 3.8.0. Specifically on a section

```julia
struct Shift{T <: AbstractVector}
    irrelevant_field::Int # for the MWE
    b::T
end

shift(b::AbstractVector) = Shift(b)
```

within a package.

I modified the type `Shift` deleting `irrelevant_field` and after evaluating something in the REPL, I got
```julia
WARNING: Detected access to binding `LogDensityTestSuite.shift` in a world prior to its definition world.
  Julia 1.12 has introduced more strict world age semantics for global bindings.
  !!! This code may malfunction under Revise.
  !!! This code will error in future versions of Julia.
Hint: Add an appropriate `invokelatest` around the access to this binding.
To make this warning an error, and hence obtain a stack trace, use `julia --depwarn=error`.
```

But everything seems to work fine, `shift` has the right methods, etc.

### LINKED ISSUE #988: Struct revision takes several minutes for packages with large type trees

### LINKED ISSUE #993: World age warning when using with numbered prompt.
Put these to `startup.jl`:
```
using Revise
import REPL

atreplinit() do repl
    if !isdefined(repl, :interface)
        repl.interface = REPL.setup_interface(repl)
    end
    REPL.numbered_prompt!(repl)
end
```
Open julia, using some package, say `using LinearAlgebra`, then I get this warning:
```
WARNING: Detected access to binding `MainInclude.Out` in a world prior to its definition world.
  Julia 1.12 has introduced more strict world age semantics for global bindings.
  !!! This code may malfunction under Revise.
  !!! This code will error in future versions of Julia.
Hint: Add an appropriate `invokelatest` around the access to this binding.
To make this warning an error, and hence obtain a stack trace, use `julia --depwarn=error`.
```

julia: 1.12.4
Revise: 3.13.1

timholy:
`collect_all_subtypes(Any)` drove struct revision by calling `InteractiveUtils.subtypes` once per type. `subtypes` rescans the names of every loaded module on each call, so the recursive walk cost one system-wide name scan per abstract type — O(#abstract-types × #names), which reaches minutes on sessions with large type trees.

Enumerate the type universe in one pass over module bindings instead (`all_named_types`), placing each type by its own identity. The cost is O(#names): ~13 ms at 29k types and ~40 ms in a 69k-type / 670-module session; the recursive walk took ~5.5 s at 29k types and the minutes reported in #988 on large trees.

The single pass also removes a world-age hazard. The recursive walk ran long enough that a concurrent revision advanced the world past the walk's `tls_world_age`; it then read freshly-defined bindings (types injected by Requires, or the struct being revised) at a world prior to their definition, printing "Detected access to binding ... prior to its definition world". A single pass completes at one consistent current world, so there is no stale window.

With the scan this cheap, drop the background task that pre-populated the field-type cache at startup: the cache now fills lazily during `old_types_with`, and the per-revision scan is fast enough to run on demand. Removing the task also retires its Apple-Silicon workaround (JuliaLang/julia#60721), which only applied to the spawned thread.

Fixes #988; fixes #993; fixes #925.


src/visit.jl
```diff
@@ -31,14 +31,53 @@ function old_methods_with(oldtypename::Core.TypeName)
...
+# (O(#abstract-types × #names)), whereas one binding sweep is O(#names).
+function all_named_types()
+    types = Base.IdSet{Type}()
+    seen = Base.IdSet{Module}()
+    work = Base.loaded_modules_array()
+    while !isempty(work)
+        m = pop!(work)
+        m in seen && continue
+        push!(seen, m)
+        for s in names(m; all=true)
```
pfitzseb:
```
@static if VERSION < v"1.12-"
    unsorted_names(m::Module; all::Bool=false, imported::Bool=false, usings=false) =
        ccall(:jl_module_names, Array{Symbol,1}, (Any, Cint, Cint), m, all, imported)
else
    unsorted_names(m::Module; all::Bool=false, imported::Bool=false, usings=false) =
        ccall(:jl_module_names, Array{Symbol,1}, (Any, Cint, Cint, Cint), m, all, imported, usings)
end
```
is significantly faster (~6x for Base), but also more brittle given that the ccall signature changed with the 1.12 release.

timholy: I didn't think it would matter compared to the type-machinery, and thus didn't want to have to worry about divergence among versions. But benchmarking indeed shows it's non-negligible, maybe another 20-30% on a large system. Worth having. Follow-ons coming.

src/visit.jl
```diff
@@ -31,14 +31,53 @@ function old_methods_with(oldtypename::Core.TypeName)
...
 end
 
 function foreach_subtype(f::Function, @nospecialize(parent_typ::Type))
+    if parent_typ === Any
+        for Ty in all_named_types()
+            f(Ty)
+        end
+        return nothing
+    end
     _foreach_subtype!(f, parent_typ, Base.IdSet{Type}())
```
pfitzseb: `_foreach_subtype!` seems like it should have strictly worse performance than doing `subtype` checking in the loop, given that it ends up calling `subtypes` (which recomputes `names` again).

timholy: That's now dead code anyway and the reason for the CI failure. Removal already queued locally.

src/visit.jl
```diff
@@ -31,12 +31,42 @@ function old_methods_with(oldtypename::Core.TypeName)
 end
 
 function collect_all_subtypes(@nospecialize(parent_typ::Type))
+    parent_typ === Any && return all_named_types()
     return _foreach_subtype!(Returns(nothing), parent_typ, Base.IdSet{Type}())
```
pfitzseb: This still calls `subtypes` and recurses. Should probably be switched to a single loop over `all_named_types()`

aviatesk: > Should probably be switched to a single loop over `all_named_types()`

+1. And we can go a bit further: the only non-`Any` caller is `record_invalidations_for_type_deletion!`, which already has the full `alltypes = collect_all_subtypes(Any)` set in scope. So instead of a fresh sweep we can just filter what's already there:

```julia
# in record_invalidations_for_type_deletion!, replacing `collect_all_subtypes(oldtype)`
oldsubtypes = Base.IdSet{Type}(t for t in alltypes if t !== oldtype && t <: oldtype)
```

That removes the last `subtypes` recursion (which re-scans `names` per type — exactly the cost this PR eliminates for `Any`), lets `_foreach_subtype!` be deleted, and avoids re-sweeping at every level of the recursion.

One subtlety worth an assert/comment if we do this: `subtypes(T)` matches by supertype `TypeName` and then `typeintersect`s the result, whereas `t <: T` does not narrow. The two differ for a specific parametric instantiation, e.g. with `struct PC{S} <: P{S} end`, `subtypes(P{Int})` yields `PC{Int}`, but `PC <: P{Int}` is `false`, so the filter would drop it.
This never bites Revise though. `oldtype` is always the canonical binding (`getglobal(mod, name)` → the `P{T} where T` UnionAll, never `P{Int}`), and the recursion stays on canonical forms. So `test_visit.jl` passes unchanged with the filter; the divergence only appears if a concrete instantiation is ever passed in.

aviatesk:
Thank you for taking this on, resolving a known performance issue in the struct-revision path that I'd implemented but left unsolved.

I was aware that `InteractiveUtils.subtypes` was the bottleneck, but I couldn't come up with a way to drive the recursive type walk without it.

timholy:
It turned out to be a 1-2 and that was what fooled me: after #1013 I thought I had identified all the ways it could be sped up (the `do` block form had an over-specialization issue). Back then I *had* spent quite a long time looking at profiles, but at the time I didn't realize the O(m*n) cause. I caught this while seeing whether https://github.com/JuliaLang/julia/pull/61804 could help; it turns out to be useless for this particular problem, but just by chance I happened to notice this deeper issue, it was really the scaling vs # of packages that I should have noticed earlier, but didn't pay enough attention to.

timholy:
I'm going to leave type-revision opt-in for at least the next minor bump. But anyone watching might want to enable it now with https://timholy.github.io/Revise.jl/stable/config/#Enabling-struct-revision

topolarity:
This is great. Thanks @timholy !

Are there any remaining slow scans / updates in struct revision, or was this the last bottleneck to clear?

timholy:
I think it was the last; AFAIK this can now become routine.

I haven't yet come on a case to actually try it out yet, but I've finally given up my attachment to 1.10 and activated type-revision so that I should sometime get a chance to try my own dog food.

PR #1078, opened 2026-06-13
Fix minor issues found by JETLS

aviatesk:
- Use the correct `Base.Filesystem.path_separator` (was undefined)
- Drop unused/redundant variable assignments
- Return the `@lock` block result directly instead of via outer `local`
- Sort `using` imports alphabetically


timholy:
Thanks!

PR #1081, opened 2026-06-14
Warn when yry() times out expecting a revision

timholy:
`yry()` waits for the file watcher to queue a pending edit, falling through after `event_timeout`. A dropped or delayed filesystem event hits that timeout and then revises nothing, so the staleness only surfaces later as a confusing "revision not applied" assertion failure (the typical macOS FSEvents CI flake).

Add `expect_revision::Bool=true`: when a revision is expected, a timeout now emits a warning naming the likely dropped event, flagging the flake at its source. Tests that legitimately expect no revision (writes to an untracked path, re-revising with nothing new, including a missing file) pass `expect_revision=false` to take the timeout silently. `@yry` forwards the keyword.

This does not fix any reported bug, its purpose is only to help diagnose occasional remaining flaky tests.

### DIFF
--- a/test/common.jl
+++ b/test/common.jl
@@ -61,15 +61,26 @@ end
 #     path re-watches a single file after each revision and, unlike the
 #     buffered per-directory watcher, has neither an event queue nor a content
 #     hash to recover a write that lands while it is between watches.
-function yry()
-    timedwait(() -> !isempty(Revise.revision_queue), event_timeout; pollint=0.02)
+#
+# `expect_revision=true` (the default) means a write should already have been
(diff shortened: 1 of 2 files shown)


timholy:
Doc failure is a network issue. What's really interesting is the macOS failure: it's a flaky test (it usually passes, but this is perhaps the most common site for failure), but the failure is *not* due to a timeout! Quite a surprising result. Will require further investigation on its own, but there's nothing blocking this from merging, it's already a significant improvement in our ability to diagnose the cause.

PR #1083, opened 2026-06-14
DEBUG: instrument the Manual-track revision flake

timholy:
Not for merge. Captures diagnostic state after each write→@yry→@test sequence in the "Manual track" testset (revise_f, revise_floc, f264) to diagnose the macOS-only "revision not applied" flake, which does not reproduce on Linux under inotify, polling, or per-file watching.

Each capture is taken after @yry and is deliberately yield-free (no I/O, no lock) so it cannot mask the race: it records the observed value, the revision_queue length, and the expressions revise() actually stored for the file. The stored exprs decide the mechanism — new body with a stale value means detected+parsed but not applied (eval/diff/world-age); old body means revise() never re-parsed the file (detection/queue timing). Dumped once to stderr at the end of the three sequences.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3790,7 +3790,37 @@ end
         pop!(LOAD_PATH)
     end
 
+    # TEMP debug instrumentation for the macOS Manual-track revision flake. Captured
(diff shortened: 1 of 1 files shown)


timholy:
Closing victoriously: the issue was a "leftover" queue item from an earlier step.

PR #1099, opened 2026-07-04
test: Write the revise_structs preference to the file that has priority

Keno:
When the tests run in a `Pkg.test` sandbox (testing Revise as a dependency of another project, as julia's own CI does), Pkg has materialized the ambient preferences — even when empty — as `JuliaLocalPreferences.toml` since JuliaLang/Pkg.jl#2920, and the preference resolution in `base/loading.jl` stops at that file, shadowing any `LocalPreferences.toml` written next to it. The `revise_structs preference` test therefore always read the default in such setups. Write to whichever preferences file actually takes priority. The checkout-as-active-project flow (julia-runtest) is unaffected: no preferences file pre-exists there, and the test keeps using `LocalPreferences.toml`.


timholy:
Thanks!

PR #1103, opened 2026-07-21
Limit broad definition fallback for `Base` submodules

aviatesk:
Definition lookup for methods owned by child modules of `Base`, `Core`, or `Compiler` could fall through to parsing every source file when the recorded source and module-local files did not contain the definition.

Keep direct and module-local lookup available for child modules so macro-generated methods such as `Float32(::Irrational{:π})` can still be found, but stop before the package-wide fallback. `Method`s owned directly by the roots retain the early guard.


timholy:
Good idea, many thanks!

PR #1109, opened 2026-07-23
handle include string types that are defined after the Revise world age

KristofferC:
happens for example for Plots that has a RelocatableFolder.Path type


pfitzseb (approved):

timholy:
Thanks! So many unexpected corner cases...

PR #1110, opened 2026-07-23
Improve signature extraction failure message

aviatesk:
Signature extraction partially evaluates source after a module has loaded.
When reconstruction fails, Revise currently leads with the underlying
runtime exception. This makes valid package code look broken and obscures
that the failure occurred only while using `mode=:sigs`.

Before:

    MethodError: no method matching iterate(::Nothing)
    ...
    Revise evaluation error at <path>/src/datatypes.jl:124

After:

    failed to extract method signatures for MPI at
    <path>/src/datatypes.jl:124
    Revise could not determine which methods this top-level expression
    defines without re-running module initialization. Generated definitions
    may depend on temporary state created by earlier top-level operations,
    such as mutation or `@eval`.
    ...
    caused by: MethodError: no method matching iterate(::Nothing)

The example above is from JuliaParallel/MPI.jl#957.

Wrap failures at the `instantiate_sigs!` `mode=:sigs` boundary in a
dedicated `SignatureExtractionError`. Preserve explicit `rethrow` behavior
and other evaluation modes, retain the synthetic user-code stacktrace, and
avoid appending Revise's internal backtrace when logging the error.

This makes state-dependent generated definitions such as those in
aviatesk/JET.jl#776, aviatesk/JET.jl#804, and aviatesk/JET.jl#826
actionable. The package-side changes in Nemocas/Nemo.jl#2208 and
JuliaParallel/MPI.jl#957 demonstrate how to remove these dependencies.


timholy:
Much better!

PR #1121, opened 2026-08-22
Fix cache_src_id MethodError on 32-bit Julia

mestinso:
`cache_src_id` hashes with a hardcoded `UInt64` seed, but `hash`'s seed argument must be word-size-native `UInt`. On 32-bit Julia this throws `MethodError: no method matching hash(::UInt64, ::UInt64)` whenever Revise's package-watching callback parses a newly precompiled package's cache header (Julia 1.11+), i.e. on almost every `using` of a new package.

Fix: `UInt64(inc.hash)` -> `UInt(inc.hash)`.

Tested on Julia 1.12.7 (i686-w64-mingw32): the error is gone, no change in behavior on 64-bit.


mestinso:
Looks like @davidanthoff beat me to it, though our fix is slightly different. Feel free to abandon....

#1120

timholy:
Thanks! I went with #1120

PR #1130, opened 2026-08-30
Close the polling-mode startup gap in revise_file_queued

### LINKED ISSUE #1129: Polling mode misses file edits made before the watcher tasks first run
With `JULIA_REVISE_POLL=1` (or any non-notifying path), the per-file watcher tasks spawned by `init_watching` take their `poll_file` baseline when they first run — i.e. at the process's first scheduler yield. A file edit made before that yield is never detected: `poll_file` baselines on the already-edited file, and the ctime recorded in `WatchList.file_ctimes` is never consulted on this path. (The notification path closes the equivalent gap with the buffered `watch_folder` registered in `init_watching`.)

Repro (Revise 3.17.0, Julia 1.12.7, Linux, local /tmp — no special filesystem needed):

```julia
# JULIA_REVISE_POLL=1 julia --startup-file=no repro.jl
using Revise
testdir = mktempdir()
dn = joinpath(testdir, "DemoPkg", "src")
mkpath(dn)
srcfile = joinpath(dn, "DemoPkg.jl")
write(srcfile, "__precompile__(false)\nmodule DemoPkg\nf() = 1\nend\n")
sleep(0.5)
push!(LOAD_PATH, testdir)
using DemoPkg
# edit BEFORE the first yield after loading:
write(srcfile, "__precompile__(false)\nmodule DemoPkg\nf() = 2\nend\n")
for i in 1:15
    sleep(2)
    Revise.revise()
    Base.invokelatest(DemoPkg.f) == 2 && (println("revised after ", 2i, "s"); exit(0))
end
println("edit never seen (bug)")
```

prints `edit never seen (bug)`. Any yield between `using DemoPkg` and the `write` (even a `println`) masks it, which makes this confusing to hit in practice: in a script that loads a package, computes without yielding (e.g. compilation-heavy work), and then edits a tracked file, the edit is silently lost forever. It may also explain the "I'm not sure why 2 sleeps are better than one" comment in `test/polling.jl`.

Fix incoming in a PR.

AsafManela:
Fixes #1129.

In `revise_file_queued`, compare the file's ctime against the baseline `init_watching` recorded in `WatchList.file_ctimes` before blocking in `wait_changed`, and treat a mismatch as an already-delivered change. The baseline is kept current on every detected change, so the check only fires for genuinely missed ones.

The new "Polling startup gap" testset fails on master and passes with this change; the existing polling testset still passes.

https://claude.ai/code/session_01FjiTcmuU8NCBHp5vJfNnR9

### DIFF
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -1345,10 +1345,10 @@ This is generally called via a [`Revise.TaskThunk`](@ref).
 
 This is used only on platforms (like BSD) which cannot use [`Revise.revise_dir_queued`](@ref).
 """
-function revise_file_queued(pkgdata::PkgData, file)
-    if !isabspath(file)
-        file = joinpath(basedir(pkgdata), file)
-    end
+function revise_file_queued(pkgdata::PkgData, filename)
+    # `file` is captured by closures below, so it must be assigned exactly once
+    # (a second assignment would force it into a `Core.Box`).
+    file = isabspath(filename) ? filename : joinpath(basedir(pkgdata), filename)
 
     dirfull, filebase = splitdir(file)
     fileexists(f) = file_exists(f) || isdir(f)
@@ -1375,10 +1375,8 @@ function revise_file_queued(pkgdata::PkgData, file)
                 status = await_watched_path(fileexists, file, dirfull)
                 if status !== :reappeared
                     if status === :gone
-                        let file=file
-                            with_logger(SimpleLogger(stderr)) do
-                                @warn "$file is not an existing file, Revise is not watching (watching resumes if it reappears)"
-                            end
+                        with_logger(SimpleLogger(stderr)) do
+                            @warn "$file is not an existing file, Revise is not watching (watching resumes if it reappears)"
                         end
                         relinquish_watch(dirfull, file)
                     end
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -179,6 +179,27 @@ function nonnotifying_path(path::AbstractString)
     return fstype == "9p" || fstype == "drvfs"
 end
 
+# Matches the default polling interval of `Base.poll_file`.
+const saved_state_poll_interval = 5.007
+
+"""
+    poll_from_saved_state(file, prev_ctime)
+
+Block until `file`'s ctime differs from `prev_ctime`, the value recorded by
+`init_watching`.
+
+This exists because `poll_file` samples its own baseline when the poll starts: a change
+that landed between `init_watching` and the first poll becomes the baseline, and the poll
+then waits for the *next* change. Seeding the comparison with the recorded ctime leaves no
+window in which a change can be adopted as the starting state.
+"""
+function poll_from_saved_state(file, prev_ctime)
+    while ctime(file) == prev_ctime
+        sleep(saved_state_poll_interval)
+    end
+    return nothing
+end
+
 function wait_changed(file)
     poll = polling_files[] || nonnotifying_path(file)
     try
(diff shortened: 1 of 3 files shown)


src/packagedef.jl
```diff
@@ -1329,8 +1329,25 @@ function revise_file_queued(pkgdata::PkgData, file)
...
-    dirfull, _ = splitdir(file)
+    dirfull, filebase = splitdir(file)
     fileexists(f) = file_exists(f) || isdir(f)
+    # Baseline recorded by `init_watching`. In per-file (polling) mode a change
+    # that lands between `init_watching` and this task's first `wait_changed`
+    # would otherwise be lost forever: `poll_file` takes its own baseline when
+    # the poll starts, and the buffered directory monitor that closes this
+    # startup gap for the notification path is unavailable here. So before
+    # blocking, compare the file's ctime against the recorded baseline and
+    # treat a mismatch as a change that already happened.
```
timholy:
```suggestion
  # `init_watching` saved the file's ctime before scheduling this task.
  # The file may have changed before this task starts. `poll_file` would
  # then use the changed file as its starting point and wait for another
  # change, so first compare the current ctime with the saved value.
```

src/packagedef.jl
```diff
@@ -1349,12 +1366,18 @@ function revise_file_queued(pkgdata::PkgData, file)
...
+            if prev === nothing || prev == 0.0 || ctime(file) == prev
+                try
+                    wait_changed(file)  # will block here until the file changes
+                catch e
+                    # issue #459
+                    (isa(e, InterruptException) && throwto_repl(e)) || throw(e)
+                end
             end
+            # Keep the baseline current so the pre-block check above only fires
+            # for genuinely missed changes, never for the one just delivered.
```
timholy:
```suggestion
            # Save the ctime observed after this change so it is not queued again.
```

src/packagedef.jl
```diff
@@ -1349,12 +1366,18 @@ function revise_file_queued(pkgdata::PkgData, file)
...
             end
-            try
-                wait_changed(file)  # will block here until the file changes
-            catch e
-                # issue #459
-                (isa(e, InterruptException) && throwto_repl(e)) || throw(e)
+            prev = stored_ctime()
+            if prev === nothing || prev == 0.0 || ctime(file) == prev
+                try
+                    wait_changed(file)  # will block here until the file changes
```
timholy: While this is a big improvement, I wonder whether there's the possibility of a race occurring between the equality check and when `wait_changed` -> `poll_file` starts running. Conceptually we might want
```julia
  if polling_files[] || nonnotifying_path(file)
      poll_from_saved_state(...)
  else
      wait_changed(file)  # notification-based watch_file
  end
```
where `poll_from_saved_state(...)` is a local rewrite of `poll_file` to start polling from the time recorded by `init_watching` rather than the time stamp that would be collected by `poll_file`.

Alternatively, we could pass the recorded data as a kwarg to `wait_changed` and put the `poll_file` rewrite there.

timholy (commented):
Thanks for the issue and fix! These race conditions are annoying, and not many people use polling file systems so it makes this especially precious.

On requested changes:
- I find Claude's prose somewhat torturous; I confess I don't always edit it in my own commits (bad me), but since I was reviewing an external contribution I thought I'd try to phrase it more clearly.
- Check my reasoning carefully on the possibility of a race that this doesn't seem to fix. If I'm wrong, obviously no fix needed, but if genuine then a small redesign may be in order.

AsafManela:
Hi Tim, I agree Claude can be way too wordy in a way that can exhaust readers. Free marginal generation I guess... But it does let me contribute to great projects like yours which would be too costly otherwise.

I agree it can use a redesign rather than a patch on the check.

The window is between `ctime(file) == prev` and the `lstat` that `poll_file` takes inside `wait_changed`. A write landing there makes `poll_file` baseline on the new content and wait for a *further* change, which is the original bug with a microsecond window instead of a multi-second one. Check-then-block can't be made airtight; only polling *from* the recorded state removes the window.

Pushed your first option: a local `poll_from_saved_state` in `revise_file_queued`, so `wait_changed` stays untouched for the notification path.

One correction to an earlier draft of this comment: I said I would store the `lstat` from `init_watching` and compare whole `StatStruct`s. That is wrong. `WatchList.file_ctimes` is a ctime, and `pkgs.jl:704` already uses ctime as the change signal on the directory path, so a parallel `StatStruct` would leave the recorded baseline and the poll comparison using different notions of "changed". ctime moves on any write, so nothing is lost by staying with it.

Also: my "lost forever" in the issue was too strong. The edit is missed until the file changes again, at which point the reread picks up both. Same user-visible effect, but the issue overstates it.

The repro revises in 2s under `JULIA_REVISE_POLL=1` and on the notification path; `test/polling.jl` is 3/3 and 2/2. The `yry` timeout warning in that file reproduces identically without my change.

Taking the comment suggestions as-is; yours read better than mine.

timholy:
> Hi Tim, I agree Claude can be way too wordy in a way that can exhaust readers. Free marginal generation I guess... But it does let me contribute to great projects like yours which would be too costly otherwise.

Completely agreed. Honestly, even packages I understand well (like Revise) were cumulatively more costly to maintain than I could really afford, so the agents are a big step forward.

But I also know that despite the AI help, these kinds of contributions still take real time to document and shepherd through. Thanks again for contributing, it's a very nice fix! I'll push some fixups to get CI passing.

PR #1139, opened 2026-09-14
Make `revise()` use the pkgid generated by `track()`

JamesWrigley:
Previously these would differ in some situations.


timholy:
Thanks!

